PR #11, opened 2018-04-13
Remove argument internalNames from ArchivePacker::CreateVolume

Brett208:
- Generate internalNames from within ArchivePacker instead of forcing users of the library to generate internalNames manually, standardizing the process.
 - Add error checking to throw an exception when creating volumes containing 2 files with the same internalName (filename not including path).

### DIFF
--- a/Archives/ArchivePacker.cpp
+++ b/Archives/ArchivePacker.cpp
@@ -1,4 +1,7 @@
 #include "ArchivePacker.h"
+#include "../XFile.h"
+#include "../StringHelper.h"
+#include <stdexcept>
 
 namespace Archives
 {
@@ -47,5 +50,20 @@ namespace Archives
 		DeleteFileA("Delete.vol");
 
 		return true;
+	}
+
+	std::vector<std::string> ArchivePacker::GetInternalNamesFromPaths(std::vector<std::string> paths)
+	{
+		std::vector<std::string> fileNames;
+
+		for (std::string fileName : paths) {
+			if (StringHelper::ContainsStringCaseInsensitive(fileNames, fileName)) {
+				throw std::runtime_error("Unable to create an archive containing multiple files with the same filename.");
+			}
+
+			fileNames.push_back(XFile::GetFilename(fileName));
+		}
+
+		return fileNames;
 	}
 }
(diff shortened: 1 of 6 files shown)


Brett208:
It isn't clear cut that the function name that ArchivePacker::GetInternalNamesFromPaths can throw an exception for 2 filenames being equivalent. However, implementing this exception inside the function seemed simpler and more efficient than implementing it in a separate function. Let me know what you think.

Brett208:
Hmm, I didn't mean to merge this branch but guess I did.

I just thought of an easy way to separate checking for duplicate internalFilenames from the function that creates the internalFilenames from the paths to be packed. I'll open a branch for this.

It is important to me to check for duplicate internalFilenames as I think allowing inclusion of duplicate filenames doesn't make sense when trying to retrieve files via the binary search. Duplicate filenames originating from different paths or the same path repeated twice should cause the packing to be aborted. 

I agree with moving code that checks for and strips .WAV extensions for CLM file inclusion into the OP2Utility code. I'll move the subroutine that does this from OP2Archive to OP2Utility in a new branch.

I'm not following your thought process on the security issue. I don't see how to check where a file will be unpacked to during the file packing process. The user will specify the directory they want the file unpacked to during the unpacking process and I wouldn't be able to predict that up front. I would prefere to leave it up to the  application or user calling the unpack function to choose a safe location where to unpack files.

Thanks,
Brett

DanRStevens:
Duplicate filenames won't harm a binary search, assuming sort order is maintained (identically compared items must be next to each other), but the search will only be able to find one of them, whichever one it hits first. I agree it should be a packing error to have two identically named files. Though with that said, if internal operations are done by index, rather than name, it would be possible to support identically named files, and still have access to both. At the very least, you'd be able to properly investigate any oddly or maliciously constructed files, where the game might see one, but unpacking it all would reveal the other (possibly by putting it later so it overwrites the malicious file when both are unpacked).

For the security issue, consider unpacking a VOL which contains any filenames such as the following:
/autoexec.bat
/config.sys
C:\Windows\System32\cmd.exe
/etc/passwd  (Linux example)
../../../../../Windows/System32/cmd.exe  (guess the relative depth)

If they were logged on as an admin, that could really mess up their system. Simply prefixing an output folder is not sufficient to default the relative path example. And if the empty string is used to represent the current working directory, which is common in many apps, the paths with a leading / will be treated as absolute rather than relative.

This is something you need to consider during extraction, since it's possible the archive was packed using a 3rd party tool that you don't control.

Brett208:
Good point about a volume being created by a third party.

When unpacking, we can add a check that throws an error if a path is found on any internalName before unpacking commences to catch this sort of thing.

PR #12, opened 2018-04-17
Sort ArchivePacker InternalNames Alphabetically

Brett208:
- Also clean code comments in ArchivePacker and ClmFile.

### DIFF
--- a/Archives/ArchivePacker.cpp
+++ b/Archives/ArchivePacker.cpp
@@ -66,4 +66,9 @@ namespace Archives
 
 		return fileNames;
 	}
+
+	bool ArchivePacker::ComparePathFilenames(const std::string path1, const std::string path2)
+	{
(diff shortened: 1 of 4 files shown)


Archives/ArchivePacker.cpp
```diff
@@ -66,4 +66,9 @@ namespace Archives
 
 		return fileNames;
 	}
+
+	bool ArchivePacker::ComparePathFilenames(const std::string path1, const std::string path2)
+	{
+		return StringHelper::StringCompareCaseInsensitive(XFile::GetFilename(path1), XFile::GetFilename(path2));
```
DanRStevens: I don't think XFile::GetFilename should be used here, as I don't think path components should be stripped from the filenames. It does actually make sense to allow VOL files to contain subfolders, much like ZIP files. The only real complication is that during extraction you might need to create folders before unpacking the files. Filename comparison would be done on the full path and filename. This of course is assuming relative paths are used, as absolute paths can be a security concern.

PR #13, opened 2018-04-18
Cleaning map data

Brett208:


OP2Maps/MapData.h
```diff
@@ -0,0 +1,61 @@
+#pragma once
+
+#include "MapHeader.h"
+#include "TileData.h"
+#include "TileSetSource.h"
+#include "TileInfo.h"
+#include "TerrainType.h"
+#include "ClipRect.h"
+#include "../StreamReader.h"
```
DanRStevens: Consider using a forward declare:
class FileStreamReader;
class SeekableStreamReader;
class StreamReader;

We don't actually need the full class definition, since we are only declaring a pointer to a class of this type. The compiler only needs to know the size of a pointer, which doesn't depend on the specifics of the class.

Using a forward declare instead avoids having to read an parse an extra header file (along with anything else that recursively includes), which can improve compile times, and reduce the amount of code that is rebuilt if the header file is edited.

OP2Maps/MapHeader.cpp
```diff
@@ -0,0 +1,8 @@
+#include "MapHeader.h"
+
+#include <cmath>
+
+unsigned int MapHeader::MapTileWidth()
+{
+	return (unsigned int)pow(2, lgMapTileWidth);
```
DanRStevens: You can avoid cmath using the formula:
tileWidth = 1 << lgTileWidth

The above formula compiles down to a single machine instruction. It's valid for twos complement integer operations, which is basically everything outside of floating point. It starts with 1 (a single low order bit set), and shifts it left by the specified number, effectively multiplying it by 2 to the power of that number. This is analogous to multiplying by powers of 10 by adding zeros to the right. A trick most programmers will be familiar with.

Using pow results in a call to a complicated looping function, resulting in more compiled code, and less efficiency. It's use is mainly for non-integer exponents. C++ does have overrides (which C does not support) for greater efficiency when the second parameter is an integer (a convenience for large powers), though it's still a general algorithm, and the compiler won't be able to optimize for the specifics of having a fixed base of 2. It's often advised to avoid using pow unless you need the full generality it offers.

OP2Maps/MapData.h
Brett208: I'm trying to avoid using the word map by itself here to keep away from using the same name as the standard library map container. Hence the class MapData and the folder OP2Maps. I'm looking at wrapping everything in the OP2Maps folder in an OP2Maps namespace. 

I'm certainly open to other naming conventions here if other things make more sense. Probably best to choose a good name now before Leeor starts integrating it though.

Good idea about the forward declare of the StreamReader class. I'll look at adding this soon.

DanRStevens:
I think `std::map` already protects sufficiently from name collisions, especially in the absence of `using` clauses.

In terms of namespaces, I think I'd prefer an entire API being wrapped in a namespace, rather than individual modules within the API. I think that's more inline with standard practice, plus it's probably wise not to get too carried away with namespaces. Namespaces are one of the weak points of C++, in that it's not really a proper module system.

DanRStevens:
I would recommend naming the folder after the main data structure in it. If that's `MapData`, I can live with that.

OP2Maps/MapData.h
Brett208: When attempting to forward delcare the three StreamReader classes, I am getting a lot of the following error:

C2027 use of undefined type ‘StreamReader’

For example: Stream Reader here is throwing the error:

void MapData::ReadMapHeader(StreamReader& streamReader)
{
	streamReader.Read((char*)&mapHeader, sizeof(mapHeader));
}

Either I'm missing something about Forward Declares or this isn't a valid use of them???

DanRStevens: If that is coming from the .cpp file, then add the `#include` there. You will very likely need a full include from the .cpp file.

PR #14, opened 2018-04-20
Strip file extensions from internalNames of files to be packed in a CLM file

Brett208:


Brett208:
When packing, the ClmFile class will reduce the character count of an internalName to 8 if the internalName is presented at more than 8 characters. This allows the file to be packed successfully. 

We could abort the packing if the internalName is greater than 8 characters, but I am happy with the current state of affairs.

DanRStevens:
Yes, you're right. It copies up to a maximum of 8 characters. It silently truncates if the name is longer. Might be good to have a warning or error in that case, though technically that is a new feature, rather than something tied to this pull request.

PR #15, opened 2018-04-20
Improve duplicate internal name checks

Brett208:
This is my first time creating a branch from a branch already containing an outstanding merge request. So this is a bit of an experiment on my part.

Solving this issue does not directly rely on the previous branch to work, but it does touch the codebase in the same places.


DanRStevens:
Also, for what it's worth, I've also been wondering how to best handle branches of branches. In the case of unrelated changes, I'm uncertain if a branch of a branch makes the most sense, or if it should be based off of master and just handle merge conflicts when the time comes.

By default, the diff contains changes I'd already reviewed from the other branch. GitHub does help a little here, in that you can see a diff of only certain commits. For a small case like this, it doesn't seem to matter too much, though it was good to know about the branch of a branch, and that I'd probably already reviewed some of the code.

Brett208:
I would prefer not to implement this search assuming that the list has been sorted. This makes the code a bit more brittle by forcing the functions to be called in a specific order that may not be apparent to the programmer making changes. 

This function is called one time when a file packing occurs, so efficiency doesn't really matter.

However, if efficiency is the goal, we should consider deleting this branch and using the orignal function that checked for duplicates while creating the internal Filename. This eliminates the second function altogether. 

```
std::vector<std::string> ArchivePacker::GetInternalNamesFromPaths(std::vector<std::string> paths)
{
	std::vector<std::string> fileNames;

	for (std::string fileName : paths) {
		if (StringHelper::ContainsStringCaseInsensitive(fileNames, fileName)) {
			throw std::runtime_error("Unable to create an archive containing multiple files with the same filename.");
		}
		fileNames.push_back(XFile::GetFilename(fileName));
	}

	return fileNames;
}
```
If a different sorting algorithm is used, I would prefer it still consider the sort case insensitive.

---

This merge request contains both the commit involved in Duplicate Internal Name checks and CLM stripping filenames. Because of this, I'm guessing that if we decided to merge this branch before the first branch, we could not as both commits are contained in this branch.

So in this situation I'm thinking a rule of thumb is to only create a branch of a branch if the second branch CANNOT be implemented without the prescribed features of the first branch. Just because the commits involve similar parts of the code base was a poor reason to create a branch off of the first branch.

DanRStevens:
Making assumptions such as requiring sorted data should of course be documented in the code. It can also be included as part of the function name to make sure it is always clear, such as `VerifySortedVectorHasNoDuplicates` or similar. Such a function could raise an exception if a duplicate is found.

As this is an internal class detail, it should have little impact on code brittleness.

For correctness reasons, the comparison used for duplicate checking should be the same as the comparison used for the sorting. Ideally this should be enforced by calling the same function in both places.

As for the sort order, that is largely determined by the file format spec. Hence case insensitive.

Considering efficiency, the check as listed here is an O(n^2) algorithm. For a large number of files, it will be slow. It's usually good practice to keep algorithms on unbounded input sizes below O(n^2) when possible. You can of course claim it shouldn't matter in this case, as packing enough files for this to be an issue is unlikely. On the other hand, the sorting is a O(n*lg(n)) algorithm, and checking a sorted list for duplicates is a O(n) algorithm. Both are likely to perform well even for very large inputs.

As for the branch, perhaps it can be rebased after the upstream branch is merged. I'm not too sure how that works after a pull request has been opened. I'm willing to explore though.

Brett208:
Dan, check out the new function. It is now implemented as O(n).

`void ArchviePacker::CheckSortedContainerForDuplicateNames(const std::vector<std::string>& internalNames);`

I rebased the changes in this branch before commiting. This caused a merge commit. This somehow managed to pull a bunch of other comments into this merge request. I'm not sure what I did wrong in the process. 

-Brett

DanRStevens:
The new function looks sensible and efficient, and clearly named.

Not sure what happened with the rebase. I assume you rebased from the Git client, rather than GitHub. Might be interesting to check a graph of the history. I'm also uncertain how GitHub handles pull requests when the branch is rebased after the pull request is opened. Are there any differences between your local client and what GitHub displays? Might be worth starting a thread on OPU about the subject.

PR #19, opened 2018-04-23
Add op2 map writer and refactor OP2 Map code

Brett208:
I'd like to continue working on and improving the OP2 Map code, but these changes are significant enough that a merge into master would be smart.


Brett208:
No worries. I already know what caused the conflicts and figured it was coming before pushing. I've got some experience fixing merge conflicts in Mercurial and want to try it out here. It is more being caused by working in MapHeader.h in 2 different branches and deleting MapHeader.cpp in the master branch while it exists in this branch than moving the files. However, you are right it would have been cleaner if I had separated renaming the directory into a separate commit.

PR #20, opened 2018-04-24
Add CellType scoped enum to Maps

Brett208:
- Decided to use a scoped enum to limit the number of individual items added to the working namespace when included.
 - Mirrors the enum CellTypes in Outpost2DLL.

** This is a good candidate for Squash and Merge **

### DIFF
--- a/Maps/CellType.h
+++ b/Maps/CellType.h
@@ -1,0 +1,38 @@
+#pragma once
+
+// CellTypes returned and set by the GameMap class
+enum class CellType
+{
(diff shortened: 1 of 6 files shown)


DanRStevens:
This looks quite reasonable, and enables additional type checking by the compiler. I rather like the result.

By squash and merge, I assume you mean a fast-forward merge, with no merge commit. As the branch contains only 1 commit, there is really nothing to squash. Both squash and rebase should have the same result here.

PR #23, opened 2018-04-27
Refactor StreamReader class series

Brett208:
- Change SeekableStreamReader::Ignore(size_t size) to be Seek(int offset). This allows for seeking both forward and backwards in a memoryStream or fileStream.
 - Add bounds check to ensure MemoryStreamReader does not allow offsetting to a negative value (before the start of the actual stream).

- This puts StreamReader in a closer position for integration as a Serializer if desired in the future.

### DIFF
--- a/StreamReader.cpp
+++ b/StreamReader.cpp
@@ -4,16 +4,14 @@
 #include <cstring>
 #include <stdexcept>
 
-using namespace std;
-
 StreamReader::~StreamReader() { }
 
 // Defers calls to C++ standard library methods
 FileStreamReader::FileStreamReader(std::string filename) {
-	file = ifstream(filename, ios::in | ios::binary);
+	file = std::ifstream(filename, std::ios::in | std::ios::binary);
 
 	if (!file.is_open()) {
-		throw runtime_error("File could not be opened.");
+		throw std::runtime_error("File could not be opened.");
 	}
 }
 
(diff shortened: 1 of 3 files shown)


StreamReader.cpp
```diff
@@ -39,18 +37,18 @@ MemoryStreamReader::MemoryStreamReader(char* buffer, size_t size) {
...
 	memcpy(buffer, streamBuffer + offset, size);
 	offset += size;
 }
 
-void MemoryStreamReader::Ignore(size_t size)
+void MemoryStreamReader::Seek(int offset)
 {
-	if (offset + size > streamSize) {
-		throw runtime_error("Size of bytes to ignore exceeds remaining size of buffer.");
+	if (this->offset + offset > streamSize || this->offset + offset < 0) {
```
DanRStevens: The `size_t` type is unsigned. When you mix signed and unsigned types, the signed type is converted to unsigned. Hence the result of the addition is unsigned. An unsigned type can never be `< 0`. The comparison against `streamSize` should be enough to catch both overflow and underflow.

Brett208: So, in this case, would it set size_t this->offset to 0 if a subtraction is performed that would otherwise take a signed number into the negative range?

I think even if protection is provided from reading outside the stream, it would be nice to inform the user they are attempting to read outside the bounds as this is probably not their intention. Either way it is currently broken as you mentioned.

Thanks.

DanRStevens: Unsigned numbers have well defined wrap around. Instead of going negative, it will turn into a large positive, and be caught by the first condition.

Issue #22, opened 2018-04-26
Allow controls for how FileStreamWriter manipulates files


Brett208:
Currently FileStreamWriter implies overwriting an existing file if it exists and assumes file being written is in binary format. Instead, I would like to add a BitField that allows specifying how a file being written to is manipulated.

Maybe:
```
enum class FileWriteFlag
{
    overwrite = 0b0000'0001, // Allows overwriting existing file
    append = 0b0000'0010, // Allows appending information to existing file
    binary = 0b0000'0100 // Specifies binary file instead of text file
}
```
Default assumption of FileStreamWriter will be to abort writing if file exists and to write in text format. I think we will mostly be using binary, but the fstream class assumes text as default (I think) so this will keep syntax close to standard library.

DanRStevens:
Good idea to add such a parameter. Could probably specify a default parameter value. Unless of course you want the programmer to really think about what they want, and force them to be explicit.

Might be worth checking documentation for other systems, such as the Windows API [CreateFile](https://msdn.microsoft.com/en-us/library/windows/desktop/aa363858(v=vs.85).aspx), to see what other flags they offer. There might be a more sensible set of flags than whatever the C API decided upon 40+ years ago. In particular, the `dwCreationDisposition` parameter is worth thinking about.

You make a very good point about being compatible with existing practice concerning text mode. It is best to avoid surprises. Though I guess that's one reason why I've traditionally hated text mode. It makes often surprising conversions to data. I've long felt the text mode default was a mistake of older APIs. Text mode corrupts binary data, but binary mode doesn't corrupt text data. Plus, it feels like the conversion is being done in the wrong place by making it a part of the file API like that, rather than a higher level layer. Interestingly, the Windows API has no concept of a text file.

DanRStevens:
I was just taking a look back at this. We've had PR #25 open for about 3 months now. Perhaps it's time to revisit. Hopefully we have a better sense of what we need here now.

In terms of support, I'd like to be able to do the following for opening a FileStreamWriter:
Open a file whether it existed before or needs to be created new (default)
Fail if the file already exists
Fail if the file doesn't exist (maybe?)
Truncate existing data to length 0
Do not truncate existing data
Start writing at either the beginning (default) or at the end (append)

[...]

PR #25, opened 2018-04-29
Add FileStreamWriter control flags

Brett208:
Add an enum called FileWriteFlag that allows manipulating file write settings. Based on options provided in Windows by dwCreationDisposition (https://msdn.microsoft.com/en-us/library/windows/desktop/aa363858(v=vs.85).aspx) This enum is mapped to appropriate flags for manipulating std::fstream.

There is a helper function FileStreamWriter::FlagSet. It is very generic though and could possibly live somewhere else. Since it is only one function, I didn't want to create a separate header file for it right now though.

Usability could possibly be approved right now by adding more flags. For example, if you want to append data to the file, there is no option. You have to recognize that appending is the default value and just not set Truncate. The tradeoff here is we are adding unnecessary flags and then I'll have to increase the codebase to handle them.

Current default settings are create files if they do not exist, append data if they do exist, and write in binary as default. Writing in binary is not default for std::fstream, but it seemed to make more sense as a default for our uses.

Thanks.

### DIFF
--- a/FileWriteFlags.h
+++ b/FileWriteFlags.h
@@ -1,0 +1,24 @@
+#pragma once
+
+namespace FileWriteFlag
+{
+	// Default values are Do not overwrite files, append to existing files, and write in binary format.
+	enum FileWriteFlag : unsigned int
+	{
+		// SET:     Allow overwritting the file if it exists (mutually exclusive from Append).
+		// NOT SET: Do not allow overwritting an exisitng file.
+		Overwrite = 0b0000'0001,
+
+		// SET:     Do not allow a new file to be created (Mutually exclusive from Overwrite).
+		// NOT SET: Allow new file creation. Overwrite must also be set.
+		DoNotCreate = 0b0000'0010,
+
+		// SET:     If file exists, default to truncating file to zero bytes
+		// NOT SET: If file exists, default to writing at end of file (Append).
+		Truncate = 0b0000'0100,
+
+		// SET:     Specifies text file format.
+		// NOT SET: Specified binary file format.
+		Text = 0b0000'1000
+	};
+}
--- a/StreamWriter.cpp
+++ b/StreamWriter.cpp
@@ -1,11 +1,24 @@
 #include "StreamWriter.h"
+#include "XFile.h"
 #include <stdexcept>
 
 StreamWriter::~StreamWriter() { }
 
-FileStreamWriter::FileStreamWriter(const std::string& filename)
+FileStreamWriter::FileStreamWriter(const std::string& filename, unsigned int writeSettings)
 {
-	fileStream.open(filename.c_str(), std::ios::trunc | std::ios::out | std::ios::binary);
+	if (!FlagSet(writeSettings, FileWriteFlag::Overwrite)) {
+		if (XFile::PathExists(filename)) {
+			throw std::runtime_error("File already exists at provided path: " + filename + ".");
+		}
+	}
+	
+	if (FlagSet(writeSettings, FileWriteFlag::DoNotCreate)) {
+		if (!XFile::PathExists(filename)) {
+			throw std::runtime_error("File does not exist at provided path: " + filename + ".");
+		}
+	}
+
+	fileStream.open(filename.c_str(), FormatFstreamMode(writeSettings));
 
 	if (!fileStream.is_open()) {
 		throw std::runtime_error("File could not be opened.");
@@ -51,4 +64,28 @@ void MemoryStreamWriter::Seek(int offset)
 	}
 
 	this->offset += offset;
+}
+
+// Transfer the writeSettings bitfield into an fstream _Mode bitfield 
+unsigned int FileStreamWriter::FormatFstreamMode(unsigned int writeSettings) const
+{
+	unsigned int fstreamMode = std::ios::out;
+
+	if (FlagSet(writeSettings, FileWriteFlag::Truncate)) {
+		fstreamMode |= std::ios::trunc;
+	}
+	else {
+		fstreamMode |= std::ios::app;
+	}
+
+	if (!FlagSet(writeSettings, FileWriteFlag::Text)) {
+		fstreamMode |= std::ios::binary;
+	}
+
+	return fstreamMode;
+}
+
+bool FileStreamWriter::FlagSet(unsigned int bitfield, unsigned int flag) const
+{
+	return ((bitfield & flag) == flag);
 }
--- a/Maps/MapWriter.cpp
+++ b/Maps/MapWriter.cpp
@@ -19,9 +19,10 @@ void MapWriter::Write(SeekableStreamWriter& mapStream, const MapData& mapData)
 	WriteTileGroups(mapData.tileGroups);
 }
 
+// Creates a new map file by truncating an existing file if it exists.
 void MapWriter::Write(const std::string& filename, const MapData& mapData)
 {
-	Write(FileStreamWriter(filename), mapData);
+	Write(FileStreamWriter(filename, FileWriteFlag::Overwrite | FileWriteFlag::Truncate), mapData);
 }
 
 void MapWriter::WriteHeader(const MapHeader& header)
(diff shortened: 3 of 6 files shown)


Brett208:
resolves #22

FileWriteFlags.h
```diff
@@ -0,0 +1,24 @@
+#pragma once
+
+namespace FileWriteFlag
```
DanRStevens: Do these flags apply to all file streams, or only write streams?

FileWriteFlags.h
```diff
@@ -0,0 +1,24 @@
...
+{
+	// Default values are Do not overwrite files, append to existing files, and write in binary format.
+	enum FileWriteFlag : unsigned int
```
DanRStevens: Consider using an enum class instead of the namespace wrapper

FileWriteFlags.h
```diff
@@ -0,0 +1,24 @@
...
+	{
+		// SET:     Allow overwritting the file if it exists (mutually exclusive from Append).
```
DanRStevens: I'm not sure these are really mutually exclusive. Append just sets the current position to the end of the file. This might not relate to read/write permissions, though usually you'd expect write permission in that case.

FileWriteFlags.h
```diff
@@ -0,0 +1,24 @@
...
+		// NOT SET: Do not allow overwritting an exisitng file.
+		Overwrite = 0b0000'0001,
+
+		// SET:     Do not allow a new file to be created (Mutually exclusive from Overwrite).
+		// NOT SET: Allow new file creation. Overwrite must also be set.
+		DoNotCreate = 0b0000'0010,
```
DanRStevens: A `Create` flag might be a little more straightforward. A write stream can default to having this flag set. A read stream would not.

FileWriteFlags.h
```diff
@@ -0,0 +1,24 @@
...
+
+		// SET:     If file exists, default to truncating file to zero bytes
+		// NOT SET: If file exists, default to writing at end of file (Append).
```
DanRStevens: This doesn't seem like the opposite of Append, particularly for read streams, where you'd almost certainly not want to truncate, nor append. For a write stream, you might want to open an existing file, preserving the contents, yet still start to overwrite from the beginning.

FileWriteFlags.h
```diff
@@ -0,0 +1,24 @@
+#pragma once
+
+namespace FileWriteFlag
+{
+	// Default values are Do not overwrite files, append to existing files, and write in binary format.
+	enum FileWriteFlag : unsigned int
+	{
+		// SET:     Allow overwritting the file if it exists (mutually exclusive from Append).
+		// NOT SET: Do not allow overwritting an exisitng file.
+		Overwrite = 0b0000'0001,
```
DanRStevens: Is this maybe just asking for Write permission? I don't believe this is a simple portable way to distinguish between write access for new data (being appended) versus overwriting of existing data.

StreamWriter.cpp
```diff
@@ -1,11 +1,24 @@
...
+#include "XFile.h"
 #include <stdexcept>
 
 StreamWriter::~StreamWriter() { }
 
-FileStreamWriter::FileStreamWriter(const std::string& filename)
+FileStreamWriter::FileStreamWriter(const std::string& filename, unsigned int writeSettings)
 {
-	fileStream.open(filename.c_str(), std::ios::trunc | std::ios::out | std::ios::binary);
+	if (!FlagSet(writeSettings, FileWriteFlag::Overwrite)) {
```
DanRStevens: When you pre-check conditions before actually performing the operation, you run the risk of race conditions. There may be other threads and other processes concurrently accessing and modifying the filesystem. What you're checking may not be what you later operate on. There have actually been security vulnerabilities that resulted from such race conditions. It would be more advisable to translate the set of API flags to a set of system flags, and rely on the in-built checks of the underlying system. The underlying system API should be able to atomically check permissions while opening the file.

FileWriteFlags.h
Brett208: If I use enum class, then I have to perform static_cast in order to do bitwise operations:

Write(FileStreamWriter(filename, static_cast<unsigned int>(FileWriteFlag::Overwrite) | static_cast<unsigned int>(FileWriteFlag::Truncate)), mapData);

Associated error if static_cast is missing: E0349 No operator "|" matches these operands.

This seems too unwieldy to me. If I treat them as an older enum but wrap them in a namespace, I don't have to perform the static_cast but I'm also not placing each member of the enum in the global namespace.

Why are you saying to consider a enum class here?

FileWriteFlags.h
Brett208: I thought the name FileWriteFlag indicated fairly clearly indicate the flags only apply to FileWriteStreams. Although I'm guessing this was not intuitive to you since you asked the question. I'm not sure how to change the name to more clearly indicate this.

FileWriteFlags.h
Brett208: I suppose you could manually move the position to the beginning of the file and start overwriting the file after selecting Append. I'm not sure why this would be done in practice though? I can remove the mutually exclusive part from the comment.

FileWriteFlags.h
Brett208: Maybe this is a dumb question, but how do I setup the program to have a flag default setting to ON? The reason I chose DoNotCreate is because the default value of 0 would then allow creating a new file. 

These flags are only designed for manipulating a File during Write situations.

Brett208:
Dan, You responded negatively to 3 of my 4 choices for file write flags and the type I used for implementing them (non-scoped enum wrapped in a namespace). I'm betting that I will just flounder in meeting your expectations on a rewrite. Perhaps you could outline a set of flags and their descriptions that you would be content with so I'm not guessing.

-Brett

FileWriteFlags.h
DanRStevens: Hmm, good point. I hadn't thought of that.

There may be a workaround though:
http://blog.bitwigglers.org/using-enum-classes-as-type-safe-bitmasks/

StreamWriter.cpp
DanRStevens: Hmm, good point. That does seem like a rather limited selection. I guess I expected more options to be available under the hood.

FileWriteFlags.h
DanRStevens: I suppose I made a bad assumption there based on a lot of other systems that have a single set of file flags for both read and write. Technically there is no reason why you should need to have a single set. At least not for streams which will only ever be read, or write, but not both. I guess the typical case with most existing APIs is to allow both reading and writing, hence a single set of flags.

FileWriteFlags.h
DanRStevens: In the method that opens the file, you can set a default parameter value for the flags parameter. I realize the bit sensibleness of what you've done. I've done it plenty of times myself, sometimes with worse names. Though, I do find negatively named bit flags tend to be confusing. What about `OpenExisting`?

Brett208:
It just occurred to me we could just allow passing the bitfield std::ios_base::openmode (http://www.cplusplus.com/reference/ios/ios_base/openmode/) into both FileStreamWriter and FileStreamReader. Both classes just wrap fstream anyways and this would allow some flexibility in how the file is written/read without reinventing anything in a way familiar to others. 

We can come back later and change to a more custom solution later if the library develops far enough and a need arises.  

If this doesn't sound right to you, then lets punt and talk offline at some point like you suggested. I'm not attached to fstream but I also don't know that right now I want to shop around for a different cross platform solution/homeroll our own.

Thanks,
Brett

DanRStevens:
That would be easier to just pass in `std::ios_base::openmode`, but that creates a leaky abstraction. Internal implementation details become exposed in the method parameter. If someday the internal implementation changes away from `fstream`, there's a good chance you'd also need to change the parameter, which would break all callers of the code.

Brett208:
Agreed on closing it out now.

PR #27, opened 2018-05-04
Allow adding and removing tilesets to existing maps

Brett208:
- Operation is aborted if tileset is contained within a TerrainType.
 - Removes references to tileset from TileGroups and the current map by setting indices to 0.

** This functionality will eventually need to be expanded for dealing with TerrainTypes.

### DIFF
--- a/Maps/MapData.cpp
+++ b/Maps/MapData.cpp
@@ -1,4 +1,6 @@
 #include "MapData.h"
+#include "../XFile.h"
+#include <stdexcept>
 
 unsigned int MapData::GetTileInfoIndex(unsigned int x, unsigned int y) const
 {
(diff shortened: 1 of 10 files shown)


Brett208:
I resolved the merge conflicts within GitHub and it only added a single commit in the process, which is good. I found GitHub's merge conflict resolution system intuitive. The only downside is that I cannot compile and check on the results before pushing the merge to the public. 

I will try to resolve through TortoiseGit for the next Merge Conflict and see how it goes again.

Brett208:
I'm considering never merging this branch into master. A system that gracefully handles replacing tilesets needs to take into account the contents of TerrainRange, which may look a lot different than this solution. 

I would like to leave this branch in place in case it is helpful to someone in the future.

PR #31, opened 2018-05-17
Refactor Function ClmFile::WriteVolume

Brett208:
- Replace char[] with std::array<char, X>
 - Remove S (struct) prefix from ClmHeader name
 - Create 2 subroutines to reduce size of function (PackFile & PrepareIndex)
 - Add brackets to one line if statements
 - Move variable initialization to first use instead of at top of function

### DIFF
--- a/Archives/ClmFile.cpp
+++ b/Archives/ClmFile.cpp
@@ -5,11 +5,8 @@
 #include <stdexcept>
 #include <algorithm>
 
-
 namespace Archives
 {
-	const unsigned int CLM_WRITE_SIZE = 0x00020000;
-
 	const int RIFF = 0x46464952;	// "RIFF"
 	const int WAVE = 0x45564157;	// "WAVE"
 	const int FMT  = 0x20746D66;	// "fmt "
@@ -289,7 +286,7 @@ namespace Archives
 		CheckSortedContainerForDuplicateNames(internalFileNames);
 
 		// Write the volume header and copy files into the volume
-		if (!WriteVolume(outFile, filesToPack.size(), fileHandle, indexEntry, internalFileNames, waveFormat))
+		if (!WriteVolume(outFile, fileHandle, indexEntry, internalFileNames, waveFormat))
 		{
 			// Error writing volume file
 			CleanUpVolumeCreate(outFile, filesToPack.size(), fileHandle, waveFormat, indexEntry);
@@ -432,12 +429,10 @@ namespace Archives
 		WaveFormatEx *waveFormat,
 		IndexEntry *indexEntry)
 	{
-		int i;
-
 		// Close the output file
 		if (outFile) CloseHandle(outFile);
 		// Close all open input files
-		for (i = 0; i < numFilesToPack; i++) {
+		for (int i = 0; i < numFilesToPack; i++) {
 			CloseHandle(fileHandle[i]);
 		}
 		// Free temporary memory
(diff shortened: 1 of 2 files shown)


Brett208:
This branch is meant to refactor the code before taking steps to actually remove the Windows specific HANDLE code.

Couple of questions:

1. Is it appropriate to pass a HANDLE as an argument without using the Address of Operator (&)? I think it is because a HANDLE is basically a pointer and shouldn't require much when creating a copy.

2. When using std::array, I found 2 ways to pass information into the std::array when reading data. Either using the member function data() or using the Address of Operator (&). Both methods worked. The documentation I reviewed used the member function data(), so I went with that approach. However, it didn't seem to mention using the Address of Operator, so not sure if these are considered equivalent or if there is a clear reason to use the member function data() over the Address of Operator.

```
// Using member function data
if (ReadFile(fileHandle, buffer.data(), numBytes, &numBytesRead, nullptr) == 0) {
	return false;
}

// Using address of operator
if (ReadFile(fileHandle, &buffer, numBytes, &numBytesRead, nullptr) == 0) {
	return false;
}
```

3. I'd like to provide some documentation on what is going on with the do loop now contained in the function ClmFile::PackFile but I'm actually not exactly sure what is going on myself.

Archives/ClmFile.cpp
```diff
@@ -463,97 +458,94 @@ namespace Archives
...
+		}
+	}
+
+	bool ClmFile::PackFile(HANDLE outFile, std::array<char, CLM_WRITE_SIZE>& buffer, unsigned long& numBytesRead, const IndexEntry& entry, HANDLE fileHandle)
+	{
+		int offset = 0;
+		do
+		{
+			unsigned long numBytes = CLM_WRITE_SIZE;
+			if (offset + numBytes > entry.dataLength) {
```
DanRStevens: This should be commented. I wasn't able to remember easily what it was doing. This is clipping the size of the data transfer so it doesn't exceed the remaining number of bytes in the input file.

Archives/ClmFile.cpp
```diff
@@ -463,97 +458,94 @@ namespace Archives
...
+		{
+			// Copy the filename into the entry
+			strncpy((char*)&entry[i].fileName, internalNames[i].c_str(), 8);
+			// Set the offset of the file
+			entry[i].dataOffset = offset;
+			offset += entry[i].dataLength;
+		}
+	}
+
+	bool ClmFile::PackFile(HANDLE outFile, std::array<char, CLM_WRITE_SIZE>& buffer, unsigned long& numBytesRead, const IndexEntry& entry, HANDLE fileHandle)
```
DanRStevens: I'm wondering if this method should take the buffer as a parameter, or just declare it locally. There may be some good reasons either way.

Archives/ClmFile.cpp
```diff
@@ -463,97 +458,94 @@ namespace Archives
...
-			offset += entry[i].dataLength;
+		unsigned long numBytes;
+		if (WriteFile(outFile, &header, sizeof(header), &numBytes, nullptr) == 0) {
+			return false;
 		}
 
-		// Write the index
-		if (WriteFile(outFile, entry, numFilesToPack * sizeof(IndexEntry), &numBytes, nullptr) == 0)
-		{
+		PrepareIndex(sizeof(header), internalNames, entry);
```
DanRStevens: Perhaps this method should return a vector of IndexEntry structs, rather than take an array to fill.

DanRStevens:
As for your questions:

1) Yes, you can and probably should pass a HANDLE without using &. A HANDLE, and other similar values are typically an [Opaque Pointer](https://en.wikipedia.org/wiki/Opaque_pointer). As such, it's already a pointer, and so using & would give you a pointer to a pointer. That would make sense for out parameters, but is unnecessary for in parameters.

2) I checked the documentation for [std::array](http://en.cppreference.com/w/cpp/container/array). It seems `data()` is the obvious documented and supported way to do what you're doing. Though reading between the lines, it sounds like either way should be correct:

> This container is an aggregate type with the same semantics as a struct holding a C-style array T[N] as its only non-static data member.

That's basically saying you won't encounter any hidden fields by accident, which might otherwise have been what & points to.

3) Heh, I had trouble with that too just now. Each iteration of the do loop will copy a buffer size chunk of data from the input to the output, with clipping to handle the last chunk of the file, which may be smaller than a full buffer.

Archives/ClmFile.cpp
```diff
@@ -5,11 +5,8 @@
 #include <stdexcept>
 #include <algorithm>
 
-
 namespace Archives
 {
-	const unsigned int CLM_WRITE_SIZE = 0x00020000;
-
 	const int RIFF = 0x46464952;	// "RIFF"
```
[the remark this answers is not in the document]
Brett208: I had to move CLM_WRITE_SIZE into the header because it was required in the function prototype of ClmFile::PackFile. I did this to reduce the number of times a new buffer was initialized. It sounds like you wouldn't mind if the buffer was recreated each time PackFile was called. I like this idea better because it simplifies the function prototype of PackFile and allows me to move CLM_WRITE_SIZE back into ClmFile.cpp. I'll go ahead and push this change soon.

I would like to make it read:
```
const uint32_t CLM_WRITE_SIZE = 0x00020000;
const uint32_t RIFF = 0x46464952;	// "RIFF"
const uint32_t WAVE = 0x45564157;	// "WAVE"
const uint32_t FMT  = 0x20746D66;	// "fmt "
const uint32_t DATA = 0x61746164;	// "data"
```

I just don't know if an unsigned value is acceptable for RIFF, WAVE, FMT, and DATA. I'm trying to set all values that will be involved in serialization to a fixed size type to eliminate possible differences between compilers when we get everything cross-platform.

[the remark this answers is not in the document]
DanRStevens: Ahh, that makes more sense now.

The RIFF, WAVE, FMT, and DATA values are written to the file as a fixed size, so yes, it makes sense for them to be a fixed sized type, such as uint32_t. Unsigned does make slightly more sense than signed, though it's just a 32-bit ID. I suppose more correct is to have 4 consecutive character codes, making a human readable tag, though I'd only do that if there was a simple and easy to read initialization syntax. Maybe the newer language extensions, or `std::array<char>` better support such initialization.

The CLM_WRITE_SIZE is an array size, so size_t might make the most sense there. That value isn't written to the file as far as I remember. Though I think you pointed out a field that had that value stored in it, so maybe the two should be tied together.

The file copy buffer doesn't need to be fully initialized, since the copy reads data into it first, and then only writes out what was read. However, large arrays on the stack do need to have each page touched to ensure the OS grows the stack properly, and doesn't just cause access violations. Hence, large arrays on the stack need to have at least 1 byte (generally 4 bytes) out of every 4096 bytes touched, which might just be a dummy read. The compiler automatically inserts this code in the function prolog. Hence there is a minor startup delay for functions with a large number of local variables.

Archives/ClmFile.cpp
Brett208: I've been thinking about this as well. If we plan on only opening one fileStream at a time when writing a ClmFile, we will not be able to set all values of IndexEntry at once. 

We will need to write default values into the ClmFile initially. Then as each file is added, seek back to the header for setting the actual values of each IndexEntry. 

So, I think what I'd like to see is expanding ClmHeader to include the IndexEntry container and look something like:

```
struct ClmHeader
{
    //TODO: Allow setting default values if ClmFile doesn't exist yet.

    // TODO: Check that index is > 0 & less than packedFilesCount
    InsertIndexEntry(int index, IndexEntry entry);

    std::array<char, 32> fileVersion; 
    WaveFormatEx waveFormat; 
    std::array<char, 6> unknown { 0, 0, 0, 0, 1, 0 }; 
    int packedFilesCount; 
    std::vector<IndexEntry> index;
}
```

Brett208:
Thanks for the research on all my questions. I'll keep using `std::array::data` then since it seems to be the way that documentation recommends accessing the underlying array data.

Okay, I understand now that we are packing a file by passing a discrete sized buffer of size CLM_WRITE_SIZE. I'm not understanding why we don't just pass the entire file into a buffer the size of IndexEntry::dataLength. Is this to reduce the amount of data in memory at a given time?

Thanks,
Brett

Archives/ClmFile.cpp
```diff
@@ -463,97 +458,94 @@ namespace Archives
...
-		std::vector<std::string> internalNames,
+		const std::vector<std::string>& internalNames,
 		WaveFormatEx *waveFormat)
 	{
 #pragma pack(push, 1)
-		struct SClmHeader
+		struct ClmHeader
 		{
-			char textBuff[32];
+			ClmHeader() { memcpy(fileVersion.data(), "OP2 Clump File Version 1.0\x01A\0\0\0\0", 32); }
```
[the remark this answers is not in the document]
DanRStevens: Hmm, the 2 constructor option is probably the best idea. That should allow flexibility and efficiency. The Reset option seems like it would be a bit awkward in practice.

I'm also fine with ignoring this concern for now. If there's no problem, there's really nothing to fix. I think my bigger concern here, is how this might affect long term maintainability. Your two constructor proposal is probably the best there, though the currently solution is probably just fine.

Archives/ClmFile.cpp
DanRStevens: We only need to know the file's name and length for the index entries. Nothing about the file's contents is required. Hence we don't actually need to open any of the files to build the index table.

I'd also like to avoid the whole can of worms of modifying a file in place by inserting new index entries. If the file is tightly packed, data will need to be moved to make room for extra index entries. That's not an efficient operation. Plus, if the program crashes, or there is a power outage during this process, and the modification was done in place, it's likely to leave the file corrupted.

DanRStevens:
Correct. Fixed memory usage, as opposed to unbounded memory usage.

Fixed memory usage provides a more consistent performance experience. Using too little buffer creates a lot of overhead and makes things slow. Using too much buffer (more than physical RAM, but less than virtual RAM) can potentially cause thrashing of the page file, and make things really really slow. Though that's getting harder to do on today's systems, assuming you're not packing DVD images into archives or similar. Exceeding virtual memory would just plain cause the program to crash, effectively creating an upper limit on the size of files you could support. Hence, a fixed sized buffer and chunked copies is generally a good idea.

For good performance, make sure the buffer is reasonable large to avoid excessive overhead, and a multiple of both the page size (traditionally 4K) and sector size (traditionally 512 bytes, though newer volumes use 4K).

Archives/ClmFile.cpp
Brett208: I think if we force the user to specify the number of and locations of all files to be packed before beginning to pack, there shouldn't be any need to move data, just fill in default index values that were already reserved in the file. I was planning on determining the size of each file after the stream was opened. If there is a way to determine this without opening a stream against the file, it could simplify the process. 

We would just want to check to ensure the file still exists and is the expected size when the file is actually opened for packing.

However this goes, it will all be done on a different branch than this one though.

DanRStevens: There are two ways I know of to [determine file size](https://stackoverflow.com/questions/5840148/how-can-i-get-a-files-size-in-c) without having to open the file. Either using <sys/stat> or <experimental/filesystem>. The filesystem method looks like the cleanest, by far, so I'll be happy when the experimental part is dead and buried.

DanRStevens: For an extra example of using stat, see [How to get filesize using stat](https://techoverflow.net/2013/08/21/how-to-get-filesize-using-stat-in-cc/).

Issue #24, opened 2018-04-28
SeekableStream Enhancements


DanRStevens:
Seekable streams should have a well defined position which can be both read and written. It also makes sense they would have a well defined length, as otherwise you could not bounds check a seek.

A common feature in seeking is to support both absolute and relative offsets. Some APIs support this through an extra parameter to the seek, though in the vast majority of cases, such a parameter ends up being a constant. I think a better design is to provide two separate methohds Seek and SeekRelative.

Further, large streams, such as files exceeding 4GB should be supported, even on 32-bit builds. As such, stream positions should be at minimum 64 bits.

As such, I propose SeekableStream should have the following methods:

```
	// Get the size of the stream
	virtual uint64_t Length() = 0;
	// Get the current position, from the start of the stream
	virtual uint64_t Position() = 0;
	// Seek to absolute position, given as offset from beginning of stream
	virtual void Seek(uint64_t position) = 0;
	// Seek ahead by a relative amount, given as offset from current position
	virtual void SeekRelative(uint64_t offset) = 0;
```

Brett208:
What do you think about allowing the user to seek backwards relative by making SeekRelative a int64_t?

I noticed std::fstream allows seeking backwards. Although I'm currently struggling to come up with a reason I would want to seek backwards. (I guess you could still seek backwards using Seek instead of SeekRelative in this construct.)

Is this an enhancement you are looking to tackle or having me work on?

Thanks.
-Brett

DanRStevens:
Seeking backwards makes sense. Can be useful if you peek at data to make a decision, such as reading a header marker to determine file or data type, and selecting an appropriate parser. The parser would most likely try to read and verify the marker itself.

Seeking backwards is more intuitive with an `int64_t` as opposed to a `uint64_t`, though unsigned values do have well defined wrap around, so it would be possible (and identical) either way. Signed values have less well defined wrap around. Though I could go either way for SeekRelative.

I can probably work on this enhancement. Though I noticed my use of Seek and SeekRelative is opposite of yours. If I just blindly implemented as suggested, it would break existing code.

Brett208:
I have no problem with what I am calling Seek becoming SeekRelative. We could push an update to the codebase renaming Seek to SeekRelative before you start, or I'm also happy if you just rename them all as you go.

[...]

PR #32, opened 2018-05-21
Refactor ClmFile::CreateVolume and subordinate functions

Brett208:
Remove all Windows specific code from packing a clmFile.

 - Rename ArchivePacker::CreateVolume as CreateArchive
 - Exchange C Style Arrays for std::vector, allowing for simplified memory management
 - Change outFile from a HANDLE to a FileStreamWriter
 - Rename outFile to clmFileWriter
 - Rename ClmFile::WriteVolume to WriteArchive
 - Move creation of clmFileWriter to ClmFile::WriteArchive
 - Replace filesToPack HANDLES with FileStreamWriters.
 - Standardize ClmFile.cpp constants as fixed width unsigned integers
 - Begin standardizing use of constant width integers when serializing and unserializing data.

### DIFF
--- a/Archives/ClmFile.cpp
+++ b/Archives/ClmFile.cpp
@@ -266,7 +266,7 @@ namespace Archives
 
 		std::vector<std::string> internalFileNames = GetInternalNamesFromPaths(filesToPack);
 		internalFileNames = StripFileNameExtensions(internalFileNames);
-		// Do not allow duplicate names when packing. Will cause undefined behaviour in binary search and file extraction.
+		// Allowing duplicate names when packing may cause unintended results during binary search and file extraction.
 		CheckSortedContainerForDuplicateNames(internalFileNames); 
 
 		// Write the archive header and copy files into the archive
@@ -332,7 +332,7 @@ namespace Archives
 	// Searches through the wave file to find the given chunk length
 	// The current stream position is set the the first byte after the chunk header
 	// Returns the chunk length if found or -1 otherwise
-	int ClmFile::FindChunk(uint32_t chunkTag, FileStreamReader& fileStreamReader)
+	int ClmFile::FindChunk(uint32_t chunkTag, SeekableStreamReader& seekableStreamReader)
 	{
 #pragma pack(push, 1)
 		struct ChunkHeader
@@ -342,7 +342,7 @@ namespace Archives
 		};
 #pragma pack(pop)
 
-		uint64_t fileSize = fileStreamReader.Length();
+		uint64_t fileSize = seekableStreamReader.Length();
 
 		if (fileSize < 20) {
 			return -1;
@@ -350,14 +350,14 @@ namespace Archives
 
 		// Seek to beginning of first internal chunk (provided it exists)
 		// Note: this seeks past the initial format tag (such as RIFF and WAVE)
-		fileStreamReader.Seek(12);
+		uint32_t currentPosition = sizeof(RiffHeader);
+		seekableStreamReader.Seek(currentPosition);
 
 		ChunkHeader header;
-		uint32_t currentPosition = 12;
 		do
 		{
 			// Read the tag
-			fileStreamReader.Read((char*)&header, sizeof(header));
+			seekableStreamReader.Read((char*)&header, sizeof(header));
 			
 			// Check if this is the right header
 			if (header.formatTag == chunkTag) {
@@ -366,7 +366,7 @@ namespace Archives
 			
 			// If not the right header, skip to next header
 			currentPosition += header.length + 8;
-			fileStreamReader.Seek(currentPosition);
+			seekableStreamReader.Seek(currentPosition);
 		} while (currentPosition < fileSize);
 
 		return -1;	// Failed to find the tag
@@ -376,9 +376,9 @@ namespace Archives
 	// Returns true if they are all the same and false otherwise.
 	bool ClmFile::CompareWaveFormats(const std::vector<WaveFormatEx>& waveFormats)
 	{
-		for (WaveFormatEx waveFormat : waveFormats)
+		for (size_t i = 1; i < waveFormats.size(); i++)
 		{
-			if (memcmp(&waveFormat, &waveFormats[0], sizeof(WaveFormatEx))) {
+			if (memcmp(&waveFormats[i], &waveFormats[0], sizeof(WaveFormatEx))) {
 				return false;
 			}
 		}
@@ -417,7 +417,7 @@ namespace Archives
 		clmFileWriter.Write((char*)indexEntries.data(), header.packedFilesCount * sizeof(IndexEntry));
 
 		// Copy files into the archive
-		for (int i = 0; i < header.packedFilesCount; i++) {
+		for (size_t i = 0; i < header.packedFilesCount; i++) {
 			PackFile(clmFileWriter, indexEntries[i], *filesToPackReaders[i]);
 		}
 	}
@@ -437,7 +437,7 @@ namespace Archives
 	}
 
 	// Write file into the Clm Archive by using the fixed memory size of CLM_WRITE_SIZE.
-	void ClmFile::PackFile(FileStreamWriter& clmFileWriter, const IndexEntry& indexEntry, FileStreamReader& fileToPackReader)
+	void ClmFile::PackFile(StreamWriter& clmWriter, const IndexEntry& indexEntry, StreamReader& fileToPackReader)
 	{
 		uint32_t numBytesToRead;
 		uint32_t offset = 0; // Max size of CLM IndexEntry::dataLength is 32 bits.
@@ -456,7 +456,7 @@ namespace Archives
 			fileToPackReader.Read(buffer.data(), numBytesToRead);
 			offset += numBytesToRead;
 
-			clmFileWriter.Write(buffer.data(), numBytesToRead);
+			clmWriter.Write(buffer.data(), numBytesToRead);
 		} while (numBytesToRead); // End loop when numBytesRead/Written is equal to 0
 	}
 
--- a/Archives/ClmFile.cpp
+++ b/Archives/ClmFile.cpp
@@ -187,8 +187,8 @@ namespace Archives
 		headerOut.formatChunk.waveFormat = m_WaveFormat;
 		headerOut.formatChunk.waveFormat.cbSize = 0;
 
-		headerOut.dataChunk.dataTag = DATA;
-		headerOut.dataChunk.dataSize = m_IndexEntry[fileIndex].dataLength;
+		headerOut.dataChunk.formatTag = DATA;
+		headerOut.dataChunk.length = m_IndexEntry[fileIndex].dataLength;
 	}
 
 	std::unique_ptr<SeekableStreamReader> ClmFile::OpenSeekableStreamReader(const char* internalFileName)
@@ -334,17 +334,9 @@ namespace Archives
 	// Returns the chunk length if found or -1 otherwise
 	int ClmFile::FindChunk(uint32_t chunkTag, SeekableStreamReader& seekableStreamReader)
 	{
-#pragma pack(push, 1)
-		struct ChunkHeader
-		{
-			uint32_t formatTag;
-			uint32_t length;
-		};
-#pragma pack(pop)
-
 		uint64_t fileSize = seekableStreamReader.Length();
 
-		if (fileSize < 20) {
+		if (fileSize < sizeof(RiffHeader) + sizeof(ChunkHeader)) {
 			return -1;
 		}
 
@@ -365,7 +357,7 @@ namespace Archives
 			}
 			
 			// If not the right header, skip to next header
-			currentPosition += header.length + 8;
+			currentPosition += header.length + sizeof(ChunkHeader);
 			seekableStreamReader.Seek(currentPosition);
 		} while (currentPosition < fileSize);
 
--- a/Archives/ClmFile.cpp
+++ b/Archives/ClmFile.cpp
@@ -1,17 +1,17 @@
 #include "ClmFile.h"
-#include "../StreamReader.h"
-#include "../StreamWriter.h"
 #include "../XFile.h"
 #include <stdexcept>
 #include <algorithm>
+#include <array>
 
 namespace Archives
 {
-	const unsigned int CLM_WRITE_SIZE = 0x00020000;
-	const int32_t RIFF = 0x46464952;	// "RIFF"
-	const int32_t WAVE = 0x45564157;	// "WAVE"
-	const int FMT  = 0x20746D66;	// "fmt "
-	const int DATA = 0x61746164;	// "data"
+	const uint32_t CLM_WRITE_SIZE = 0x00020000;
+	const uint32_t RIFF = 0x46464952;	// "RIFF" Resource Interchange File Format
+	const uint32_t WAVE = 0x45564157;	// "WAVE"
+	const uint32_t FMT  = 0x20746D66;	// "fmt "
+	const uint32_t DATA = 0x61746164;	// "data"
+
 	ClmFile::ClmFile(const char *fileName) : ArchiveFile(fileName)
 	{
 		m_FileName = nullptr;
@@ -170,7 +170,7 @@ namespace Archives
 
 			} while (numBytesRead);
 		}
-		catch (std::exception e)
+		catch (std::exception& e)
 		{
 			throw std::runtime_error("Error attempting to extracted uncompressed file " + pathOut + ". Internal Error Message: " + e.what());
 		}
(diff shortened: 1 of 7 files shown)


Brett208:
Sorry for the lengthy number of changes. With this update, all Windows specific code has been removed from the CLM Archive packing code. 

Eventually, I would like to remove the bool return on ArchivePacker::CreateArchive & ArchivePacker::Repack. Then we can throw actual exceptions that provide better details on what part of the process failed. However, this would require updating VolFile as well and I didn't want to continuing expanding the number of changes. So right now, ClmFile::CreateArchive will catch exceptions and convert them to FALSE function returns.

Since we are using the RIFF file format, I'm thinking that the file size is limited to a chunk size of unsigned 32 bits? I was briefly reading that there are ways around this, but I'm guessing that Outpost 2 would not support it? I tried to mirror this by using uint32_t to represent file stream offsets within individual files for packing. Don't mind changing this up if there is a more appropriate way to handle.

I had to instantiate the files to pack FileStreamReaders inside a std::unique_ptr in order to store them in the vector. Executing vector::push_back on a vector<FileStreamReader> would cause Compiler Error C2280 (https://docs.microsoft.com/en-us/cpp/error-messages/compiler-errors-1/compiler-error-c2280) due to the lack of a copy constructor for FileStreamReader. Changing to std::vector<std::unique_ptr<FileStreamReader>> fixes this. Unfortunately, it is a really long type to pass around to subordinate functions. 

Apparently a copy constructor does not exist for std::ifstream. 

I wasn't sure if it would be smart to add a copy constructor for a class containing a file handle, so I went with the std::unique_ptr instead. I must admit that I am weak on C++ constructors, especially with Move and Copy Constructors.

```
//Copy Constructor not created
//FileStreamReader(const FileStreamReader& fileStreamReader);
```

I used the name filesToPackReaders for the container of all FileStreamReaders for packing. I think this name is descriptive, but it is sort of clunky. I would be happy to exchange it if there was a better idea.

I required the following two new functions for FileStreamReader, FileStreamReader::Length & FileStreamReader::Seek. I tried adding them as listed in the issue SeekableStream Enhancements #24. However, I did not fully implement them through the full class hierarchy.

This is a major overhaul of the code, and I hope that I didn't miss any important details in the process. I am happy with the results but am certainly willing to sit on this a while as needed for continued refinement.

DanRStevens:
Catching the exception and returning false seems like an appropriate intermediate step. That was a good call.

For file packing, a limit is imposed by the CLM format. I believe the CLM file is limited to 4GB/32-bit in some respects, due to internal offset and size parameters being stored as 32-bit. I suppose technically that could result in 8GB of addressable space once you sum them together, so it might still make sense to support a larger offset size. I suspect WAV files could have multiple data chunks, each up to 4GB in length, though support for processing that might be limited. At any rate, the limits are such that this is an issue we could easily ignore for now.

The filesToPackReaders does sound a bit clunky, but if the plan is to remove this array and instead open each file as it's being processed, this parameter will be dropped/replaced anyway. The unique_ptr issue is similarly related.

Archives/ClmFile.cpp
```diff
@@ -222,242 +222,175 @@ namespace Archives
...
+		if (fileSize < 20) {
+			return -1;
+		}
 
-		fileSize = GetFileSize(file, nullptr);
-		if (fileSize < 20) return -1;
 		// Seek to beginning of first internal chunk (provided it exists)
 		// Note: this seeks past the initial format tag (such as RIFF and WAVE)
-		if (SetFilePointer(file, 12, nullptr, FILE_BEGIN) == -1) return -1;
+		fileStreamReader.Seek(12);
```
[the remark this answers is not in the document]
Brett208: I was thinking the same thing, but wasn't 100% sure what the 12 bytes were. I think it is the RiffHeader. If so, I'll update to read sizeof(RiffHeader).

Archives/ClmFile.cpp
```diff
@@ -222,242 +222,175 @@ namespace Archives
...
-	}
-
-	// Compares wave format structures in the array waveFormat
+	// Compares wave format structures in the waveFormats container
 	// Returns true if they are all the same and false otherwise.
-	bool ClmFile::CompareWaveFormats(int numFilesToPack, WaveFormatEx *waveFormat)
+	bool ClmFile::CompareWaveFormats(const std::vector<WaveFormatEx>& waveFormats)
 	{
-		for (int i = 1; i < numFilesToPack; i++)
+		for (WaveFormatEx waveFormat : waveFormats)
```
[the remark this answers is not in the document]
Brett208: The function is returning true if all wave formats are duplicates which should not result in a bug in its current form. However, it doesn't make sense to compare the first index of WaveFormats with itself before checking the rest. Definitely a mistake on my part. I'll fix it shortly.

Archives/ClmFile.h
```diff
@@ -77,22 +78,22 @@ namespace Archives
...
-			IndexEntry *entry, const std::vector<std::string>& internalNames, WaveFormatEx *waveFormat);
-		void ClmFile::PrepareIndex(int headerSize, const std::vector<std::string>& internalNames, IndexEntry* entry);
-		bool PackFile(HANDLE outFile, const IndexEntry& entry, HANDLE fileHandle);
+		bool ReadAllWaveHeaders(std::vector<std::unique_ptr<FileStreamReader>>& filesToPackReaders, std::vector<WaveFormatEx>& waveFormats, std::vector<IndexEntry>& indexEntries);
+		int FindChunk(uint32_t chunkTag, FileStreamReader& fileStreamReader);
+		bool CompareWaveFormats(const std::vector<WaveFormatEx>& waveFormats);
+		void WriteArchive(std::string& archiveFileName, std::vector<std::unique_ptr<FileStreamReader>>& filesToPackReaders,
+			std::vector<IndexEntry>& indexEntries, const std::vector<std::string>& internalNames, const WaveFormatEx& waveFormat);
+		void PrepareIndex(int headerSize, const std::vector<std::string>& internalNames, std::vector<IndexEntry>& indexEntries);
+		void PackFile(FileStreamWriter& clmFileWriter, const IndexEntry& entry, FileStreamReader& fileToPackReader);
```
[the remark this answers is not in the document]
Brett208: I don't mind updating to SeekableStreamReader&. I'm curious why you would prefer SeekableStreamReader in this situation though as we know it is implemented as a FileStreamReader. Is it just to make maybe switching to a different SeekableStreamReader easier sometime in the future?

StreamReader.cpp
```diff
@@ -11,7 +11,7 @@ FileStreamReader::FileStreamReader(std::string filename) {
 	file = std::ifstream(filename, std::ios::in | std::ios::binary);
 
 	if (!file.is_open()) {
-		throw std::runtime_error("File could not be opened.");
+		throw std::runtime_error("The following file could not be opened: " + filename + ".");
```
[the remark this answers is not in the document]
Brett208: Sounds good, I'll update shortly.

Archives/ClmFile.cpp
```diff
@@ -170,7 +170,7 @@ namespace Archives
 
 			} while (numBytesRead);
 		}
-		catch (std::exception e)
+		catch (std::exception& e)
```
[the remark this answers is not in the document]
Brett208: Took me way too long to figure the pun out...

Archives/ClmFile.cpp
```diff
@@ -222,242 +222,175 @@ namespace Archives
...
-			return false;
+		// Check if all wave formats are the same
+		if (!CompareWaveFormats(waveFormats)) {	
+			return false; // Not all wave formats match
 		}
 
 		std::vector<std::string> internalFileNames = GetInternalNamesFromPaths(filesToPack);
 		internalFileNames = StripFileNameExtensions(internalFileNames);
-		CheckSortedContainerForDuplicateNames(internalFileNames);
+		// Do not allow duplicate names when packing. Will cause undefined behaviour in binary search and file extraction.
```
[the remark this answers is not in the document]
Brett208: What do you think of this comment instead:

```
// Allowing duplicate names when packing may cause unintended results during binary search and file extraction.
```
I think that comment would apply whether we go with abort or warning in the future. I do want some sort of comment here so it is clear where the user will find trouble if duplicate names are allowed. I didn't even realize Outpost 2 was using a binary search to find the files when I first encountered the bug.

Archives/ClmFile.cpp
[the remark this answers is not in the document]
DanRStevens: Correct: [Wave File Format](http://soundfile.sapp.org/doc/WaveFormat/) (Or at least a canonical example, if not the full gory details of all the possible sections).

Archives/ClmFile.cpp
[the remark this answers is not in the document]
DanRStevens: That sounds like a good description. You're right that we should say something.

Archives/ClmFile.h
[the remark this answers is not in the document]
DanRStevens: Yes. In general, try to declare the most generic type that will work efficiently. It makes the code more flexible for reuse later. Perhaps in the future we might allow packing of data from a network socket, or another archive file. I don't know why we would, but we could without any code changes using the more general data type. It's also a way of documenting what the actual real required features are.

Archives/ClmFile.cpp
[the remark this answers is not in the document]
DanRStevens: Ahh, right! The check is opposite what I was thinking. I was thinking in terms of a duplicate filename check, where of course you'd expect them to all be different. In that case, maybe don't worry about this then, since the code will be subject to change if we start processing each Wave file one at a time, using delayed opening.

Brett208:
Just committed changes to the StreamReader family. This will allow using StreamReader and SeekableStreamReader interfaces within the ClmFile class in the next commit. 

The function MemoryStreamReader::Read is a bit complicated. std::memcpy uses size_t for the size of memory copied. I think this function would successfully allow reading larger files than supported by size_t if necessary. Open to suggestions here if there are better alternatives.

This would be a nice time to have some coverage using unit tests.

Thanks,
Brett

StreamReader.cpp
```diff
@@ -37,34 +37,58 @@ void FileStreamReader::Seek(uint64_t position) {
...
+	// Depending on the system, size_t may not be able to store the max value of uint64_t. 
+	// SIZE_MAX is the largest value that may be stored in size_t for the given system.
+	do {
+		size_t nextReadLength = SIZE_MAX;
+		// Check if remaining value of size will fit in a size_t.
+		if (size < SIZE_MAX) {
+			nextReadLength = static_cast<size_t>(size);
+		}
+
+		std::memcpy(buffer, streamBuffer + position, nextReadLength);
```
DanRStevens: This will overwrite the same section of buffer with each chunk.

StreamReader.cpp
```diff
@@ -37,34 +37,58 @@ void FileStreamReader::Seek(uint64_t position) {
...
-	if (offset + size > streamSize) {
+	if (position + size > streamSize) {
 		throw std::runtime_error("Size of bytes to read exceeds remaining size of buffer.");
 	}
 
-	memcpy(buffer, streamBuffer + offset, size);
-	offset += size;
+	// std::memcpy takes a _Size to copy of size_t. 
+	// Depending on the system, size_t may not be able to store the max value of uint64_t. 
+	// SIZE_MAX is the largest value that may be stored in size_t for the given system.
```
DanRStevens: Very interesting comment.

The stream position may be 64-bit, though the size that can be read into memory at any one time will be limited by the address space, which is represented by `size_t`. It makes sense for size to be declared as `size_t` as this is memory space, while the file offset is 64-bit to ensure files greater than 4GB (or greater than the memory size) can be supported.

StreamReader.cpp
```diff
@@ -37,34 +37,58 @@ void FileStreamReader::Seek(uint64_t position) {
...
+	} while (size != 0); // Continue reading in chunks until all data has been copied.
 }
 
-void MemoryStreamReader::SeekRelative(int offset)
+uint64_t MemoryStreamReader::Length() {
+	return streamSize;
+}
+
+void MemoryStreamReader::Seek(uint64_t position) {
+	if (position > this->position) {
```
DanRStevens: Looks like this should be `position > streamSize`.

StreamReader.h
```diff
@@ -2,28 +2,35 @@
 
 #include <fstream>
 #include <string>
+#include <cstdint>
 
 class StreamReader {
 public:
 	virtual ~StreamReader() = 0;
-	virtual void Read(char* buffer, size_t size) = 0;
+	virtual void Read(char* buffer, uint64_t size) = 0;
```
DanRStevens: I think this should remain `size_t`, as memory reads are always limited by addressable memory size.

StreamReader.h
```diff
@@ -2,28 +2,35 @@
...
+	virtual void SeekRelative(int64_t offset) = 0;
 };
 
 class FileStreamReader : public SeekableStreamReader {
 public:
 	FileStreamReader(std::string fileName);
 	~FileStreamReader();
 
-	void Read(char* buffer, size_t size);
+	void Read(char* buffer, uint64_t size);
```
DanRStevens: Ditto.

StreamReader.h
```diff
@@ -32,12 +39,14 @@ class FileStreamReader : public SeekableStreamReader {
 
 class MemoryStreamReader : public SeekableStreamReader {
 public:
-	MemoryStreamReader(char* buffer, size_t size);
-	void Read(char* buffer, size_t size);
-	void SeekRelative(int offset);
+	MemoryStreamReader(char* buffer, uint64_t size);
```
DanRStevens: I'm a bit on the fence here, but size could also be of type `size_t` here too, at least for the constructor parameter. The implementation of this class is necessarily restricted to memory size, and the constructor will always be implementation specific. (As in, it can never be virtual).

Virtual method signatures will of course have to match the parent class, so they will need to take `uint64_t`. The return value of virtual methods must be covariant (roughly meaning the same, or a subclass of), which might allow `size_t` as the return type, though I'm not sure that's guaranteed to be covariant with `uint64_t` on all platforms.

Internal field sizes could potentially be either `size_t` or `uint64_t`. I suppose it depends on casting requirements and convenience.

We could revisit this later as a low priority item.

StreamReader.h
```diff
@@ -32,12 +39,14 @@ class FileStreamReader : public SeekableStreamReader {
...
+	uint64_t Length();
+	void Seek(uint64_t position);
+	void SeekRelative(int64_t offset);
 
 private:
 	char* streamBuffer;
-	size_t streamSize;
-	size_t offset;
+	uint64_t streamSize;
+	uint64_t position;
```
DanRStevens: See previous note about internal field sizes.

StreamReader.cpp
```diff
@@ -37,34 +37,58 @@ void FileStreamReader::Seek(uint64_t position) {
...
+		size_t nextReadLength = SIZE_MAX;
+		// Check if remaining value of size will fit in a size_t.
+		if (size < SIZE_MAX) {
+			nextReadLength = static_cast<size_t>(size);
+		}
+
+		std::memcpy(buffer, streamBuffer + position, nextReadLength);
+		position += nextReadLength;
+		size -= nextReadLength;
+	} while (size != 0); // Continue reading in chunks until all data has been copied.
```
DanRStevens: I think we may be better off with the `memcpy`. If the range different between `size_t` and `uint64_t` is an issue, perhaps we can raise an exception in that case.

StreamReader.cpp
Brett208: Thanks for catching this. Will fix shortly.

StreamReader.cpp
Brett208: I didn't follow how the overwrite was occurring, but I'm switching back to the simpler model so it is a mute point.

Brett208:
Just made 2 commits. The first commit made suggested changes to the StreamReader class family. The second made suggested changes to ClmFile. I'll try leaving this branch open for a bit longer in case there are any more suggestions.

I think my next move will either be adding exceptions to ArchivePacker::CreateArchive (will require work in both VolFile & ClmFile implementations) or continue to remove Windows specific code from ClmFile.

Thanks,
Brett

StreamReader.cpp
```diff
@@ -90,5 +79,6 @@ void MemoryStreamReader::SeekRelative(int64_t offset)
 		throw std::runtime_error("Change in offset places read position outside bounds of buffer.");
 	}
 
-	this->position += offset;
+	// offset is checked against size of streamSize, which cannot exceed SIZE_MAX (max size of size_t)
+	this->position += static_cast<size_t>(offset);
```
DanRStevens: I believe there is some potential for bounds issues here. It's not `offset`, but rather the sum `position + offset` which is checked against `streamSize`. There's potential for wraparound to occur before the check happens, and it's a bit complicated to think about when different types are mixed, especially since `size_t` and `int64_t` are not necessarily ordered as one being bigger than the other, and they differ in signedness. At least we're not mixing 3 different data types together now though.

Since both `position` and `streamSize` are unsigned values (now of the same type), they might hold a large positive value just short of the maximum range, with `position` being less than `streamSize`. Using a small positive `offset`, greater than the number of bytes left available, the sum `position + offset` might overflow and wrap around to a small positive value (the result will be unsigned). That small positive value will almost certainly be less than the large positive value of `streamSize`. Thus a small relative seek of only a few bytes could reset a stream position from near the end back to near the beginning, assuming the stream size was quite large.

In practice, we won't be using stream sizes that are big enough for that to be an issue, and OS limitations may keep us from hitting that issue (unlikely you can allocate a buffer spanning half the address space or more), though we should at least comment this. In assembly it would be very easy to check overflow/carry flags, though in portable C++ it gets a little more complicated.

Archives/ClmFile.cpp
```diff
@@ -342,22 +342,22 @@ namespace Archives
 		};
 #pragma pack(pop)
 
-		uint64_t fileSize = fileStreamReader.Length();
+		uint64_t fileSize = seekableStreamReader.Length();
 
 		if (fileSize < 20) {
```
DanRStevens: sizeof(RiffHeader) + sizeof(ChunkHeader)

StreamReader.cpp
DanRStevens: Yes, moot point now. As for what I meant, the destination `buffer` was not updated in the loop, only the source `streamBuffer + position`. Hence it would overwrite the same area of the destination buffer each time through the loop. The more fundamental issue though, is a buffer can never be bigger than `size_t` allows, so there's no way to iterate over `SIZE_MAX` chunks of memory and actually have it point to different memory.

Archives/ClmFile.cpp
Brett208: Will incorporate soon.

StreamReader.cpp
```diff
@@ -42,44 +42,33 @@ void FileStreamReader::SeekRelative(int64_t offset) {
...
+	position += size;
 }
 
 uint64_t MemoryStreamReader::Length() {
 	return streamSize;
 }
 
 void MemoryStreamReader::Seek(uint64_t position) {
-	if (position > this->position) {
+	if (position >= streamSize) {
```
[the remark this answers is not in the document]
Brett208: I haven't thought about this situation. I'll update shortly to allow though.

StreamReader.cpp
Brett208: I took a stab at catching the bounds issues. Will add to next commit.

Archives/ClmFile.cpp
```diff
@@ -366,7 +366,7 @@ namespace Archives
 			
 			// If not the right header, skip to next header
 			currentPosition += header.length + 8;
-			fileStreamReader.Seek(currentPosition);
+			seekableStreamReader.Seek(currentPosition);
 		} while (currentPosition < fileSize);
```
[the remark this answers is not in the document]
Brett208: I wouldn't mind seeing this loop a little less complex, but it is working and I'm happy with the current form. 

On a similar note. Are the structs DataChunk in ClmFile.h and ChunkHeader in ClmFile.cpp representing the same thing? If so, I'd like to combine them.

Thanks,
Brett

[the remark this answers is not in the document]
DanRStevens: Hmm, good catch, they are the same. The ChunkHeader name is probably the most accurate and generic, since there is both a format chunk, and a data chunk, at minimum, in a wave file.

As the structs are both declared within functions, one of them will need to move somewhere more globally accessible.

It may make some sense to separate out some of the WAVE file code into a separate class for reading/writing .wav files. I had tried doing that at one point, but found the Stream code wasn't quite advanced enough yet to support it well. In particular, being about to return a slice of a stream would have been useful for providing access to the raw wave data, without the file headers. Ideally, such a stream slice should be bounds checked, so as to prevent seeking or reading outside of the slice bounds.

For now perhaps we can just move the struct up to the top of the .cpp file, or maybe the private section of the header file. Or at the very least, use the same name for the duplicate.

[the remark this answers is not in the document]
Brett208: I'll use the definition of ChunkHeader from ClmFile::FindChunk by pulling it into the Header file so it can be used in both places.

I would agree on basic read/write of WAVE in a separate class. Not sure if we will make it that far. For now it would be nice to maybe just organize the structs involved in reading a generic wave file separate from the ClmFile struct to make things more clear.

Brett208:
@DanRStevens 

What do you think of this function for bounds checking on MemoryStreamReader?

```
void MemoryStreamReader::SeekRelative(int64_t offset)
{
    // Cast position for testing into a uint64_t. A negative value will wrap around to a positive value.
    uint64_t positionCast = static_cast<uint64_t>(this->position) + offset;

    if (positionCast > streamSize ||
        (offset > 0 && positionCast < this->position) || // Check if offset wraps past max size.
        (offset < 0 && positionCast > this->position)) // Check if offset wraps past min size. 
        {
            Throw std::runtime_error("Change in offset puts read position outside bounds of buffer.");
        }

        // offset is checked against size of streamSize, which cannot exceed SIZE_MAX (max size of size_t)
        this->position += static_cast<size_t>(offset); 
}
```

Thanks,
Brett

Archives/ClmFile.cpp
```diff
@@ -365,7 +357,7 @@ namespace Archives
 			}
 			
 			// If not the right header, skip to next header
-			currentPosition += header.length + 8;
+			currentPosition += header.length + sizeof(ChunkHeader);
```
DanRStevens: Nice catch. I didn't really notice that one.

Brett208:
I forgot to commit the StreamReader changes before merging this branch, so they are just committed against the master branch.

Understand there may be room for better checks against SeekRelative's offset. Since where we are at seems in the right direction, I've committed it and we can move on to other things for now. Thank you for reviewing.

I've locally removed the last Window's specific HANDLE from ClmFile. Everything works except the internal name of packed files is displaying some random characters on the end of the packed filenames. I suspect it is related to problems with the null terminator. If I cannot solve in the short term, I'll post for some help. 

-Brett

DanRStevens:
That does sound like a null termination issue. Not sure if you're saying there's garbage output shown by the program, or if this is just garbage on disk as seen with a hex editor. It's possible the files as written on disk will have garbage from uninitialized memory at the end of filenames, particularly if the filenames are short. A null terminator should prevent that from displaying. Ideally, the filename buffers should be completed filled with zero padding in the case of short filenames. Writing unitialized memory to disk could be seen as a possible security leak. Actually, I'd once thought of using uninitialized buffer space to leak data in network programs through carefully constructed code. It would have been very difficult to catch the data leak from a source code analysis, and even harder to prove intent.

PR #33, opened 2018-05-28
Remove Windows header from ClmFile

Brett208:
- When opening or extracting from a CLM File, use FileStreamWriter instead of HANDLE
 - Update ClmFile to use C++11/14 practices when opening and extracting CLM files.
 - Use sizeof(X) instead of hardcoded integers when opening or extracting from CLM files.
 - Separate WAVE specific structures and constants into a separate header (WaveFile.h)
 - Clean ArchivePacker.cpp

### DIFF
--- a/Archives/ArchivePacker.cpp
+++ b/Archives/ArchivePacker.cpp
@@ -75,6 +75,10 @@ namespace Archives
 
 	void ArchivePacker::CheckSortedContainerForDuplicateNames(const std::vector<std::string>& internalNames) 
 	{
+		if (internalNames.size() == 0) {
+			return;
+		}
+
 		for (size_t i = 0; i < internalNames.size() - 1; ++i)
 		{
 			if (StringHelper::CheckIfStringsAreEqual(internalNames[i], internalNames[i + 1])) {
--- a/Archives/ArchivePacker.cpp
+++ b/Archives/ArchivePacker.cpp
@@ -66,7 +66,7 @@ namespace Archives
 	{
 		std::vector<std::string> fileNames;
 
-		for (std::string& fileName : paths) {
+		for (const std::string& fileName : paths) {
 			fileNames.push_back(XFile::GetFilename(fileName));
 		}
 
@@ -75,13 +75,9 @@ namespace Archives
 
 	void ArchivePacker::CheckSortedContainerForDuplicateNames(const std::vector<std::string>& internalNames) 
 	{
-		if (internalNames.size() == 0) {
-			return;
-		}
-
-		for (size_t i = 0; i < internalNames.size() - 1; ++i)
+		for (size_t i = 1; i < internalNames.size(); ++i)
 		{
-			if (StringHelper::CheckIfStringsAreEqual(internalNames[i], internalNames[i + 1])) {
+			if (StringHelper::CheckIfStringsAreEqual(internalNames[i - 1], internalNames[i])) {
 				throw std::runtime_error("Unable to create an archive containing files with the same filename. Duplicate filename: " + internalNames[i]);
 			}
 		}
--- a/Archives/ArchivePacker.cpp
+++ b/Archives/ArchivePacker.cpp
@@ -50,13 +50,14 @@ namespace Archives
 	// Returns true if successful and false otherwise
 	bool ArchivePacker::ReplaceFileWithFile(const char *fileToReplace, const char *newFile)
 	{
-		MoveFileA(fileToReplace, "Delete.vol");
+		std::string tempFileName("Temporary.vol");
+		MoveFileA(fileToReplace, tempFileName.c_str());
 		
 		if (MoveFileA(newFile, fileToReplace) == 0) {
 			return false;
 		}
 
-		DeleteFileA("Delete.vol");
+		DeleteFileA(tempFileName.c_str());
 
 		return true;
 	}
--- a/Archives/ArchivePacker.cpp
+++ b/Archives/ArchivePacker.cpp
@@ -12,7 +12,9 @@ namespace Archives
 
 	ArchivePacker::~ArchivePacker()
 	{
-		if (m_OutFileHandle) CloseHandle(m_OutFileHandle);
+		if (m_OutFileHandle) {
+			CloseHandle(m_OutFileHandle);
+		}
 	}
 
 	// Opens the file for output and returns nonzero if successful
@@ -37,7 +39,10 @@ namespace Archives
 	// Closes the file whose handle is in m_OutFileHandle
 	void ArchivePacker::CloseOutputFile()
 	{
-		if (m_OutFileHandle) CloseHandle(m_OutFileHandle);
+		if (m_OutFileHandle) {
+			CloseHandle(m_OutFileHandle);
+		}
+
 		m_OutFileHandle = nullptr;
 	}
 
@@ -46,7 +51,11 @@ namespace Archives
 	bool ArchivePacker::ReplaceFileWithFile(const char *fileToReplace, const char *newFile)
 	{
 		MoveFileA(fileToReplace, "Delete.vol");
-		if (MoveFileA(newFile, fileToReplace) == 0) return false;
+		
+		if (MoveFileA(newFile, fileToReplace) == 0) {
+			return false;
+		}
+
 		DeleteFileA("Delete.vol");
 
 		return true;
@@ -56,7 +65,7 @@ namespace Archives
 	{
 		std::vector<std::string> fileNames;
 
-		for (std::string fileName : paths) {
+		for (std::string& fileName : paths) {
 			fileNames.push_back(XFile::GetFilename(fileName));
 		}
 
--- a/Archives/ClmFile.cpp
+++ b/Archives/ClmFile.cpp
@@ -39,11 +39,11 @@ namespace Archives
 		}
 
 		if (!clmHeader.CheckUnknown()) {
-			throw std::runtime_error("Unkown CLM header field is incorrect.");
+			throw std::runtime_error("Unknown field in CLM header is incorrect.");
 		}
 
 		if (clmHeader.packedFilesCount < 1) {
-			throw std::runtime_error("Packed files count must be 1 or greater.");
+			throw std::runtime_error("Packed file count must be 1 or greater.");
 		}
 
 		m_NumberOfPackedFiles = clmHeader.packedFilesCount;
--- a/Archives/ClmFile.cpp
+++ b/Archives/ClmFile.cpp
@@ -321,7 +321,7 @@ namespace Archives
 
 	// Compares wave format structures in the waveFormats container
 	// Returns true if they are all the same and false otherwise.
-	bool ClmFile::CompareWaveFormats(const std::vector<WaveFormatEx>& waveFormats)
+	bool ClmFile::CompareWaveFormats(const std::vector<WaveFormatEx>& waveFormats) const
 	{
 		for (size_t i = 1; i < waveFormats.size(); i++)
 		{
@@ -406,9 +406,8 @@ namespace Archives
 	}
 
 
-
-	const std::array<char, 32> standardFileVersion { 'O', 'P', '2', ' ', 'C', 'l', 'u', 'm', 'p', ' ',
-		'F', 'i', 'l', 'e', ' ', 'V', 'e', 'r', 's', 'i', 'o', 'n', ' ', '1', '.', '0', '\x01A', '\0', '\0', '\0', '\0' };
+	// A null terminator (\0) is automatically assigned to the end of the string when placing it within the std::array
+	const std::array<char, 32> standardFileVersion { "OP2 Clump File Version 1.0\x01A\0\0\0\0" };
 	const std::array<char, 6> standardUnknown { 0, 0, 0, 0, 1, 0 };
 
 	ClmFile::ClmHeader::ClmHeader(WaveFormatEx waveFormat, uint32_t packedFilesCount) {
--- a/Archives/ClmFile.cpp
+++ b/Archives/ClmFile.cpp
@@ -9,19 +9,8 @@ namespace Archives
 
 	ClmFile::ClmFile(const char *fileName) : ArchiveFile(fileName)
 	{
-		try {
-			clmFileReader = std::make_unique<FileStreamReader>(fileName);
-		}
-		catch (std::exception& e) {
-			throw std::runtime_error("Could not open clm file " + std::string(fileName) + ".");
-		}
-
-		try {
-			ReadHeader();
-		} 
-		catch (std::exception& e) {
-			throw std::runtime_error("Invalid clm header read from file " + std::string(fileName) + ". Internal error message: " + e.what() );
-		}
+		clmFileReader = std::make_unique<FileStreamReader>(fileName);
+		ReadHeader();
 	}
 
 	ClmFile::~ClmFile() { }
@@ -34,27 +23,18 @@ namespace Archives
 	{
 		clmFileReader->Read((char*)&clmHeader, sizeof(ClmHeader));
 		
-		if (!clmHeader.CheckFileVersion()) {
-			throw std::runtime_error("CLM file version is incorrect.");
+		try {
+			clmHeader.VerifyFileVersion();
+			clmHeader.VerifyUnknown();
 		}
-
-		if (!clmHeader.CheckUnknown()) {
-			throw std::runtime_error("Unknown field in CLM header is incorrect.");
-		}
-
-		if (clmHeader.packedFilesCount < 1) {
-			throw std::runtime_error("Packed file count must be 1 or greater.");
+		catch (std::exception& e) {
+			throw std::runtime_error("Invalid clm header read from file " + m_VolumeFileName + ". " + e.what());
 		}
 
 		m_NumberOfPackedFiles = clmHeader.packedFilesCount;
 
 		indexEntries = std::vector<IndexEntry>(m_NumberOfPackedFiles);
 		clmFileReader->Read((char*)indexEntries.data(), m_NumberOfPackedFiles * sizeof(IndexEntry));
-		
-		// Null terminate all packed fileNames.
-		for (IndexEntry& indexEntry : indexEntries) {
-			indexEntry.fileName[8] = 0;
-		}
 	}
 
 
@@ -100,7 +80,7 @@ namespace Archives
 		{
 			FileStreamWriter waveFileWriter(pathOut);
 
-			waveFileWriter.Write((char*)&header, sizeof(header));
+			waveFileWriter.Write(header);
 
 			// Seek to the beginning of the file data (in the .clm file)
 			clmFileReader->Seek(indexEntries[fileIndex].dataOffset);
@@ -118,7 +98,7 @@ namespace Archives
 					numBytesToRead = indexEntries[fileIndex].dataLength - offset;
 				}
 
-				clmFileReader->Read(buffer.data(), CLM_WRITE_SIZE);
+				clmFileReader->Read(buffer.data(), numBytesToRead);
 
 				offset += numBytesToRead;
 
@@ -138,7 +118,7 @@ namespace Archives
 		headerOut.riffHeader.chunkSize = sizeof(headerOut.riffHeader.waveTag) + sizeof(FormatChunk) + sizeof(ChunkHeader) + indexEntries[fileIndex].dataLength;
 
 		headerOut.formatChunk.fmtTag = FMT;
-		headerOut.formatChunk.formatSize = sizeof(WaveFormatEx);
+		headerOut.formatChunk.formatSize = sizeof(headerOut.formatChunk.waveFormat);
 		headerOut.formatChunk.waveFormat = clmHeader.waveFormat;
 		headerOut.formatChunk.waveFormat.cbSize = 0;
 
@@ -174,7 +154,7 @@ namespace Archives
 		for (int i = 0; i < m_NumberOfPackedFiles; i++)
 		{
 			//Filename is equivalent to internalName since filename is a relative path from current directory.
-			filesToPack[i] = std::string(GetInternalFileName(i)) + ".wav";
+			filesToPack[i] = GetInternalFileName(i) + ".wav";
 		}
 
 		return CreateArchive("temp.clm", filesToPack);
@@ -186,10 +166,6 @@ namespace Archives
 	// Returns nonzero if successful and zero otherwise.
 	bool ClmFile::CreateArchive(std::string archiveFileName, std::vector<std::string> filesToPack)
 	{
-		if (filesToPack.size() < 1) {
-			return false; //CLM files require at least one audio file present to properly write settings.
-		}
-
 		// Sort files alphabetically based on the fileName only (not including the full path).
 		// Packed files must be locatable by a binary search of their fileName.
 		std::sort(filesToPack.begin(), filesToPack.end(), ComparePathFilenames);
@@ -226,6 +202,15 @@ namespace Archives
 
 		// Write the archive header and copy files into the archive
 		try {
+			WaveFormatEx waveFormat;
+
+			if (waveFormats.size() > 0) {
+				waveFormat = waveFormats[0];
+			}
+			else {
+				waveFormat = CreateDefaultWaveFormat();
+			}
+
 			WriteArchive(archiveFileName, filesToPackReaders, indexEntries, internalFileNames, waveFormats[0]);
 		}
 		catch (std::exception& e) {
@@ -250,7 +235,7 @@ namespace Archives
 		for (size_t i = 0; i < filesToPackReaders.size(); i++)
 		{
 			// Read the file header
-			filesToPackReaders[i]->Read((char*)&header, sizeof(header));
+			filesToPackReaders[i]->Read(header);
 			if (header.riffTag != RIFF || header.waveTag != WAVE) {
 				return false;		// Error reading header
 			}
@@ -267,7 +252,7 @@ namespace Archives
 			}
 
 			// Read in the wave format
-			filesToPackReaders[i]->Read((char*)&waveFormats[i], sizeof(WaveFormatEx));
+			filesToPackReaders[i]->Read(waveFormats[i]);
 			waveFormats[i].cbSize = 0;
 
 			// Find the start of the data
@@ -321,7 +306,7 @@ namespace Archives
 
 	// Compares wave format structures in the waveFormats container
 	// Returns true if they are all the same and false otherwise.
-	bool ClmFile::CompareWaveFormats(const std::vector<WaveFormatEx>& waveFormats) const
+	bool ClmFile::CompareWaveFormats(const std::vector<WaveFormatEx>& waveFormats)
 	{
 		for (size_t i = 1; i < waveFormats.size(); i++)
 		{
@@ -339,12 +324,12 @@ namespace Archives
 		const std::vector<std::string>& internalNames,
 		const WaveFormatEx& waveFormat)
 	{
-		// ClmFile cannot contain more than 32 bit size internal file cound.
+		// ClmFile cannot contain more than 32 bit size internal file count.
 		ClmHeader header(waveFormat, static_cast<uint32_t>(internalNames.size()));
 
 		FileStreamWriter clmFileWriter(archiveFileName);
 
-		clmFileWriter.Write((char*)&header, sizeof(header));
+		clmFileWriter.Write(header);
 
 		// Prepare and write Archive Index
 		PrepareIndex(sizeof(header), internalNames, indexEntries);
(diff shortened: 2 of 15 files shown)


Brett208:
This code doesn't fully remove Windows.h from the ClmFile class because it is pulled in from ArchivePacker.h. However, no features from Windows.h are used by ClmFile. Since VolFile uses Windows.h feature included in ArchivePacker.h, I couldn't remove Windows.h from ArchivePacker with this branch.

Wave file specific structures and constants are moved into a separate header called WaveFile.h. If we eventually remove ClmFile and instead just use a ClmPacker and ClmUnpacker class pair, Wave information will need to be available to both classes. It is also a first step towards perhaps having a WAVE specific reader/writer.

I would like to better encapsulate the WAVE constants since they now live in a header file and suggestions would be welcome. I thought this was low enough risk though since they are currently wrapped in the Archives namespace.

ClmHeader was expanded as a struct. The same constant is used when either checking the fileVersion is correct or setting the default fileVersion. Unfortunately, the way the file version is set is pretty ugly. I couldn't find a better way to set it using a std::array. There are now 2 constructors on ClmHeader as discussed in other branch. One for use when reading from disk and the other for use when writing a new ClmFile to disk.

```
const std::array<char, 32> standardFileVersion { 'O', 'P', '2', ' ', 'C', 'l', 'u', 'm', 'p', ' ',
		'F', 'i', 'l', 'e', ' ', 'V', 'e', 'r', 's', 'i', 'o', 'n', ' ', '1', '.', '0', '\x01A', '\0', '\0', '\0', '\0' };
```

Let me know what other improvements are noticed or bugs need fixing. Following tests were performed:

 * Writing a new CLM file and playing in game.
 * Extracting a CLM file and playing music in VLC.
 * Listing the packed file names and sizes of an existing CLM file

Thanks,
Brett

DanRStevens:
Hmm, maybe this StackOverflow answer on [How to initialize a std::array<char, n> with a string literal](https://stackoverflow.com/questions/33484233/how-to-initialize-a-stdarraychar-n-with-a-string-literal-omitting-the-trail) will be useful. Though rather than omitting the trailing `\0`, the buffer can be `\0` filled up to the declared size.

Cursory browse through looks fine. The change description makes sense. You seem to have identified a good intermediate point to stop at. I'll try to give the code a more detailed look through tomorrow.

Brett208:
Good call on the std::array<char, X>. I didn't realize it would automatically add a null terminator before attempting to assign the value to the std::array. Taking this into account allowed setting the constant standardFileVersion in a more human readable way. Just committed the change.

```
// A null terminator (\0) is automatically assigned to the end of the string when placing it within the std::array
const std::array<char, 32> standardFileVersion { "OP2 Clump File Version 1.0\x01A\0\0\0\0" };
```

As discussed in the other thread, we need to decide how to handle the null termination of IndexEntry::fileName. Currently, I am setting the 8th index (9th value) to the null terminator. However, this is a c style array with 8 values. While this is compiling, it seems like a bad idea? My preference would probably be to not include the null terminator in the array, but append it when retrieving the fileName in a GetFileName helper function. 

```
char fileName[8];

...

indexEntries = std::vector<IndexEntry>(m_NumberOfPackedFiles);
clmFileReader->Read((char*)indexEntries.data(), m_NumberOfPackedFiles * sizeof(IndexEntry));

...

// Null terminate all packed fileNames.
for (IndexEntry& indexEntry : indexEntries) {
	indexEntry.fileName[8] = 0;
}
```

DanRStevens:
Indeed, that is a bad idea. It compiles because C++ doesn't bounds check arrays for you. You are overwriting adjacent memory when you null terminate like that. In the case of the struct mentioned, the null terminator will overwrite part of the following `dataOffset` field. That can corrupt the file index and lead to glitches in the sound.

As you've suggested, it may not be possible to null terminate within the fixed length fileName field, and so you can instead null terminate in a GetFileName helper function.

One of the downsides of using null terminated strings, is that it often requires copying strings to null terminate them. You can't null terminate in place if you can't modify the original, either due to it being const, or embedded in a longer string, or just stored in a completely filled fixed length buffer.

One note, I would recommend zero filling the end of the fileName array for short filenames. If there is only 4 characters in the filename, you should write 4 `\0` bytes to the end of the fileName array.

Archives/ClmFile.cpp
```diff
@@ -180,15 +135,15 @@ namespace Archives
 	{
 		headerOut.riffHeader.riffTag = RIFF;
 		headerOut.riffHeader.waveTag = WAVE;
-		headerOut.riffHeader.chunkSize = sizeof(headerOut.formatChunk) + 12 + m_IndexEntry[fileIndex].dataLength;
+		headerOut.riffHeader.chunkSize = sizeof(headerOut.riffHeader.waveTag) + sizeof(FormatChunk) + sizeof(ChunkHeader) + indexEntries[fileIndex].dataLength;
```
DanRStevens: This is a bit hard to read.

Initially it seemed like it sums sizes out of the order of appearance according to the file format, though it does actually match. The structs were declared in the order they should likely appear: RiffHeader, FormatChunk, ChunkHeader (for data), followed by the data.

It seems like RiffHeader and FormatChunk have hidden embedded ChunkHeader fields. The first 8 bytes seem to be a ChunkHeader, which contains the contents of the RIFF file. Hence the riffHeader chunk size seems to be everything after the first 8 byte header.

I believe the structs were modelled after the documentation, though perhaps they could be reworked to match the obvious embedded ChunkHeader usage. That might make the size calculation here a little more obvious. Perhaps move the waveTag field to another struct, and wrap it with the FormatChunk and data ChunkHeader. I am somewhat hesitant to deviate from documentation though. Thoughts?

Archives/ClmFile.cpp
```diff
@@ -180,15 +135,15 @@ namespace Archives
...
 
 		headerOut.formatChunk.fmtTag = FMT;
-		headerOut.formatChunk.formatSize = sizeof(headerOut.formatChunk.waveFormat);
-		headerOut.formatChunk.waveFormat = m_WaveFormat;
+		headerOut.formatChunk.formatSize = sizeof(WaveFormatEx);
```
DanRStevens: It may be better to use the size of the field, rather than the size of the struct. If the field declaration is changed, the sizes will auto update to match the declaration. (Perhaps someday there will be a WaveFormatExEx?)

Archives/ClmFile.cpp
```diff
@@ -366,7 +321,7 @@ namespace Archives
 
 	// Compares wave format structures in the waveFormats container
 	// Returns true if they are all the same and false otherwise.
-	bool ClmFile::CompareWaveFormats(const std::vector<WaveFormatEx>& waveFormats)
+	bool ClmFile::CompareWaveFormats(const std::vector<WaveFormatEx>& waveFormats) const
```
DanRStevens: Hmm, maybe consider making this function `static` rather than `const`. It doesn't modify any member variables, and hence const is appropriate, but it also doesn't even read any member variables either, so `static` would be even more appropriate. Also, to eliminate the need for the template files, we should be making all the write functions static so they can be called without an existing class instance. It was wrong of me to make the file writing function virtual.

Archives/ClmFile.cpp
```diff
@@ -384,21 +339,8 @@ namespace Archives
...
-			WaveFormatEx waveFormat;
-			std::array<char, 6> unknown { 0, 0, 0, 0, 1, 0 };
-			uint32_t packedFilesCount;
-		};
-#pragma pack(pop)
-
-		ClmHeader header;
-		header.waveFormat = waveFormat;
-		header.packedFilesCount = internalNames.size();
+		// ClmFile cannot contain more than 32 bit size internal file cound.
```
DanRStevens: Typo: 'count'

Archives/ClmFile.cpp
```diff
@@ -456,10 +398,31 @@ namespace Archives
...
 		return strippedExtensions;
 	}
+
+
+	// A null terminator (\0) is automatically assigned to the end of the string when placing it within the std::array
+	const std::array<char, 32> standardFileVersion { "OP2 Clump File Version 1.0\x01A\0\0\0\0" };
+	const std::array<char, 6> standardUnknown { 0, 0, 0, 0, 1, 0 };
+
+	ClmFile::ClmHeader::ClmHeader(WaveFormatEx waveFormat, uint32_t packedFilesCount) {
+		std::memcpy(fileVersion.data(), standardFileVersion.data(), fileVersion.size());
```
DanRStevens: I believe `std::array` allows copy through assignment:
```
fileVersion = standardFileVersion;
```

Archives/ClmFile.cpp
```diff
@@ -456,10 +398,31 @@ namespace Archives
...
+		std::memcpy(unknown.data(), standardUnknown.data(), unknown.size());
```
DanRStevens: Ditto

Archives/ClmFile.cpp
```diff
@@ -456,10 +398,31 @@ namespace Archives
...
+
+		this->waveFormat = waveFormat;
+		this->packedFilesCount = packedFilesCount;
+	}
+
+	bool ClmFile::ClmHeader::CheckFileVersion() const {
+		return std::memcmp(fileVersion.data(), standardFileVersion.data(), fileVersion.size()) == 0;
```
DanRStevens: Similarly I believe `std::array` has overloaded `operator ==`:
```
return fileVersion == standardFileVersion;
```

Archives/ClmFile.cpp
```diff
@@ -456,10 +398,31 @@ namespace Archives
...
+	};
+
+	bool ClmFile::ClmHeader::CheckUnknown() const {
+		return std::memcmp(unknown.data(), standardUnknown.data(), unknown.size()) == 0;
```
DanRStevens: Ditto

Archives/ClmFile.h
```diff
@@ -30,48 +31,18 @@ namespace Archives
...
-			uint32_t length;
-		};
-
-	    // http://soundfile.sapp.org/doc/WaveFormat/
-		struct WaveHeader
-		{
-			RiffHeader riffHeader;
-			FormatChunk formatChunk;
-			ChunkHeader dataChunk;
+			bool CheckFileVersion() const;
```
DanRStevens: I like how you declared these methods as `const`.

Archives/VolFile.cpp
```diff
@@ -27,18 +27,18 @@ namespace Archives
 
 
 
-	const char* VolFile::GetInternalFileName(int index)
+	std::string VolFile::GetInternalFileName(int index)
```
DanRStevens: Yes, this makes a lot of sense. I can see how this changes ties in with other changes. The null termination issue for CLM internal names makes a `const char*` much more awkward to work with.

Archives/ArchivePacker.cpp
```diff
@@ -46,7 +51,11 @@ namespace Archives
 	bool ArchivePacker::ReplaceFileWithFile(const char *fileToReplace, const char *newFile)
 	{
 		MoveFileA(fileToReplace, "Delete.vol");
```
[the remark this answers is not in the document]
Brett208: When I implemented ADD and REMOVE using OP2Archive, I didn't use ArchivePacker::Repack. Instead I followed the following workflow:

1. Create a temp directory
2. extract contents of vol file to temp directory
3. modify temp directory files by either adding files from other sources or removing files
4. Delete original volume.
5. Pack new volume with original name.

I would be fine removing Repack and this function altogether from the library. 

Here is the function to create a random directory:

```
string ConsoleHelper::CreateTempDirectory()
{
	srand(static_cast<int>(time(NULL))); // For creating a unique directory.
	int number = rand();

	string directory("./OP2ArchiveTemp-" + to_string(number));

	XFile::NewDirectory(directory);

	return directory;
}
```

Archives/ClmFile.cpp
```diff
@@ -2,104 +2,69 @@
...
+		try {
+			clmFileReader = std::make_unique<FileStreamReader>(fileName);
+		}
+		catch (std::exception& e) {
 			throw std::runtime_error("Could not open clm file " + std::string(fileName) + ".");
 		}
 
-		if (ReadHeader() == false) {
-			throw std::runtime_error("Invalid clm header in " + std::string(fileName) + ".");
+		try {
```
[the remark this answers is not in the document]
Brett208: I designed catching the exception this way so the beginning part of the error message is the same no matter where the exception in the header was found. It doesn't bother me to store the filename at the class level or rewriting the exceptions to not require catching before passing again. The alternative would be to pass the filename as an argument into the function.

Archives/ClmFile.cpp
```diff
@@ -2,104 +2,69 @@
...
+	// Throws an error if packed file index is not valid.
+	std::string ClmFile::GetInternalFileName(int index)
 	{
 		CheckPackedFileIndexBounds(index);
 
-		return m_FileName[index];
+		return indexEntries[index].GetFileName();
 	}
 
 	int ClmFile::GetInternalFileIndex(const char *internalFileName)
```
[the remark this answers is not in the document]
Brett208: I must be missing what you are implying. I thought the update was using a std::string by using IndexEntry::GetFileName?

Archives/ClmFile.cpp
```diff
@@ -134,41 +98,32 @@ namespace Archives
 
 		try
 		{
-			FileStreamWriter fileStreamWriter(pathOut);
+			FileStreamWriter waveFileWriter(pathOut);
 
-			fileStreamWriter.Write((char*)&header, sizeof(header));
+			waveFileWriter.Write((char*)&header, sizeof(header));
```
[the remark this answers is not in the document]
Brett208: I like the template idea and will investigate implementing it. This reminds me of generics in C#. 

I'm not sure of the pros and cons of using char* vs void* for the buffer. If it seems more correct to you, let's switch to void* but I would prefer switching implementations from char* to void* in the master branch as it will cause a large number of code changes that are unrelated to this branch.

Archives/ClmFile.cpp
```diff
@@ -134,41 +98,32 @@ namespace Archives
...
 
-				if (totalSize + numBytesRead > m_IndexEntry[fileIndex].dataLength) {
-					numBytesRead = m_IndexEntry[fileIndex].dataLength - totalSize;
+				// Check if less than CLM_WRITE_SIZE of data remains for writing to disk.
+				if (offset + numBytesToRead > indexEntries[fileIndex].dataLength) {
+					numBytesToRead = indexEntries[fileIndex].dataLength - offset;
 				}
 
-				totalSize += numBytesRead;
+				clmFileReader->Read(buffer.data(), CLM_WRITE_SIZE);
```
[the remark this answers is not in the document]
Brett208: Nice catch. Using CLM_WRITE_SIZE there is a bug. I'm assuming that std::ifstream allows attempting to read past the end of the buffer just like the Windows API, which prevented an error from occuring. Although this was not the intent. I'll fix the bug soon.

MemoryStreamReader::Read does throw a bug when trying to read past the end of the buffer. I believe it makes more sense to force throwing an error on FileStreamReader::Read when going past the end of the stream as well. This way the same behaviour is encountered no matter which child of SeekableStreamReader is used. I'll fix this as well.

Archives/ClmFile.cpp
Brett208: Okay, I'll change it back. I was trying to make it more readable but I see where you are coming from. Lets just not plan on making a WaveFormatExEx if we can help it!

Archives/ClmFile.cpp
Brett208: I was thinking of eventually separating ClmFile into ClmWriter and ClmReader. This could be ClmPacker and ClmUnpacker as well. I hadn't thought about using static functions for writing the CLM File. Would you prefer to see the ClmFile class remain with both packing and unpacking available? If we want to separate into 2 classes, it might be best to wait until later to separate them.

I'll go ahead and mark this function as static in any case.

Archives/ClmFile.cpp
Brett208: I'll try this out since it would be easier to read. than using memcpy.

Archives/ClmFile.cpp
```diff
@@ -2,104 +2,69 @@
...
 	{
-		char buff[32];
-		unsigned long numBytes;
-		int i;
-
-		if (ReadFile(m_FileHandle, buff, 32, &numBytes, nullptr) == 0) return false;
-		if (memcmp(buff, "OP2 Clump File Version 1.0\x01A\0\0\0\0", 32)) return false;
+		clmFileReader->Read((char*)&clmHeader, sizeof(ClmHeader));
+		
+		if (!clmHeader.CheckFileVersion()) {
```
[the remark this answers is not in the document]
Brett208: I suppose we can have both a check and a verify function. The verify function can just wrap the check function and throw the exception.

Brett208:
Thanks for the comments. I'll take a crack at implementing them in the near future. The only thing I'm hesitant about is spending a lot of time in ArchiveUnpacker right now. It might be better to focus on removing Windows.h from VolFile first before overhauling ArchiveUnpacker?

I just realized that I can open a review instead of doing single comments which might keep from flooding inboxes so badly. I'll try that out next time.

-Brett

Archives/ArchivePacker.cpp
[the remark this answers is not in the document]
DanRStevens: Perhaps we should revisit this at a future time then. Which makes sense, since that part wasn't even changed in your commits, just something I noticed while reviewing.

Archives/ClmFile.cpp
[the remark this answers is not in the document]
DanRStevens: I think I have a slight preference for storing the filename as a class member. Consistent phrasing of error messages is nice. Related, is to have specific exception classes if you want to monitor for exceptions from a specific source. Though if the stream open throws its own exception, that won't be CLM specific.

Archives/ClmFile.cpp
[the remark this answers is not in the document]
DanRStevens: That allows the most flexibility. I often do that.

The decision for me, is based on possible use cases. If checking something results in a normal control flow decision, you should have a method returning a value. It's bad design to use exceptions for normal program flow control. If that happens, or is likely to happen, you need a function with a proper return value. On the other hand, if you just need something to be true to continue, otherwise abort with an error, an exception raising function should be provided. Some checks are a bit split where it depends on context as to whether it constitutes a normal control flow decision, or an unexpected error condition. Those are the cases where I like to provide both, with one implemented in terms of the other.

My gut feeling here is there isn't a whole lot of reason for checking that doesn't lead to aborting with an error. Perhaps a get method, where the current value can be displayed, but the only time I see a check happening is when we abort further processing.

Archives/ClmFile.cpp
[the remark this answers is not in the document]
DanRStevens: This one is misleading, since the comment is on a non-changed line of code. This is about `const char *internalFileName`, not about `.GetFileName()`. Sorry for the confusion.

Archives/ClmFile.cpp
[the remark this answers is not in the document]
DanRStevens: Good call. Let's make these changes in another branch.

Archives/ClmFile.cpp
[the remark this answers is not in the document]
DanRStevens: Ahh. Well, parameter fix aside, the change in exception behaviour is probably work for another branch. We may also want to start a discussion on the various levels of exception safety. In particular, what happens to the stream pointer after a failed partial read.

Archives/ClmFile.cpp
DanRStevens: Yeah, trying to separate them seems like a lot more work, and I'm not sure it's entirely justified. I think we should move the packing to static methods while keeping it part of the same class. That should be easier to do, and it should eliminate the need for template files.

The framework I use for evaluating if something should be a static method, or a class, is the time frame over which operations are performed. If it's a single uninterruptible operation, such as pack, a static method makes sense. If instead you have an object representing a resource, such as a CLM file, and operations can be performed over time, such as periodically opening or reading from a new stream, then an object makes more sense. The CLM file is not typically opened, parsed, and loaded all into memory at once as some collection of wave objects, where upon the CLM file is then close and forgotten about. A CLM object for reading makes sense as you can perform multiple read operations over time, rather than a single one time operation.

DanRStevens:
You're right, in that some of my suggestions are more appropriate for future work in a new branch. Removing the dependence on Windows.h seems more important than some of those longer term ideas.

I haven't figure out how to open a review for replies to comments. Would be nice to know how to collect all that together in one go.

Archives/ClmFile.cpp
Brett208: I agree it is difficult to read. I think I would prefer modifying the structs somehow to make it easier to read/understand. I'm not sure what that would look like right now though.

Brett208:
Just updated this branch with changes. 

Last thing that needs looked at from my perspective is zero filling the fileName array before writing when the fileName is less than 8 characters.

I realized that ClmFile was not using the base variable m_VolumeFileName. This was utilized as opposed to creating a new variable to store it as discussed.

Items that were discussed in this branch that should be considered in future branches:

 * Switching Read/Write functions to use void* instead of char*
 * Looking into applying the Read/Write template functions to autodetect sizeof to the base StreamReader/Writer classes. Currently they only apply to the child classes File/MemoryStreamReader/Writer.
 * Standardize where StreamReader/Writer point when recovering from a failure. 
 * Reviewing edge cases and fail methods in ArchivePacker::Repack.
 * Making The Packing functions in ClmFile static so they do not rely on opening an existing ClmFile before packing a new one. Possibly separting ClmFile into ClmPacker/UnPacker.

Archives/ClmFile.cpp
```diff
@@ -34,27 +23,18 @@ namespace Archives
...
-		if (!clmHeader.CheckUnknown()) {
-			throw std::runtime_error("Unknown field in CLM header is incorrect.");
+		try {
+			clmHeader.VerifyFileVersion();
+			clmHeader.VerifyUnknown();
 		}
-
-		if (clmHeader.packedFilesCount < 1) {
-			throw std::runtime_error("Packed file count must be 1 or greater.");
+		catch (std::exception& e) {
```
DanRStevens: We can probably avoid using this catch. The source exceptions from the verify functions could include the CLM filename in their error message.

Archives/ClmFile.cpp
DanRStevens: Yes, lets defer. If we end up writing a Wave file class, that would be a good time to revisit this.

Archives/ClmFile.cpp
```diff
@@ -226,6 +202,15 @@ namespace Archives
...
 		// Write the archive header and copy files into the archive
 		try {
+			WaveFormatEx waveFormat;
+
+			if (waveFormats.size() > 0) {
+				waveFormat = waveFormats[0];
+			}
+			else {
+				waveFormat = CreateDefaultWaveFormat();
+			}
```
DanRStevens: This is a very reasonable intermediate step.

Archives/ClmFile.cpp
```diff
@@ -405,24 +390,57 @@ namespace Archives
...
+		waveFormat.wFormatTag = 1; // WAVE_FORMAT_PCM
+		waveFormat.nChannels = 1; // mono
+		waveFormat.nSamplesPerSec = 22050; // 22.05KHz
+		waveFormat.nAvgBytesPerSec = 44100; // nSamplesPerSec *nBlockAlign
+		waveFormat.nBlockAlign = 2; // 2 bytes/sample = nChannels * wBitsPerSample / 8
+		waveFormat.wBitsPerSample = 16;
+		waveFormat.cbSize = 0;
+
+		return waveFormat;
+	}
```
DanRStevens: This struct can probably be a static class member using the struct initialization syntax:
```
static const WaveFormatEx DefaultWaveFormat = {1, 1, 22050, 44100, 2, 16, 0};
```
You can split that into multiple lines if you want to retain the per field comments.

Archives/ClmFile.cpp
```diff
@@ -405,24 +390,57 @@ namespace Archives
...
 	};
 
 	bool ClmFile::ClmHeader::CheckUnknown() const {
-		return std::memcmp(unknown.data(), standardUnknown.data(), unknown.size()) == 0;
+		return unknown == standardUnknown;
+	}
+
+	void ClmFile::ClmHeader::VerifyFileVersion() const {
+		if (!CheckFileVersion()) {
+			throw std::runtime_error("CLM file version is incorrect.");
```
DanRStevens: Might make sense to include the CLM filename here, if it's available.

Archives/ClmFile.cpp
```diff
@@ -405,24 +390,57 @@ namespace Archives
...
+		}
+	}
+
+	void ClmFile::ClmHeader::VerifyUnknown() const {
+		if (!CheckUnknown()) {
+			throw std::runtime_error("Unknown field in CLM header is incorrect.");
```
DanRStevens: Ditto

Archives/ClmFile.cpp
```diff
@@ -405,24 +390,57 @@ namespace Archives
...
+		}
+	}
+
+	std::string ClmFile::IndexEntry::GetFileName() const {
+		char fileNameOut[9];
+		strncpy_s(fileNameOut, fileName.data(), 8);
+
+		return std::string(fileNameOut);
```
DanRStevens: This looks correct, and doesn't leak memory, though it does do a double string copy. There's probably a way to create a `std::string` with 9 characters of buffer space, and copy the `std::array` directly into that and then null terminate it.

This StackOverflow question on [Copying a std::array to a std::string](https://stackoverflow.com/questions/10900345/stdarraychar-n-to-stdstring) may help. It probably handles null termination for you.

StreamReader.cpp
```diff
@@ -20,6 +20,11 @@ FileStreamReader::~FileStreamReader() {
 }
 
 void FileStreamReader::Read(char* buffer, size_t size) {
+	// Unable to add a std::streampos to a size_t without casting.
+	if (static_cast<size_t>(file.tellg()) + size > Length()) {
```
DanRStevens: This looks correct, but causes seeking for each Read due to the implementation of `Length`. There's a potential for a large performance impact here, especially if the seeking affects buffering, such as by clearing read-ahead buffers.

It may be more efficient to seek back to the original position in the case of a failed or partial read than to pre-check.

It might also make sense to make no guarantees about the file pointer or buffer contents after a failed or partial read. This is the issue of exception safety. No undefined behaviour has occurred yet, but the object may have been left in an unknown state, making future calls to certain methods potentially unsafe. Not all methods though, as object destruction must necessarily still be safe.

It's nice to avoid undefined behaviour when possible. As such, I think I'd recommend pre-recording the position, wrapping the read in a try/catch, where the catch restores the file position and rethrows the same exception.

Archives/ClmFile.cpp
Brett208: Sounds good.

Archives/ClmFile.cpp
Brett208: I left the CLM's filename outside of the a ClmHeader for 2 reasons. 

1. It would add an extra data field that doesn't exist in an actual CLM file and complicate reading/writing a bit. 
2. When packing a new CLM file, the filename may not exist when the ClmHeader is created (well, this is only partially true since you probably want to specifiy the planned CLM filename before constructing it in memory.)

I'd like to leave the CLM's filename outside of the clmHeader struct.

Archives/ClmFile.cpp
Brett208: I was having difficulty getting from std::array to std::string with a proper null terminator without performing a double copy. I'll check the link you provided out and give it another shot as I don't like my current solution either.

StreamReader.cpp
Brett208: Agreed there is a performance issue. I was avoiding recording the length of the file separately in case the file size changed mid read. While this makes sense in the StreamWriter, it doesn't here because the file should not be changing size during a read only operation. I'll add the pre-recorded length.

I may wait to work on improved exception handling until a nother branch. Partially because I don't understand how std::ifstream itself throws exceptions/notifies of errors.

DanRStevens: Correct, the file can change size. Even if it's a read only stream, another process on the system might change the size of the file. Any set of check+do operations which are not atomic will always be subject to race conditions, and so can never ensure correctness. This is why it's best to react to the result of the read, rather than trying to pre-check it. 

I was thinking record the current position at the beginning of the read, and if there is an error during the read, to catch it and seek to the previously recorded position and then (re)throw an error. The performance impact is less, only a read of the current position for each call to Read, and only seeking to recover the internal stream state after an error.

Of course, now that I'm thinking about it, even that would be a non-atomic set of operations. If the file size changed because it was truncated to 0, the seek would not be very meaningful.

This is probably why some APIs choose to avoid the performance impact of trying to make guarantees under error conditions. Sometimes you can't really make a meaningful guarantee, so why bother take the performance hit. Just document what can and can't be relied upon during those circumstances.

If you wanted to go with no guarantee in the case of exceptions, I'd be willing to go with that. We'd just need to document it.

Archives/ClmFile.cpp
DanRStevens: I did some testing, and this seems like it should work:
```
std::string ClmFile::IndexEntry::GetFileName() const {
    return std::string(fileName.data(), fileName.size());
}
```
That should grab 8 chars from the buffer, put them into a `std::string` object, and null terminate it (max 9 chars in the string buffer, including the null terminator). If the string is shorter, such that it doesn't take up the full 8 characters, the earlier null will terminate the string.

It uses the `string` constructor:
```
string(const char* s, size_t n);
```

Brett208:
Just pushed another commit. Barring anything major, I'd like to merge this branch as progress is stagnating. 

There is some interesting stuff going on in IndexEntry::GetFileName. Dan's suggestion worked, but needed tweaking for ensuring the std::string's size didn't include multiple null terminators when they were present. It was causing comparison checks to fail. There may be a more elegant and faster way to complete this check.

Thanks,
-Brett

Archives/ClmFile.h
```diff
@@ -71,7 +71,7 @@ namespace Archives
 		void PackFile(StreamWriter& clmWriter, const IndexEntry& entry, StreamReader& fileToPackReader);
 		std::vector<std::string> StripFileNameExtensions(std::vector<std::string> paths);
 		void InitializeWaveHeader(WaveHeader& headerOut, int fileIndex);
-		WaveFormatEx CreateDefaultWaveFormat();
+		static WaveFormatEx CreateDefaultWaveFormat();
```
DanRStevens: Ahh. I see the miscommunication now. I meant the data structure itself could be a static class member. I suppose it makes sense for the function returning the data to be static, though I was thinking in terms of eliminating the function and just accessing the static const struct directly.

Granted my idea might not be quite perfect, in that it's questionable if the default format should be part of the class (representing the file format), or part of the front-end (representing what Outpost 2 expects for its use of this file format). Perhaps a decision for another day.

At any rate, your current solution is correct code. I'm just trying to clarify my original intention in the first comment. Any changes here are entirely optional.

StreamReader.h
```diff
@@ -40,6 +40,7 @@ class FileStreamReader : public SeekableStreamReader {
 
 private:
 	std::ifstream file;
+	uint64_t streamSize;
```
DanRStevens: This makes me a bit squeamish. We can't know for certain the stream size won't change due to some external process. I'm kind of thinking of reverting the recent size checks on the file stream, documenting the read past end of file behaviour as undefined for now, and opening a new issue to track this as a future problem.

Brett208:
IndexEntry::GetFileName seems to be working for me with both 8 char strings and less than 8 char strings. I'm not a fan of the iterator, but it seems necessary. I had to adjust StringHelper::RemoveStrings to work properly. Not sure if this was an existing bug in RemoveStrings or if IndexEntry::GetFileName is not quite correct yet. If there is a better way to accomplish this or behavior is incorrect, lets open this in a bug report.

FileStream changes for pre-computing length and checking FileStreamReader's length during read operations were reverted as counter-suggested. 

I hadn't realized that variables declared in a ranged based for loop could be declared const. I amended as appropriate within ClmFile and ArchivePacker.

Thanks,
Brett

DanRStevens:
And so the beast was merged. Hooray! And there was much rejoicing.

PR #35, opened 2018-06-03
Linux gcc fixes

DanRStevens:
A few minor fixes my compiler pointed out.


Brett208:
Compiled using MSVC without issues.

PR #38, opened 2018-06-04
Move stream helper templates to base classes

### LINKED ISSUE #37: Review StreamReader/Writer Read/Write template functions to autodetect sizeof buffer
Currently, FileStreamReader, MemoryStreamReader, FileStreamWriter, and MemoryStreamWriter contain a template function to autodetect the type being read into/out of and assign the size automatically. 

These functions cannot be used when referencing the base classes StreamReader, StreamWriter, SeekableStreamReader, and SeekableStreamWriter. I do not believe C++ supports virtual functions as being templates.

Perhaps there is a solution that would allow calling the template based Read/Write from the base classes that I am missing.

DanRStevens:
This fixes #37.

### DIFF
--- a/StreamReader.h
+++ b/StreamReader.h
@@ -8,6 +8,11 @@ class StreamReader {
 public:
 	virtual ~StreamReader() = 0;
 	virtual void Read(char* buffer, size_t size) = 0;
+
+	// Inline templated convenience methods, to easily read arbitrary data types
+	template<typename T> inline void Read(T& object) {
+		Read((char*)&object, sizeof(object));
+	}
 };
 
 class SeekableStreamReader : public StreamReader {
@@ -29,11 +34,6 @@ public:
 
 	void Read(char* buffer, size_t size);
 
-	// Inline templated convenience methods, to easily read arbitrary data types
-	template<typename T> inline void Read(T& object) {
-		Read((char*)&object, sizeof(object));
-	}
-
 	uint64_t Length();
 	void Seek(uint64_t position);
 	void SeekRelative(int64_t offset);
@@ -47,11 +47,6 @@ class MemoryStreamReader : public SeekableStreamReader {
 public:
 	MemoryStreamReader(char* buffer, size_t size);
 	void Read(char* buffer, size_t size);
-
-	// Inline templated convenience methods, to easily read arbitrary data types
-	template<typename T> inline void Read(T& object) {
-		Read((char*)&object, sizeof(object));
-	}
 
 	uint64_t Length();
 	void Seek(uint64_t position);
--- a/StreamWriter.h
+++ b/StreamWriter.h
@@ -8,6 +8,11 @@ class StreamWriter
 public:
 	virtual ~StreamWriter() = 0;
 	virtual void Write(const char* buffer, size_t size) = 0;
+
+	// Inline templated convenience methods, to easily read arbitrary data types
+	template<typename T> inline void Write(T& object) {
+		Write((char*)&object, sizeof(object));
+	}
 };
 
 class SeekableStreamWriter : public StreamWriter
@@ -24,11 +29,6 @@ public:
 	~FileStreamWriter();
 	void Write(const char* buffer, size_t size);
 
-	// Inline templated convenience methods, to easily read arbitrary data types
-	template<typename T> inline void Write(T& object) {
-		Write((char*)&object, sizeof(object));
-	}
-
 	// Change position forward or backword in buffer.
 	void SeekRelative(int offset);
 
@@ -44,11 +44,6 @@ public:
 	MemoryStreamWriter(char* buffer, size_t size);
 
 	void Write(const char* buffer, size_t size);
-
-	// Inline templated convenience methods, to easily read arbitrary data types
-	template<typename T> inline void Write(T& object) {
-		Write((char*)&object, sizeof(object));
-	}
 
 	// Change position forward or backword in buffer.
 	void SeekRelative(int offset);


Brett208:
I'm attempting to test this change by integrating it in ClmFile.cpp. Using Line 83 inside `ClmFile::ExtractFile`. A FileStreamWriter is being used to write a ClmHeader. Th version of the code in this branch produces the error below.

```
waveFileWriter.Write(header);
```

Error	C2660 'FileStreamWriter::Write': function does not take 1 arguments OP2Utility	c:\ . . . \op2archive\op2utility\archives\clmfile.cpp 83

Adding the template Write function into FileStreamWriter fixes the error. I tried cleaning the solution and rebuilding. Are we sure the template function can live in the base class only and work in all the child classes?

Thanks,
-Brett

DanRStevens:
The template methods should be inherited by the sub-classes. I don't get any errors on Linux, so I'm a little uncertain what is happening or how to address it. It could be a compiler difference, perhaps with one compiler diverging from the standard on this aspect.

The error might have made sense if `header` was declared `const`, since the template was not updated to accept a const until a later commit, though that doesn't seem to be the issue here.

Perhaps I should move this commit out of line with the others so it's not blocking other changes. We can revisit this later.

Brett208:
I spent a bit of time searching the web about this. Didn't come up with anything concrete.

I boiled the problem code down to a small sample project. I'd be curious to see if Linux squawks at all when compiling if you don't mind trying. I got 2 errors using MSVC which are listed in the comments of main.cpp.

[TemplateTest.zip](https://github.com/OutpostUniverse/OP2Utility/files/2074660/TemplateTest.zip)

-Brett

DanRStevens:
I get a warning about an unrelated mistake. I assume you meant to pass `&test` at the first argument. Otherwise you're casting 34 as the pointer value, which is almost certainly going to point somewhere invalid. Not that you're using the pointer in the sample code.

Other than that unrelated warning, it compiles cleanly and runs as expected.

DanRStevens:
Oh geeze, I missed that. Sorry, no, it gives an error for me too:

> g++ -std=c++11 *.cpp
main.cpp: In function ‘int main()’:
main.cpp:11:19: error: no matching function for call to ‘SizeOutputChild::Write(int*)’
  child.Write(&test); 
                   ^
In file included from main.cpp:1:0:
SizeOutput.h:19:7: note: candidate: virtual void SizeOutputChild::Write(const void*, size_t)
  void Write(const void* object, size_t size) override;
       ^
SizeOutput.h:19:7: note:   candidate expects 2 arguments, 1 provided

DanRStevens:
Ahh, you know what happened. I wasn't compiling the code in the Archives subfolder when I made this change. This is from before I added a Linux makefile from the linux-makefile branch. I was doing ad-hoc compiling and testing before, which in my carelessness only included the directory I was working in.

The change in my first commit does appear to break on Linux.

This will be easier to address once the Linux makefile change are incorporated from the other branch. A few other errors were fixed there. Let's revisit this after those changes have been incorporated.

DanRStevens:
Ok, I know what the problem is now. It's addressed in a StackOverflow question on [Template Member Function Inheritance](https://stackoverflow.com/questions/26575887/template-member-function-inheritance?rq=1).

The template and the virtual function have the same name. That alone is fine, however, when you derive from a base class, and re-declare the name to override the virtual function, the virtual function name will hide the template name.

One way to get around this problem is with a `using` clause:
```
class FileStreamReader : public SeekableStreamReader {
public:
	using StreamReader::Read;
```

However, the `using` clause would be required in every derived class, which is a pain. Perhaps acceptable, since the details are hidden within the Stream classes, however, there may be Stream derived classes written later that will exist outside the library.

Another solution is to use a different name between the template and the virtual function. This simply avoids the name hiding problem. Ideally, we'd want to avoid having two public names that do the same thing, so the virtual method could be marked `protected`. The template method name would have to be the public one to gain the benefits from it.

```
protected:
	virtual void ReadImplementation(void* buffer, size_t size) = 0;
public:
	// Inline templated convenience methods, to easily read arbitrary data types
	template<typename T> inline void Read(T& object) {
		ReadImplementation(&object, sizeof(object));
	}
	inline void Read(void* buffer, size_t size) {
		ReadImplementation(buffer, size);		
	}
```

Library users would see the public Read interface, while library implementers would write a protected ReadImplementation method. The public Read template that bridges the two would inherit properly, without hiding, and without needing explicit mention, in each derived class.

I'll pause here in case you have any thoughts on the proposed workarounds.

Brett208:
Nice job researching! I can accept either workaround. 

I think a protected virtual function appeals more than using statements in child classes. I think the protected virtual function would be easier to understand when extending the Stream base classes.

`ReadImplementation` is a bit clunky, but it is descriptive enough and I cannot think of anything better off the top of my head. I'm of the mindset to go ahead and use the name.

DanRStevens:
Darn, I was hoping you'd come up with a better name! I think that name is really clunky too.

I was tempted to use something shorter, like ReadImpl, but I hate excessive abbreviations, and feel they just serve to make code harder to read, without actually saving any real typing time. After all, if the typical programmer really is only able to write 8-16 lines of code per day on average, then programming speed is hardly dependent on typing speed. More like thinking speed, which is aided by clarity, not abbreviations.

I'm thinking I'll go with the protected member solution. We can rename the method later if we come up with something better.

Let's implement this after the file re-organization.

----

To add some additional notes, the issue of **name hiding** is related to the [Fragile Base Class](https://en.wikipedia.org/wiki/Fragile_base_class) problem. Changes to the base class can cause expected changes in behaviour to code using a derived class. The Wikipedia article gives an example of mutual recursion, which affects C++ as much as Java.

Another example of the fragile base class problem is related to changes in method overload resolution, which name hiding does protect from. An example being when a new method gets added to the base class, which is a better matching overload than a current call to a derived class method, and causes the new base class method to steal a call from an older derived class method. Perhaps one takes a `char` parameter while the other takes an `int` parameter.
```
Derived d;
d.someMethod(5); // Could match either parameter type `int` or `char`
```

Brett208:
I'm going to look at deleting this branch and redoing the work in a new branch. This one has become pretty stale and would require some fairly significant merge conflict resolution. It will probably be easier/quicker to just reapply the work on a clean branch.

DanRStevens:
I was thinking similar about this branch getting kind of stale. One way forward might be to reset the branch head to the current master branch, make entirely new updates, and then force push to overwrite the changes for this branch. There's no merge to worry about then. Plus, it would retain the above discussion in the same pull request.

PR #39, opened 2018-06-04
Seekable stream updates

DanRStevens:
Fleshed out the Seekable interfaces, and provided implementations for the File and Memory based stream classes.

This branch was based off the stream-template-helpers branch, which moved the template helper methods into the base class. When reviewing, it may be helpful to view the diff against that base (unless it's already merged into master).

### DIFF
--- a/StreamReader.cpp
+++ b/StreamReader.cpp
@@ -24,9 +24,9 @@ void FileStreamReader::Read(char* buffer, size_t size) {
 }
 
 uint64_t FileStreamReader::Length() {
-	uint64_t currentPosition = file.tellg();
+	auto currentPosition = file.tellg();
 	file.seekg(0, std::ios_base::end);
-	uint64_t streamSize = file.tellg();
+	auto streamSize = file.tellg();
 
 	file.seekg(currentPosition);
 
--- a/StreamReader.cpp
+++ b/StreamReader.cpp
@@ -24,13 +24,15 @@ void FileStreamReader::Read(char* buffer, size_t size) {
 }
 
 uint64_t FileStreamReader::Length() {
-	auto currentPosition = file.tellg();
-	file.seekg(0, std::ios_base::end);
-	auto streamSize = file.tellg();
-
-	file.seekg(currentPosition);
-
+	auto currentPosition = file.tellg();  // Record current position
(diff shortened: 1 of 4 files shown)


DanRStevens:
This should close #24.

Brett208:
Good idea adding the override keyword to the overriding function prototypes to improve readability. I didn't see any issues reading through the changes.

As the StreamReader/Writer classes continue to grow, at some point I would like to see them separated into different files. IE a MemoryStreamReader.cpp, FileStreamReader.cpp, etc. It might make sense to put them all in a separate folder at some point as well. Perhaps a folder named Serialization?

Thanks,
-Brett

DanRStevens:
Ok, partly as a test to see what would happen, I rebased the changes so they don't include the template changes, and forced pushed the updated branch (it has re-written history to remove the template related commit). Looks like it just appends a new set of changes to the thread.

I agree about splitting the classes into separate files, and moving them to a new folder. Let's do that in a future branch. I've opened Issue #43 to track that.

Brett208:
The compiler caught the following warning from MemoryStreamWriter::Seek on the line `this->offset = offset`. This warning I think is pre-existant from this changeset, but am recording here as reminder to fix in future.

Warning	C4244	'=': conversion from 'uint64_t' to 'size_t', possible loss of data

```
void MemoryStreamWriter::Seek(uint64_t offset)
{
	// Checking if offset goes below 0 is unnecessary. Arithmetic on a signed and unsigned number results 
	// in a signed number that will wraparound to a large positive and be caught.
	if (offset > streamSize) {
		throw std::runtime_error("Change in offset places read position outside bounds of buffer.");
	}

	this->offset = offset;
}
```

Brett208:
I tested this branch by:

1. Creating a new CLM archive with 26 files.
2. Extracting a full CLM archive.
3. Listing contents of CLM archive.

No unexpected behaviour noticed so I'm going to merge and close #24.

Thanks,
Brett

DanRStevens:
Thanks for mentioning the MSVC warning. That could be silenced with an explicit cast. The check above already ensures it can't be out of bounds, so it should be safe to do, though we might want to mention that in a comment before the cast.

Issue #36, opened 2018-06-03
Switch StreamReader/Writer to use void* when representing a buffer


Brett208:
Currently char* is used to represent buffers. Perhaps void* would be more appropriate?

I'm not actually certain what the pros and cons of each choice are.

DanRStevens:
It's not a big difference in semantics, though `void*` is a little more correct. With `void*` it is a pointer to an untyped memory location. A buffer can represent anything, and with streams the data is sliced into objects of a variety of types. With `char*` there's a suggestion the buffer points to characters.

As characters are the closest datatype to a byte, it's still fairly natural to represent buffers as `char*`. It also allows for the convenience of pointer arithmetic, such as `buffer + offset`. In C/C++, the syntax `pointer + number` actually represents the address `pointer + number * sizeof(dataType)`. Hence if the datatype is `char`, the arithmetic works out nicely in terms of bytes. If you try that with a `void*` it's an error as there is no well defined `sizeof` to work with. You would need to cast to another pointer type to perform arithmetic.

In C, you can copy both ways between `void*` and other pointer types without casting. In C++ it's a little more restrictive, in that you can copy into a `void*` without casting, but you need a cast to convert back out of it.

Also noteworthy is that C did not originally have the `void*` type, and so `char*` was the standard way to represent a pointer to a buffer before the ISO/ANSI C standard introduced `void*` as a type.

The trade-offs:
`char*`: allows pointer arithmetic without casting, but requires casting for type conversion.
`void*`: type conversions are cast-free (both ways for C, to `void*` for C++), but requires casting for pointer arithmetic

For an API, having `void*` in the interface simplifies usage for the caller, while pointer arithmetic can be done internally with casting, making for a nicer interface.

Example:
```
#include <iostream>

void Func(void* data) {
  std::cout << "void* data = " << std::hex << data << std::endl;
}

int main() {
  char buff1[4];
  int buff2[4];

  // Func can take pointers to any type without casting
  Func(&buff1);
  Func(&buff2);

  // One-way cast-free type conversions
  void* ptr1 = &buff1;
  //char* ptr2 = ptr1; // Error with default checks, warning with -fpermissive (g++)

  return 0;
}
```

Brett208:
@DanRStevens, Thank you for quality writeup on the issue. 

[...]

PR #40, opened 2018-06-04
Void ptr streams

DanRStevens:
This changes Stream interfaces to use `void*` instead of `char*`.

This should allow the caller to pass a pointer to any datatype without needing to cast at the call site. This simplifies usage of the library interface.

This is based on the seekable-stream-updates branch, which in turn is based on the stream-template-helpers branch. When reviewing, it may be easier to check the diff from that base (assuming it hasn't already been merged into master).

### DIFF
--- a/StreamReader.cpp
+++ b/StreamReader.cpp
@@ -19,8 +19,8 @@ FileStreamReader::~FileStreamReader() {
 	file.close();
 }
 
-void FileStreamReader::Read(char* buffer, size_t size) {
-	file.read(buffer, size);
+void FileStreamReader::Read(void* buffer, size_t size) {
+	file.read(static_cast<char*>(buffer), size);
 }
 
 uint64_t FileStreamReader::Length() {
@@ -50,7 +50,7 @@ MemoryStreamReader::MemoryStreamReader(char* buffer, size_t size) {
 	position = 0;
 }
 
-void MemoryStreamReader::Read(char* buffer, size_t size) 
+void MemoryStreamReader::Read(void* buffer, size_t size)
 {
 	if (position + size > streamSize) {
 		throw std::runtime_error("Size of bytes to read exceeds remaining size of buffer.");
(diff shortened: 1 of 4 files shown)


Brett208:
Based on comments in the #36, I think I agree with switching from char* to void*. Will merging this branch into the master break current usage of the StreamReader/Writer classes in the codebase that assume passing a char* buffer? I don't remember doing any pointer arithmetic in current code base. I'd rather not leave the codebase broken when the merge is made.

There is a typo in the commit message for Fix FileStreamWriter to seek using put pointer. Second to last word should read pointer instead of pinter. How difficult is it to rewrite a commit message in Git? Not sure fixing it is worth the effort though.

Thanks,
Brett

DanRStevens:
Changes should not break current usage. Existing `char*` casts or variables will be silently converted to `void*` during the call, and everything will work.

The one downside here, is a lot of code still has an explicit cast to `char*` which will no longer be needed, and the compiler will not issue any warnings or errors due to the silent conversion to `void*`. This is one reason why casting is discouraged, since it effectively disables a lot of compiler checks. We should at some point remove the `char*` casts at the points of call. We could temporarily change the interface method to a type other that `char*` or `void*` as a hack to have the compiler flag all these call points.

There is some internal pointer arithmetic when indexing into the internal buffer, though that is already accounted for in these commits. The internal buffer pointer is still a `char*`, with a `static_cast` converting the type inside the constructor. It was either there, or in the Read/Write method, where the pointer arithmetic was performed.

I'll rebase this patch on master so it's not dependent on the other commits from the base branches. Though there will be merge conflicts, and in the interests of not having to resolve them all multiple times, lets push this off for a bit to allow the other changes a chance to settle. One of those changes was fixing the typo in the commit message of the parent branch.

Let's revisit these changes later.

DanRStevens:
Here are the rebased changes, which should apply cleanly.

Brett208:
This branch compiles and executes fine when running EXTRACT and CREATE commands on CLM Files.

I also temporarily pulled the (char*) cast off of a write operation and everything worked.

Merging and closing the branch.

I'll open a new issue for removing unnessaray char* casts from MapWriter.cpp, MapReader.cpp, and ClmFile.cpp.

PR #41, opened 2018-06-04
Make stream writer template param const

DanRStevens:
Allow write template helper to be used with const objects.

This is based off the branch void-ptr-streams, which in turn is based off of other branches. When reviewing, it may be helpful to view the diff against that base (unless it's already merged into master).


DanRStevens:
Rebased on recent changes, with the problem template move commit removed.

PR #42, opened 2018-06-05
Linux makefile

DanRStevens:
Add a Linux makefile. This should help accelerate and standardize testing on Linux, which was previously done with ad-hoc compile commands.

Includes fixes a few other issues I didn't catch with my initial Linux testing, since I wasn't trying to build files from sub-folders with the ad-hoc testing.

### DIFF
--- a/Archives/VolFile.cpp
+++ b/Archives/VolFile.cpp
@@ -479,7 +479,7 @@ namespace Archives
 		for (temp = 0; temp < m_NumberOfIndexEntries; temp++)
 		{
 			// Make sure entry is valid
-			if (m_Index[temp].fileNameOffset == -1)
+			if (m_Index[temp].fileNameOffset == -1u)
 				break;
 		}
 		m_NumberOfPackedFiles = temp;		// Store the number of used index entries
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 x64
 Debug
 Release
+build/
 /*.user
 .vs
--- a/Maps/MapReader.cpp
+++ b/Maps/MapReader.cpp
@@ -2,6 +2,7 @@
 #include "../StreamReader.h"
 #include <iostream>
 #include <stdexcept>
+#include <cstring>
 
 MapData MapReader::Read(std::string filename, bool savedGame)
 {
@@ -61,7 +62,7 @@ void MapReader::ReadTilesetHeader()
 	char buffer[10];
 	streamReader->Read(buffer, sizeof(buffer));
 
-	if (strncmp(buffer, "TILE SET\x1a", sizeof(buffer))) {
+	if (std::strncmp(buffer, "TILE SET\x1a", sizeof(buffer))) {
 		throw std::runtime_error("'TILE SET' string not found.");
 	}
 }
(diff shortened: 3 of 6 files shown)


Archives/VolFile.cpp
```diff
@@ -479,7 +479,7 @@ namespace Archives
 		for (temp = 0; temp < m_NumberOfIndexEntries; temp++)
 		{
 			// Make sure entry is valid
-			if (m_Index[temp].fileNameOffset == -1)
+			if (m_Index[temp].fileNameOffset == -1u)
```
Brett208: What does -1u mean here? I thought u casts a number to be unsigned. Here we are assigning 1 to be negative after saying it is unsigned?

I'm also not sure about the test in general now that I look at it. fileNameOffset is an unsigned value, so checking it for equivalence against -1 seems odd to me. I'm guessing -1 here means the largest unsigned value?

Brett208:
I posted a question on -1u since it isn't something I've seen before. Otherwise, I think this is a good idea.

In the long term I'd like to consider cMake for the build system, which would mean eliminating both the Visual Studio solution and the makefile from the code base. However, that is a long way off and the makefile makes good sense in the meantime.

DanRStevens:
Agreed on wanting a cross platform build solution. I was hesitant to include a makefile for that reason, which was how I got into that mess with the template code move earlier. Though you're right, a cross platform solution is a bit of a ways off.

As for the makefile, it's a fairly generic build-all-C++-source-files-in-subtree makefile I had sitting around, which needed only minor adaptation and cleanup to the header section to work with this project. It's generic, with no references to specific source files or folders, so there should be no maintenance burden to keep it up to date with project file or structure changes.

If you want to change out the `-1u` for `UINT_MAX`, feel free to go ahead, or ask and I can update it in a future commit.

Archives/VolFile.cpp
Brett208: Thanks for the rundown on -1u. I'm happy if it stays, esp if there is precedence. It would be nice to add a short comment to explain it for anyone unfamiliar. Funny that it fixes a warning on your compiler as it causes a warning on MSVC.

Warning C4146 unary minus operator applied to unsigned type, result still unsigned OP2Utility c:\ . . . op2archive\op2utility\archives\volfile.cpp 482

Brett208:
@DanRStevens 

It sounds like I shouldn't need to do anything when adding source files and directories to the project to keep the makefile up to date. How do you want to handle adding new files on your side as they will need to be added to the Visual Studio project file and filter file? Or do you just want me to be the one that commits new directories and files as a temp fix until we get cMake up and running?

The following line produces a warning on MSVC. We can deal with it in a different committ in addition to perhaps adding a comment on why -1u is being used.

```if (m_Index[temp].fileNameOffset == -1u) ```

Warning C4146 unary minus operator applied to unsigned type, result still unsigned OP2Utility c:\ . . . op2archive\op2utility\archives\volfile.cpp 482

I tested this branch by using it when running OP2Archive. No issues noted when compiling and executing a CREATE operation on a new CLM file.

Thanks,
Brett

DanRStevens:
Ahh, I see what you're saying now. If the changes causes a warning on MSVC, then we should change the code. Ideally code should compile without warnings. Let's go with the `UINT_MAX` and `<climits>` solution then. That should work the same, and compile without warnings on any compiler. We can do that in a separate branch.

PR #45, opened 2018-06-06
Map reader writer updates

DanRStevens:
This fixes reference issues with the MapReader and MapWriter classes.

A non-const lvalue-reference is taken by MapWriter, but a temporary was being passed in, which causes an error on conforming compilers. The workaround is to create a local variable.

The MapReader used `unique_ptr` in the interface, which burdens the caller. The workaround is the same as above, using lvalue-references and named variables.

### DIFF
--- a/Maps/MapReader.cpp
+++ b/Maps/MapReader.cpp
@@ -6,16 +6,13 @@
 
 MapData MapReader::Read(std::string filename, bool savedGame)
 {
-	return Read(std::make_unique<FileStreamReader>(filename), savedGame);
+	FileStreamReader mapReader(filename);
+	return Read(mapReader, savedGame);
 }
 
-MapData MapReader::Read(std::unique_ptr<SeekableStreamReader> mapStream, bool savedGame)
+MapData MapReader::Read(SeekableStreamReader& mapStream, bool savedGame)
 {
-	if (!mapStream) {
-		throw std::runtime_error("Provided map or save stream does not exist or is malformed.");
-	}
-
-	streamReader = std::move(mapStream);
+	streamReader = &mapStream;
 
 	MapData mapData;
 
--- a/Maps/MapReader.h
+++ b/Maps/MapReader.h
@@ -10,11 +10,11 @@ class SeekableStreamReader;
 class MapReader
 {
 public:
-	MapData Read(std::unique_ptr<SeekableStreamReader> mapStream, bool savedGame = false);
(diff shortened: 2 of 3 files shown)


DanRStevens:
This related to the discussion in Issue #47.

Brett208:
Tested this code using a local modified OP2MapImager. It successfully images a map and saves a new map to disk. No compiler warnings or errors from MSVC. 

I think what you mean by converting the internal implementation to lvalue is switching `SeekableStreamReader*` to `SeekableStreamReader&`. 

Strangely, I had never though to declare a variable using & in this manner. I've only used it in function arguments and when converting an existing argument if that makes sense.

If we can change the private reference `SeekableStreamReader*` to `SeekableStreamReader&` while only forward declaring The SeekableStreamReader class instead of including the entire StreamReader header file, I would like to switch the internal implementation. This will make it a little easier to read as it will eliminate all the -> calls and when I see pointers, I'm typically looking for the an associated destructor even though one isn't needed here.

Thanks,
Brett

Brett208:
I will use a Rebase and merge for this branch. A few of the other merges today would have been more appropriate as rebase and merge but I made them merge commits. I'll try to be more careful before just pushing the 'Create a merge commit' button.

DanRStevens:
Correct about converting the internal implementation.

Forward declaration is sufficient to declare either a pointer (`*`) or reference (`&`). They are implemented the same under the hood. The compiler only needs to know the size of a pointer to make room for such variables. The full declaration is only needed when the pointer is de-referenced, such as by using `->`, or `.`. That will only occur in the implementation file (.cpp), or inline methods in a header. The full declaration is also needed for object construction, such as non-pointer non-reference member variables, or using `new`.

----

I tried cleaning up my local branch after this merge, and I noticed Git complain this change was not fully merged. I believe this is related to the rebase and merge, as the rebase rewrites history, which changes commit IDs. It may also have to do with the merge order. Not sure if you've encountered this with your work. I can just force delete the old branch, though I'm hesitant to adopt a long term workflow that requires force deleting things.

I'm also noticing GitHub doesn't say anything about deleting the branch. It does show the pull request as closed. I guess it makes sense, since you wouldn't want to do multiple rebase and merge, or multiple squash and merge commits from the same branch into master. You might do that with a merge commit merge though, where you periodically bring two branches in sync with each other.

Let's continue to experiment until we figure it out.

Issue #46, opened 2018-06-06
MapReader/Writer can be updated to use references


DanRStevens:
These classes do not expect the map file stream to ever be null. As such, internal pointers within these classes can be converted to references. This allows use of the `.` syntax instead of the `->` syntax.

Currently, such changes will produce compiler warnings about uninitialized references since the MapReader/Writer classes have no constructor. This might potentially be addressed by making the Read/Write methods the constructors.

As an intermediate step, I propose we refactor the code to use references, and make the public methods constructors. That should addresses the pointer/reference change, and the compiler warning message about uninitialized references in a class with no constructor.

However, currently the MapReader and MapWriter classes are being used more like a namespace than a class. The one class features, is how it private methods are auto passed a hidden parameter to the stream being parsed. From the outside, these classes each have effectively one public method, which is expected to be called exactly one time. There doesn't make a strong case for a class. Instantiating this class is similar to just calling a function.

Maybe we should refactor to a namespace, and be explicit about the parameter being passed to the private methods.

PR #48, opened 2018-06-06
Convert MapReader/Writer from class to namespace

DanRStevens:
At this point, this pull request is more to start a code review and discussion about a work in progress relating to issue #46. I don't recommend it be merged at this point, as the code generates warnings.

This branch includes changes from the map-reader-writer-updates branch, as it depends on those changes.

The question, is how might these changes proceed, and the warnings be resolved. One possibility is to refactor the Read/Write methods into MapReader/MapWriter constructors. Another possibility is to refactor the MapReader/MapWriter into namespaces rather than classes. The code might also be left as is, or undergo some other refactoring.

An example warning from the current intermediate step is:

> Maps/MapReader.h:17:24: warning: non-static reference ‘SeekableStreamReader& MapReader::streamReader’ in class without a constructor [-Wuninitialized]
  SeekableStreamReader& streamReader;

### DIFF
--- a/Maps/MapReader.cpp
+++ b/Maps/MapReader.cpp
@@ -12,7 +12,7 @@ MapData MapReader::Read(std::string filename, bool savedGame)
 
 MapData MapReader::Read(SeekableStreamReader& mapStream, bool savedGame)
 {
-	streamReader = &mapStream;
+	streamReader = mapStream;
 
 	MapData mapData;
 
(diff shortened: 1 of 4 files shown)


DanRStevens:
I do not recommend using a static here. A static is basically a global variable, and that doesn't make sense for this code. In particular, using static and global variables has implications for threading (possibly relevant here some day), and re-entry (not really relevant here since maps don't contain other maps). It's not possible to process two maps at the same time if we use statics or globals.

Short term: I recommend converting the Read and Write methods into constructors for the MapReader and MapWriter classes.

Long term: To be determined.

Brett208:
So does this mean if we move Archive::CreateArchive to a static set of functions, we would have the same threading issues when creating archives as we do reading/writing maps here?

Is there a precedence for creating a class that only operates when its constructor is called? I've never heard of a paradigm like this before. It seems a bit weird to me. Not that we cannot do something that is off the beaten path.

DanRStevens:
Static functions are fine. It's static data that's the problem. If functions use local variables, it's not an issue. Each call gets their own copy. Sharing of variables between independent calls is what creates the problem.

I'm sure I've probably tried creating constructor only classes before, though mostly just experimenting and not really fully realizing quite what I was doing, or why I was doing it that way. In hindsight, it felt like an abuse of the OO-paradigm. Sort of like how Java tries to force everything to be a class. It kind of goes too far. OO has it's uses, but if you're not playing to the strengths of OO, it might not be worth trying to force a problem into an OO solution.

Brett208:
Okay, that makes sense about the static data. If multiple threads call the static function simultaneously, there is no guarantee they won't both be modifying the static data.

I was looking at the function prototypes for MapReader and MapWriter. They are all currently monadic. Making them all dyadic functions by passing in the applicable stream really isn't that bad. My earlier comment about bloating the arugments was out of proportion. Perhaps the best course would be making 2 namespaces, one for the reader and one for the writer. 

If we want to go the namespace route, I think it would be smart to move all the currently private function prototypes in the header file to the top of he .cpp file. This way only the single public function will be declared in the header file and it will be easy to figure out how to use the library and harder to abuse the library by calling the previously private functions.

Thanks,
Brett

DanRStevens:
Alright, agreed. In particular, the part about the private functions only going into the .cpp file makes a lot of sense.

You surprised me there with "monadic" and "dyadic" :nerd_face:

DanRStevens:
Ahh, I think I know what was holding that other branch open now. This pull request had the other branch set as the merge destination, rather than master. I had done that to get a cleaner simpler diff of changes unique to this branch.

Brett208:
Yeah, I thought about commenting on the data sizes as well. Might be smart to take care of the read/write template branch first and then some of the data size cleaning can just be removing the size from the Read/Write function input and autocalculated by the template function.

Oh yeah, I learned the terms Monadic and Dyadic functions from Clean Code by Robert Martin (https://www.amazon.com/Clean-Code-Handbook-Software-Craftsmanship/dp/0132350882/). Probably shapes some of my coding conventions.

My local copy of this branch is warning me on delete that I have not fully merged it into the master. I performed a rebase and merge commit. This would support the theory that the warning originates from making commits on the master between branch creation and the rebase and merge.

DanRStevens:
Exactly what I was thinking. Need to prioritize updating that template code.

The book looks interesting. I think I'll check that out.

Perhaps we should consider doing more regular non-fast-forward merge commit merges.

PR #49, opened 2018-06-06
Eliminate use of manual memory management

DanRStevens:
This code contained a bug, mixing new[] with delete instead of delete[].
This could potentially cause memory corruption under some C++ runtimes.

New code uses std::string with reserve to avoid manual memory
management. This is more correct, and simpler.

### DIFF
--- a/Maps/MapReader.cpp
+++ b/Maps/MapReader.cpp
@@ -141,27 +141,12 @@ TileGroup MapReader::ReadTileGroup()
 // String must be stored in file as string length followed by char[].
 std::string MapReader::ReadString()
 {
-	size_t stringLength;
+	uint32_t stringLength;
 	streamReader->Read(&stringLength, sizeof(stringLength));
 
-	if (stringLength == 0) {
-		return std::string();
-	}
+	std::string s;
+	s.resize(stringLength);
+	streamReader->Read(&s[0], stringLength);
 
-	char* cString = new char[stringLength];
-
-	try
-	{
-		streamReader->Read(cString, stringLength);
-		std::string s(cString, stringLength);
-		delete cString;
-
-		return s;
-	}
-	catch (std::exception e)
-	{
-		delete cString;
-
-		throw e;
-	}
+	return s;
 }


Brett208:
This compiles without complaints but produces unexpected results when reading a map. Error thrown is bad allocation.

When attempting to read Ashes.map, the first tilesetSource is read as well0000ÌÌÌÌÌÌÌÌ.

I can probably help troubleshoot this some more in the future, but out of time for now.

I would like to see this function cleaned up though as my implementation was poor.

Thanks,
Brett

DanRStevens:
Ok, this might work better. I used `resize` instead of `reserve`. The `resize` method also updates the string length. The one downside is it wastes time initializing the buffer with nulls, only to later fill it with data read from the file. I didn't see any standard library methods that update the size without also initializing data, which I suppose kind of makes sense considering the design of `std::string`.

Thanks for the bug report.

DanRStevens:
I updated this to resolve merge conflicts so it would be easier to test and (possibly) apply.

Brett208:
I read the comment about how to fix it, but I didn't realize the code base had been updated. I was waiting to see another commit message pop with the fix in it. Did you delete the original buggy commit?

I just tested the branch again. It reads the map fine now and produces a nice render using OP2MapImager. I step debugged through just to check and a couple of strings were read properly.

Thanks,
Brett

DanRStevens:
Ahh, I can see how that would be missed. Especially with my phrasing of the message.

I did a rebase on master, and then force pushed to the branch, so it would have overwritten the previous commit.

Any Git oddities due to the rebase on your end when checking out and testing the branch? What about deleting the local branch copy? I'm wondering how the rebase affects other people.

Related, I noticed one of our other rebased and merged branches hasn't been deleted, nor does it have the option to delete it from the GitHub pull request page. From the branch page it says that branch can't be deleted because an open pull request references it. After some investigation, I think it's because another open pull request was built off the same branch. I guess we'll find out soon enough.

Brett208:
Occasionally I've seen branches that warn they are not fully merged to HEAD even though they have been. I haven't figured out what makes this occur yet, but haven't investigated closely. I've checked and haven't noticed any missing code (besides that bad merge conflict on the stream code). Although it seems bad practice to not address this as ignoring this warning could cause deleting branches that are not fully merged.

I'm okay with rebase and force push, esp on non-master branches, but I would prefer a comment stating it has occurred. I'm not aware of a way to tell that the force-push has occurred without closely examining the commits. It can cause weirdness on my side if I'm not expecting it and have the deleted commit downloaded. I typically always read a branch/merge request comments before pulling and testing so I would know to expect the force push.

DanRStevens:
Agreed about not wanting a workflow where we get used to ignoring warning messages. I just know that will become a problem someday. Plus it wastes time checking history because I'm extra careful about deleting things when I see the warning.

I believe the warning is seen when the commit id is different between the branch and master. The commit id is a hash of various commit info, such as the commit time, and parent commit id. Any updates or changes to history will change the commit id. Hence, rebase and squash could change commit ids by virtue of changing the parent the diff is applied to.

Rebase and squash merges don't always lead to warnings though. I think that's because they don't always cause history rewrites. In particular, if a branch was based off of master, and is merged back in with no intervening commits, there is no new history to rebase on top of. In that case you get a simple fast forward merge.

I wonder what happens if we delay deleting the feature branch until both people have had a chance to fetch new changes that incorporate the merge.

PR #50, opened 2018-06-06
Use UINT_MAX in check for invalid index entry

DanRStevens:
Alright, alright, one last fix. Now I'm going to bed. :smiley: 

Removes the `-1u` that was causing the compiler warning, and replaces it with `UINT_MAX` and an include for `<climits>`.


Brett208:
This change compiles without warnings on MSVC. Also, I feel like UINT_MAX is easier to understand than seeing -1u so a win for readability. Only downside is bringing in the new header file.

Thanks,
Brett

Issue #51, opened 2018-06-07
Remove unnecessary char* casts from calls to Stream Read/Write


Brett208:
Remove now unnecessary char* casts from calls to Stream Read/Write due to changing Stream Read/Write buffers to void*. All calls should be limited to files MapWriter.cpp, MapReader.cpp, and ClmFile.cpp.

PR #52, opened 2018-06-07
Remove `char*` cast for stream reading and writing

DanRStevens:
This addresses Issue #51.

### DIFF
--- a/Archives/ClmFile.cpp
+++ b/Archives/ClmFile.cpp
@@ -21,7 +21,7 @@ namespace Archives
 	// Throws an error is problems encountered while reading the header.
 	void ClmFile::ReadHeader()
 	{
-		clmFileReader->Read((char*)&clmHeader, sizeof(ClmHeader));
+		clmFileReader->Read(&clmHeader, sizeof(ClmHeader));
 		
 		try {
 			clmHeader.VerifyFileVersion();
@@ -34,7 +34,7 @@ namespace Archives
 		m_NumberOfPackedFiles = clmHeader.packedFilesCount;
 
 		indexEntries = std::vector<IndexEntry>(m_NumberOfPackedFiles);
-		clmFileReader->Read((char*)indexEntries.data(), m_NumberOfPackedFiles * sizeof(IndexEntry));
+		clmFileReader->Read(indexEntries.data(), m_NumberOfPackedFiles * sizeof(IndexEntry));
 	}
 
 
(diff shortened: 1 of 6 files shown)


DanRStevens:
I used global search and replace to help with this, so it should have caught all casts assuming no code formatting differences. I also verified there were no calls/casts in OP2Archive nor OP2Mapper.

This patch should close #51.

Brett208:
When attempting to compile this code, I get the following errors. I think it is because ClipRect is const and the Read function does not take a const parameter. I checked and putting the (char*) cast back in or using a (void*) cast fixes the errors.

Error (active) E0167 argument of type "const ClipRect *" is incompatible with parameter of type "void *" OP2Utility c:\ . . . \OP2Archive\OP2Utility\Maps\MapReader.cpp 54	

Error	 C2664 'void StreamReader::Read(void *,size_t)': cannot convert argument 1 from 'const ClipRect *' to 'void *'	OP2Utility c:\ . . . \op2archive\op2utility\maps\mapreader.cpp 54	

```
void MapReader::ReadClipRect(const ClipRect& clipRect)
{
	streamReader->Read(&clipRect, sizeof(clipRect));
}
```

DanRStevens:
Quite right, I seem to have missed that error from my compiler. Thanks for mentioning it. I guess I'm a little error blind right now since the code still doesn't compile 100% on Linux, so I'm used to always seeing errors of some kind. It's getting cleaner though.

The compiler is correct in flagging that as an error. The parameter should not be declared as `const` since the Read will necessarily write into it. This is actually unrelated to the Write template updates, though seemingly similar, where `const` is valid.

DanRStevens:
There was a merge conflict so I went ahead and rebased the changes to fix the conflict. I suppose a merge commit would be more proper to some people, though the history was also getting a bit messy.

On another note, the old C-style cast was hiding the const error by silently casting it away. A newer C++ style `static_cast` would have given a warning there, since it's not allowed to change const-ness.

Brett208:
Yeah, that const ClipRect& argument was definetly a bug I left for you. Good call on using static_cast in the future for improved warning/error detection. Not sure why I did that in the first place.

Compiles without errors on MSVC. Extracts all contents of a CLM file with OP2Archive without error. Reads a map into memory and saves a render to disk with OP2MapImager without error.

I thought that I had submitted a merge conflict fix earlier but I guess it didn't go through. Thanks for cleaning up the merge conflict.

Issue #43, opened 2018-06-05
Split Stream related classes into their own files, and organize into subfolder


DanRStevens:
The Stream classes are currently all lumped together in StreamReader/StreamWriter .h/.cpp files. They should be split into their own files for each class. The resulting files should be organized into a Streams folder.

Brett208:
I'd like to handle this task. It will make integrating with the Visual Studio project and filter file easier.

Streams sounds better for the folder name than my idea of Serialization.

If we are close to resolving the current open branches enhancing StreamReader/Writer, I'll wait for them to close as it will cause less merge conflicts. 

Thanks,
Brett

DanRStevens:
I think you're right, you may be better able to handle this one. I hadn't originally considered the Visual Studio project files.

I updated and pushed all my current changes, except for the template move change, which remains broken. It's trivial though, so we can just redo that after a refactor, assuming we find a way to fix it.

PR #53, opened 2018-06-08
Create streams directory

Brett208:
How to we want to handle including the Stream directory in other files?

I was thinking about leaving a StreamWriter.h and StreamReader.h header. These would each bring in 4 class StreamWriter, SeekableStreamWriter, FileStreamWriter, and MemoryStreamWriter.

To do this, I guess I need to create a StreamWriterBase.h and SeekableStreamWriter.h header. Otherwise I would have to place the FileStreamWriter and MemoryStreamWriter includes below the StreamWriter and SeekableStreamWriter definitions in StreamWriter.h. This seems wrong. Check the current commit on this branch to see this in action.

We could force the end user to bring in each individually. IE if they want to use FileStreamWriter, they would include FileStreamWriter.h. This seems like a lot of includes though because they probably are going to want the SeekableStreamWriter.h as well if not other stream classes.

Thanks,
Brett


OP2Utility.h
```diff
@@ -1,7 +1,7 @@
 #pragma once
 
 #include "StringHelper.h"
-#include "StreamReader.h"
+#include "Streams/StreamReader.h"
```
DanRStevens: What is the purpose of OP2Utility.h? Is this a generic import header to bring in everything the library has to offer, or does it have more limited project specific usage? If it's a generic import header, then it should probably import everything. If not, I assume it's correct as is.

Streams/MemoryStreamWriter.h
```diff
@@ -0,0 +1,37 @@
+#pragma once
+
+#include "StreamWriter.h"
```
DanRStevens: This should import SeekableStreamWriter, the direct parent. The SeekableStreamWriter class will need to import StreamWriter, so we get that header indirectly.

Streams/StreamWriter.cpp
```diff
@@ -0,0 +1,3 @@
+#include "StreamWriter.h"
+
+StreamWriter::~StreamWriter() { }
```
DanRStevens: StreamWriter is an interface, so we could reasonably omit a .cpp file. It's possible to allow for an inline empty destructor in the header file even though it's declared pure virtual. The pure virtual means that method doesn't require an implementation, and marks the class as abstract. It does not mean the function does not have an implementation. A rather subtle distinction.

Streams/StreamWriter.h
```diff
@@ -0,0 +1,29 @@
...
+	
+	// Seek to absolute position, given as offset from beginning of stream
+	virtual void Seek(uint64_t offset) = 0;
+
+	// Seek by a relative amount, given as offset from current position
+	virtual void SeekRelative(int64_t offset) = 0;
+};
+
+#include "FileStreamWriter.h"
+#include "MemoryStreamWriter.h"
```
DanRStevens: We should not be including the headers for derived objects. The inclusion should go the other direction.

Brett208: Just realized the GitHub flags when no newline is detected at the end of a file. Kind of neat feature.

Brett208:
Thanks for the thoughts on how to structure the headers.

It didn't occur to me that FileStreamWriter would automatically bring SeekableStreamWriter and StreamWriter into a project when referenced. So you could create a FileStreamWriter and reference it as a SeekableStreamWriter later in the file without adding a new include. 

For now I'll create individual headers for each stream class. If we feel like it adds too many headers into the rest of the files referencing streams, later we can add some sort of convenience header(s).

Brett208:
Should resolve issue #43 

Merge conflicts are too complex to resolve using the GitHub web editor. I'll resolve them locally when ready to merge.

I'm not sure how I feel about the complex file structure in the includes. For example: '#include "../Streams/MemoryStreamReader.h"'. On one hand it is nice to know exactly where the header is, but it also sort of clutters the includes.

Streams/StreamWriter.h
DanRStevens: Yes, handy indeed. Some editors also have settings to warn about missing newlines, or to automatically insert newlines at the end of files. I don't much like the automatic edits myself, but I do like the show whitespace setting so I can see when something is off. It felt a bit odd and ugly at first, but I quickly came to love it.

Brett208:
Sounds good on the includes. Lets leave as is for now then.

PR #54, opened 2018-06-09
Remove windows.h from ArchivePacker.h (And put it in ArchivePacker.cpp)

Brett208:
- Remove HANDLE m_OutFileHandle and associated helper functions from ArchivePacker
 - Use FileStreamWriter when creating a new Volume File
 - Transition internals of VolFile::CreateArchive to exception based error handling
 - Make CreateVolumeInfo:fileCount a const function


Brett208:
In order to remove windows.h from ArchivePacker.h, I had to do a lot of work in VolFile.cpp/.h.

I saw 2 places in VolFile.cpp where it would be nice to provide some better documentation (mostly because I don't know exactly what is going on myself...):

In VolFile::WriteFiles, annotate the following line:

```volWriter.Write(&temp, (-volInfo.indexEntry[i].fileSize) & 3);```

In `VolFile::WriteTag`, annotate the following line:

```buffer[1] = length | 0x80000000;```

Additional areas for future improvement:
 - Consider switching several of the c style arrays to std::arrays
 - Within 	CreateVolumeInfo, switch the following fields to std::vector:
    - IndexEntry *indexEntry;
    - int *fileNameLength;
    - HANDLE *fileHandle;

I think maybe these future improvements could be applied in a separate branch to keep this one from growing too much? I'm also okay diving in if that is what we prefer.

File: Archives/VolFile.h [hunk 1]
```diff
@@ -63,18 +64,19 @@ namespace Archives
...
 			{
 				return filesToPack.size();
 			}
 		};
 
 		int ReadTag(int offset, const char *tagText);
-		bool WriteTag(int length, const char *tagText);
-		bool CopyFileIntoVolume(HANDLE inputFile, int size);
+		void WriteTag(SeekableStreamWriter& volWriter, int length, const char *tagText);
+		void CopyFileIntoVolume(SeekableStreamWriter& volWriter, HANDLE inputFile, int size);
```
DanRStevens: These two methods might only need a StreamWriter, rather than a SeekableStreamWriter. They are both sequential in nature.

Archives/VolFile.h
```diff
@@ -63,18 +64,19 @@ namespace Archives
...
 		bool ReadVolHeader();
-		bool WriteFiles(CreateVolumeInfo &volInfo);
-		bool WriteHeader(CreateVolumeInfo &volInfo);
+		void WriteVolume(const std::string& fileName, const CreateVolumeInfo& volInfo);
+		void WriteFiles(SeekableStreamWriter& volWriter, const CreateVolumeInfo &volInfo);
+		void WriteHeader(SeekableStreamWriter& volWriter, const CreateVolumeInfo &volInfo);
```
DanRStevens: Ditto.

DanRStevens:
The WriteTag function can be explained better. It might even be possible to replace it by providing a Tag struct with a constructor, and feeding a temporary object (as a const lvalue-reference) right into the Write template method. It would still be a 1-liner at the point of call. The downside being, I don't think we'd get quite the nice exception message in case of an error.

In terms of the strange math, it's calculating how much padding bytes are needed to get the section lengths up to a multiple of 4 bytes.

The `& 3` is a bitmask used to quickly work modulo 4. In general, if **n** is a power of 2, you can work modulo **n** by **AND**ing the bits of a binary number with **n - 1**. The **AND** operation being faster than modulo.

If **x** is a multiple of **n**, then it is congruent to zero modulo **n**:
**x = 0 (mod n)**.

The negatives are to calculate additive inverses modulo **n**. If you add a number with its additive inverse, you get **0**. i.e. **1 + (-1) = 0**. Negatives are just the additive inverses for integers. Modulo arithmetic also has additive inverses: **x + (-x) = 0 (mod n)**. Though when working modulo **n**, we generally don't express values as negative, nor do we generally express values as being greater than **n**. For **0 <= x < n**, the additive inverse of **x** is: **-x = -x + n = n - x (mod n)**, where **0 <= n - x < n** as well. i.e. **3 + (4-3) = 3 + 1 = 4 = 0 (mod 4)**.

What that all goes to say is, **y = -x mod n** is such that **0 <= y < n**, and **x + y** is a multiple of **n**. That is, it's the smallest non-negative value that can be added to **x** to round it up to a multiple of **n**.

For **n = 4**:
**y = -x mod n = -x mod 4 = -x & (4-1) = -x & 3**

Archives/VolFile.cpp
```diff
@@ -226,85 +226,56 @@ namespace Archives
...
 		// Write the string table
-		if (WriteTag(volInfo.paddedStringTableLength, "vols") == false) {
-			return false;
-		}
+		WriteTag(volWriter, volInfo.paddedStringTableLength, "vols");
 
-		if (WriteFile(m_OutFileHandle, &volInfo.stringTableLength, 4, &numBytesWritten, nullptr) == 0) {
-			return false;
-		}
+		volWriter.Write(&volInfo.stringTableLength, sizeof(int));
```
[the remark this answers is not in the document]
Brett208: I'll fix this before merging.

Archives/VolFile.h
Brett208: Good point. I'll update this before merging.

Brett208:
I had to read the bitfield info 3 times to follow. I like the results though. What do you think about the comments below? I think they are enough to point someone in the right direction without over-commenting the code.

```
// Use a bitmask to increase tag size to a multiple of 4 bytes
buffer[1] = length | 0x80000000;

// Pad the end of the packed file to ensure the length is a multiple of 4 bytes
// Use a bitmask to quickly calculate the modulo 4 (remainder) of fileSize
volWriter.Write(&temp, (-volInfo.indexEntry[i].fileSize) & 3);
```

It would probably help readability to make the tags a struct and separate some of their specific needs out from the general read/write process. But that is a project for a different branch.

Thanks for the help.

Archives/VolFile.cpp
```diff
@@ -367,11 +370,12 @@ namespace Archives
...
 
 	// Writes a section tag to the open output file.
-	void VolFile::WriteTag(SeekableStreamWriter& volWriter, int length, const char *tagText)
+	void VolFile::WriteTag(StreamWriter& volWriter, int length, const char *tagText)
 	{
 		int buffer[2];
 
 		buffer[0] = *(int*)tagText;
+		// Use a bitmask to increase tag size to a multiple of 4 bytes
 		buffer[1] = length | 0x80000000;
```
DanRStevens: I've explained poorly and forgot to mention this part. This comment is wrong. Here a bit field is being used. The code is setting the highest most bit, which is used to indicate how the section should be padded. I believe it uses padding up to either a 2 or 4 byte boundary. We should refactor the code (in the future) to use structs with an embedded bit field. That would make this code much more clear, and make errors less likely.

For now, perhaps we can comment this as:
Set padding bit to indicate 4 byte padding

Archives/VolFile.cpp
```diff
@@ -241,16 +241,19 @@ namespace Archives
 
 			try {
 				CopyFileIntoVolume(volWriter, volInfo.fileHandle[i], volInfo.indexEntry[i].fileSize);
-				int temp = 0; // Pad with 0 bytes
-				volWriter.Write(&temp, (-volInfo.indexEntry[i].fileSize) & 3);
+				int padding = 0;
+				
+				// Pad the end of the packed file to ensure the length is a multiple of 4 bytes
+				// Use a bitmask to quickly calculate the modulo 4 (remainder) of fileSize
+				volWriter.Write(&padding, (-volInfo.indexEntry[i].fileSize) & 3);
```
DanRStevens: Slight correction. The padding is inter-file padding. It does not increase the recorded size of any file being packed. Rather, it adds space after the file. By doing this, a full 4 byte read starting within the file will always be valid, even if it extends beyond the end of the file. The later higher order bytes from such a read are guaranteed to be 0, which shouldn't produce any unexpected results.

The padding also means the next file will always start on a 4 byte boundary. If a binary file has packed 4 byte fields, which is common, especially on 32-bit systems, the 4 byte padding will ensure all fields will be properly aligned if mapped into memory, which allows for higher performance access.

Perhaps the comment should read:
Add padding after the file to ensure we end on a 4 byte boundary

DanRStevens:
This branch might be a good candidate for a squash and merge.

Brett208:
Thanks for the clarification on the bit operation comments. I felt like this branch was a major change so left it as a normal merge commit. 

I'll plan to remove windows.h from ArchivePacker.cpp next.

PR #55, opened 2018-06-10
Re-remove `char*` casts from templates

DanRStevens:
These seem to have been accidentally re-introduced during recent refactoring.


DanRStevens:
Knowing what a pain merge conflicts are with moved files, where everything is highlighted as changed, I knew to look for this :smirk:

PR #59, opened 2018-06-10
Make destructor virtual instead of pure virtual

DanRStevens:
Adding both pure virtual (= 0), and an explicit implementation ({}) to a
definition violates the C++ standard. Such code may be accepted by MSVC,
though causes an error in g++.

The point of a pure virtual method is that it allows the method
definition to be omitted, and marks the class as abstract, preventing
instances from being directly instantiated. Marking any method as pure
virtual is enough to mark a class as abstract.

A pure virtual method can still be optionally defined. It may be
accessed using the scope resolution operator. This might allow a derived
class to call a default implementation from the base class. Naturally
the base class method must be defined in such a case:
```
// Abstract base class
class Base {
public:
  virtual void someMethod() = 0; // Pure virtual
};

void Base::someMethod() {
  // Some default implementation
}

class Derived : public Base {
public:
  virtual void someMethod() override {
    Base::someMethod(); // Explicit call to base class implementation
  }
};
```

In the case of a destructor, the derived class destructor will have an
implicit call to the base class destructor. As such, `Base::~Base` must
be defined. Hence the implementation of a pure virtual destructor in a
base class is not really optional. Omitting the base class destructor
results in an unresolved external symbol error from the linker when
trying to link any derived class destructors.

By not marking a virtual base class destructor as pure, it will be
provided with a default (essentially empty) implementation, which will
satify calls from derived class destructors. As such, there is little
reason to mark a base class virtual destructor as pure, unless you need
to do so to make the class abstract. Since marking any method as pure
virtual makes the class abstract, you only need to select the destructor
if there is no other suitable method that can be marked as pure virtual.

----

The StreamReader and StreamWriter base classes are already abstract by
virtual of haing a pure virtual Read or Write method. As such, there is
no reason to also make the destructor pure virtual.

### DIFF
--- a/Streams/StreamReader.h
+++ b/Streams/StreamReader.h
@@ -4,6 +4,6 @@
 
 class StreamReader {
 public:
-	virtual ~StreamReader();
+	virtual ~StreamReader() = default;
 	virtual void Read(void* buffer, size_t size) = 0;
 };
--- a/Streams/StreamReader.h
+++ b/Streams/StreamReader.h
@@ -4,6 +4,6 @@
 
 class StreamReader {
 public:
-	virtual ~StreamReader() = 0 { };
+	virtual ~StreamReader();
 	virtual void Read(void* buffer, size_t size) = 0;
 };
--- a/Streams/StreamWriter.h
+++ b/Streams/StreamWriter.h
@@ -4,6 +4,6 @@
 
 class StreamWriter {
 public:
-	virtual ~StreamWriter();
+	virtual ~StreamWriter() = default;
 	virtual void Write(const void* buffer, size_t size) = 0;
 };
--- a/Streams/StreamWriter.h
+++ b/Streams/StreamWriter.h
@@ -4,6 +4,6 @@
 
 class StreamWriter {
 public:
-	virtual ~StreamWriter() = 0 { };
+	virtual ~StreamWriter();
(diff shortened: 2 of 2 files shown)


DanRStevens:
You can probably tell from the length of that commit message that I did a bit of research on that one. Forgot to give one of the relevant links: [Destructors](https://en.cppreference.com/w/cpp/language/destructor)

Brett208:
Using the default keyword on the StreamReader/Writer base destructors allowed the code to compile and run without errors or warnings. Rebased and Merged into the master branch.

@DanRStevens Upon attempting to delete, this branch is giving me the same warning about not be fully merged into the HEAD branch. I'll leave the remote and local copy of the branch intact in case you want to investigate.

Thanks,
Brett

DanRStevens:
Ok, interesting. I got a warning:

> warning: deleting branch 'virtual-dtor-fix' that has been merged to
         'refs/remotes/origin/virtual-dtor-fix', but not yet merged to HEAD.
Deleted branch virtual-dtor-fix (was 6f5f813).

Despite the warning, the delete was successful. I then deleted the branch on GitHub, and fetched changed with `git fetch --prune`, which cleaned up my remote tracking branch.

In the end, all copies of the branch were deleted, without me having to force delete anything.

PR #61, opened 2018-06-11
Remove windows.h from ArchivePacker.cpp

Brett208:
- Improve error handling and rollback of files on failure in ArchivePacker::ReplaceFileWithFile
 - Rename ArchiveUnpacker class variables as ArchiveX instead of VolumeX
 - Hook ClmFile::Repack into ArchivePacker::ReplaceFileWithFile

### DIFF
--- a/Archives/ArchivePacker.cpp
+++ b/Archives/ArchivePacker.cpp
@@ -1,7 +1,7 @@
 #include "ArchivePacker.h"
 #include "../XFile.h"
 #include "../StringHelper.h"
-#include <windows.h>
+#include <cstdio>
 #include <stdexcept>
 
 namespace Archives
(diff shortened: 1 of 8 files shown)


Archives/ArchivePacker.cpp
```diff
@@ -1,28 +1,38 @@
...
-	// newFile has it's name changed to that of fileToReplace and the old file is deleted.
-	// Returns true if successful and false otherwise
-	bool ArchivePacker::ReplaceFileWithFile(const char *fileToReplace, const char *newFile)
+	// newFile has its name changed to fileToReplace and the old file is deleted.
+	void ArchivePacker::ReplaceFileWithFile(const std::string& fileToReplace, const std::string& newFile)
 	{
-		std::string tempFileName("Temporary.vol");
-		MoveFileA(fileToReplace, tempFileName.c_str());
+		// If C++17 filesystem comes out of experimental status on both MSVC and Linux compilers, 
+		// consider switching to std::filesystem::rename & std::filesystem::remove
```
DanRStevens: This makes sense, I didn't realize there was a non-experimental/filesystem version of rename and remove.

Side note, I was initially confused by the identically named `std::remove` (and `std::remove_if`) from `<algorithm>`, though they have different signatures so there is no conflict.

Perhaps we should have wrapper functions for rename and remove in XFile, so as to minimize changes when we switch over. Or rather, since XFile already imports filesystem, we can just have it call the new methods right now.

In particular, the new [`std::experimental::rename`](http://en.cppreference.com/w/cpp/experimental/fs/rename) guarantees semantics such that we can avoid the intermediate temporary filename. That would simplify error checking considerably.

As for compatibility, XFile already imports filesystem, which the MinGW cross compiler fails on. Hence I haven't been testing or reporting issues on that one file. Just now though, I've installed a newer version of Clang, which can handle `#include <experimental/filesystem>`. Naturally though, it fails on `#include <windows.h>`, so again I have yet to find a way to build the entire project from Linux.

I would recommend creating wrappers in XFile, and calling the new experimental/filesystem code from there. We might only need `rename` for this feature. This can be done in a future branch.

Edit: I'm just realizing the scope here. The `rename` method from the new filesystem header can completely replace the current function `ReplaceFileWithFile`. We should dump this current function, and replace calls with an XFile wrapper that makes use of the new `filesystem::rename`.

Archives/ClmFile.cpp
```diff
@@ -157,7 +157,19 @@ namespace Archives
...
+		}
+
+		// Rename the output file to the desired file
+		try {
+			ReplaceFileWithFile(m_ArchiveFileName, tempFileName);
+			return true;
+		}
+		catch (std::exception& e) {
+			return false;
+		}
```
DanRStevens: Huh, I'm surprised by this change. That seems like quite the glaring inconsistency between the Clm code and the Vol code. It looks like a big bug on my part, and I'm wondering now if I ever tested the code.

Well, a valiant effort then to correct shoddy code, complicated by the messy return codes.

Brett208:
I went ahead and created a wrapper for Filesystem::rename within XFile. Inside XFile.cpp, we should consider changing the using statement for experimental/filesystem into a rename of the namespace to fs or something else. This way we could create an XFile::rename function or other same name functions while avoiding naming conflicts. I didn't want to make this kind of sweeping change on this branch though. 

RenameFile isn't the greatest name choice, since it can also rename other objects besides a file. I'm happy with it for now though.

See the commit message for some more details.

When (or if) filesystem comes out of experimental on both MSVC and whichever primary Linux compiler we are using, I'd like to consider eliminating XFile from the project. the C++17 filesystem namespace is basically what we want so there will be no reason to homeroll our own cross-platform filesystem support. There may be a couple functions that we want to keep, but it would become more of a FilesystemHelper namespace to supplement the std library than what it is right now. 

Since I'm not sure that I want to keep XFile in the long run, I haven't been putting good time into refactoring it, which it could use.

Thanks,
Brett

DanRStevens:
Lets hold off on making changes to the using statements in XFile for now. Especially if we might change or eliminate it later. In particular, when filesystem is accepted into the mainstream, renaming its namespace to something like fs would be more awkward in the future.

PR #63, opened 2018-06-14
Conform Archives size_t calls to Clang Compiler Requirements

Brett208:
Also works with MSVC without warnings

 - Add std namespace prefix to all size_t calls
 - Ensure all files containing size_t include a header that defines size_t
 - Add brackets to all single line if and for statements in VolFile.cpp
 - Use auto keyword where appropriate in loops


Brett208:
Thanks for fixing the size_t arguments in the Streams and MapReader/Writer code. I had just done the Archive code as a test.

PR #64, opened 2018-06-15
Update VolFile::CreateVolumeInfo Struct to use STL Containers

Brett208:
- Replace pointers to c-style arrays with std::vector
 - Remove member container int* fileNameLength
 - Update container names to be plural
 - Remove some no longer required memory management
 - Refactor VolFile::PrepareHeader

### DIFF
--- a/Archives/VolFile.cpp
+++ b/Archives/VolFile.cpp
@@ -229,15 +229,9 @@ namespace Archives
 	void VolFile::CleanUpVolumeCreate(CreateVolumeInfo &volInfo)
 	{
 		// Close all input files
-		for (std::size_t i = 0; i < volInfo.fileCount(); i++)
-		{
-			CloseHandle(volInfo.fileHandle[i]);
+		for (auto& fileHandle : volInfo.fileHandles) {
+			CloseHandle(fileHandle);
 		}
-
-		// Release memory used to create the index
-		delete[] volInfo.indexEntry;
-		delete[] volInfo.fileHandle;
-		delete[] volInfo.fileNameLength;
 	}
 
 	void VolFile::WriteFiles(StreamWriter& volWriter, const CreateVolumeInfo &volInfo)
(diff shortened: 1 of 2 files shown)


Brett208:
There is some missing error handling if something goes wrong with the STL containers. I'd like to save addressing this until HANDLE is replaced with one of our stream classes.

It would be nice to refactor VolFile::PrepareHeader into smaller functions. Quite a lot is happening in there right now. 

To ease readability, I removed the tempPointer in the code below. @DanRStevens does this cost us any efficiency?
```
// Calculate offsets to the files
for (std::size_t i = 1; i < volInfo.fileCount(); i++)
{
	IndexEntry previousIndex = volInfo.indexEntries[i - 1];
	volInfo.indexEntries[i].dataBlockOffset = (previousIndex.dataBlockOffset + previousIndex.fileSize + 11) & ~3;
}

Code compiles without warnings or errors. A vol file is created with the code that properly interfaces with Outpost2.exe.

Next step for me will be more work in VolFile::CreateVolumeInfo. Updating to fixed width variables and replacing the HANDLE with a stream.

File: Archives/VolFile.cpp [hunk 1]
```diff
@@ -284,42 +279,36 @@ namespace Archives
...
 				FILE_ATTRIBUTE_NORMAL,	// file attributes
 				nullptr);				// template
+			
+			volInfo.fileHandles.push_back(fileHandle);
 
-			if (volInfo.fileHandle[i] == INVALID_HANDLE_VALUE)// Check for errors
+			if (fileHandle == INVALID_HANDLE_VALUE)// Check for errors
 			{
 				// Error opening file. Close already opened files and return error
 				for (i--; i >= 0; i--) {
```
DanRStevens: Unrelated to your changes here, but this is an infinite loop for unsigned types. Perhaps the type of `i` was changed to unsigned, or perhaps the bug always existed. We should restructure this code if continued changes don't manage to eliminate it.

Archives/VolFile.cpp
```diff
@@ -329,36 +318,37 @@ namespace Archives
...
 		volInfo.paddedIndexTableLength = (volInfo.indexTableLength + 3) & ~3;
-		volInfo.indexEntry[0].dataBlockOffset = volInfo.paddedStringTableLength + volInfo.paddedIndexTableLength + 32;
+		volInfo.indexEntries[0].dataBlockOffset = volInfo.paddedStringTableLength + volInfo.paddedIndexTableLength + 32;
 		
 		// Calculate offsets to the files
 		for (std::size_t i = 1; i < volInfo.fileCount(); i++)
 		{
-			tempPtr = &volInfo.indexEntry[i - 1];
-			volInfo.indexEntry[i].dataBlockOffset = (tempPtr->dataBlockOffset + tempPtr->fileSize + 11) & ~3;
+			IndexEntry previousIndex = volInfo.indexEntries[i - 1];
```
DanRStevens: I think we should use a reference type here:
```
IndexEntry& previousIndex = volInfo.indexEntries[i - 1];
```
A non-reference type will call the copy constructor and create a duplicate of the object. That is less efficient. Using the reference type does improve the syntax, while retaining the efficiency of using a pointer.

Might even make it a `const` reference, since it's just being used to read.

Brett208:
Thanks for the review. I pushed the reference fix when calculating file offsets.

I thought the for loop looked suspect, but didn't mess with it because as you mentioned, it will not exist when we change the HANDLE out with a stream.

Thanks,
Brett

PR #65, opened 2018-06-17
Remove MemoryMappedFile & switch to fixed width types in VolFile

Brett208:
- Start using fixed width types in VolFile code to improve Read/Write stream process
 - Remove MemoryMappedFile class to remove VolFile's dependency on Windows.h
 - Refactor VolFile::ReadVolHeader (Still needs more cleaning)


Brett208:
This branch started as changing variables in VolFile as fixed width and unsigned where appropriate. This messed with some of the bit operations in VolFile which I had difficulty restoring. To solve the problem, I ended up just removing MemoryMappedFile and eliminating several sections that were giving me trouble. This seemed to make more sense than struggling through the bit operation changes only to delete them a few commits later. Unfortunately, the side effect is a large change-set.

It would be helpful to continue cleaning the function VolFile::ReadVolHeader in the future.

We may want to revisit the function VolFile::ReadStringTable as there is probably a cleaner way to accomplish it. The function currently works though.

Getting close to eliminating windows.h entirely...

Brett208:
I've ran into some bugs when opening certain volumes. One bug that I think I've fixed is not seeking past the padding if present at the end of the StringTable. There seems to be another bug associated with locating the beginning of a file for unpacking. These did not manifest on my test cases, probably because I was just lucky and the padding didn't exist.

I've also locally improved the error messaging when reading from the volume header.

Archives/VolFile.cpp
```diff
@@ -389,103 +382,102 @@ namespace Archives
...
-
-		// Check for fileName string table tag ("vols" tag)
-		// ------------------------------------------------
-		if ((m_StringTableLength = ReadTag(16, "vols")) == -1) {
-			return false;
-		}
-
-		m_ActualStringTableLength = *((int*)m_BaseOfFile + 6);	// Store actual length
-		m_StringTable = (char*)m_BaseOfFile + 28;				// Store pointer to table
+		ReadStringTable();
```
DanRStevens: The string table read should probably happen after the checks on header sizes. Those checks are largely to prevent problems from reading past the end of the file.

Archives/VolFile.cpp
```diff
@@ -389,103 +382,102 @@ namespace Archives
...
 
 		// Make sure the string table fits in the header (and next header fits too)
 		if (m_HeaderLength < m_StringTableLength + 20) {
```
DanRStevens: Note: this is `sizeof(stringTableTagStruct) + sizeof(stringTableLength) + sizeof(indexTableTagStruct)`

Archives/VolFile.cpp
```diff
@@ -389,103 +382,102 @@ namespace Archives
...
-		// ---------------------------------------
-		// Store the offset to the Index table while it's still calculated
-		m_Index = (IndexEntry*)((char*)m_BaseOfFile + temp + 8);
-		if ((m_IndexTableLength = ReadTag(temp, "voli")) == -1) {
-			return false;
+		m_IndexTableLength = ReadTag(std::array<char, 4> { 'v', 'o', 'l', 'i' });
+		if (m_IndexTableLength > 0) {
+			m_IndexEntries.resize(m_IndexTableLength / sizeof(IndexEntry));
+			archiveFileReader->Read(m_IndexEntries.data(), m_IndexTableLength);
 		}
```
DanRStevens: This does not account for padding bytes. There is no guarantee the size of the index section is an integral multiple of `sizeof(IndexEntry)`, and in some of the stock VOL files, it isn't. It's possible additional bytes will need to be skipped over here.

Archives/VolFile.cpp
```diff
@@ -389,103 +382,102 @@ namespace Archives
...
+	void VolFile::ReadStringTable()
+	{
+		uint32_t actualStringTableLength;
+		archiveFileReader->Read(&actualStringTableLength, sizeof(actualStringTableLength));
+
+		std::string charBuffer;
+		charBuffer.resize(actualStringTableLength);
+		archiveFileReader->Read(&charBuffer[0], actualStringTableLength);
+
+		m_StringTable.push_back("");
```
DanRStevens: Why?

Archives/VolFile.cpp
```diff
@@ -389,103 +382,102 @@ namespace Archives
...
+			if (charBuffer[i] == '\0') {
+				m_StringTable.push_back("");
+				continue;
+			}
+
+			m_StringTable[m_StringTable.size() - 1].push_back(charBuffer[i]);
+		}
+
+		m_StringTable.erase(m_StringTable.begin() + m_StringTable.size() - 1);
+	}
```
DanRStevens: I'm rather confused by this function. In particular, if a string starts with `'\0'`, it is already `""`, so I don't think there is a need for special casing.

I'm uncertain why `erase` is being used at the end.

Archives/VolFile.cpp
```diff
@@ -389,103 +382,102 @@ namespace Archives
...
 			// Make sure entry is valid
-			if (m_Index[temp].fileNameOffset == UINT_MAX) {
+			if (m_IndexEntries[packedFileCount].fileNameOffset == UINT_MAX) {
 				break;
 			}
 		}
-		m_NumberOfPackedFiles = temp;		// Store the number of used index entries
-
-		return true;						// Header read successfully
+		m_NumberOfPackedFiles = packedFileCount;
```
DanRStevens: Consider using `m_NumberOfPackedFiles` directly in the loop and eliminating the local variable.

Alternatively, the loop could start with `for (; packedFileCount < m_numberOfIndexEntries; packedFileCount++)`.

Archives/ArchiveUnpacker.h
```diff
@@ -14,7 +15,7 @@ namespace Archives
 		virtual ~ArchiveUnpacker();
 
 		std::string GetVolumeFileName() { return m_ArchiveFileName; };
-		int GetVolumeFileSize() { return m_ArchiveFileSize; };
+		uint32_t GetVolumeFileSize() { return m_ArchiveFileSize; };
```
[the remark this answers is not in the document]
Brett208: Will update to read uin64_t in current open branch.

Archives/VolFile.cpp
```diff
@@ -6,33 +6,29 @@
...
 
-
 namespace Archives
 {
 	VolFile::VolFile(const char *fileName) : ArchiveFile(fileName)
 	{
-		if (MemoryMapFile(fileName)) {
-			throw std::runtime_error("Could not open vol file " + std::string(fileName) + ".");
-		}
+		archiveFileReader = std::make_unique<FileStreamReader>(fileName);
```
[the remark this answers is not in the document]
Brett208: I was just learning about constructor initializer lists when making the FileSliceReader, so good timing about posting this. I'll apply this suggestion in VolFile, ArchiveUnpacker, and ClmFile in the current open branch. The only reason they were wrapped in the unique pointer is because I couldn't figure out how to initialize them properly in the constructor earlier.

Archives/VolFile.cpp
```diff
@@ -6,33 +6,29 @@
...
 
-		m_ArchiveFileSize = m_MappedFileSize;
+		// Archive file size is limited to unsigned 4 bytes in length
+		m_ArchiveFileSize = static_cast<uint32_t>(archiveFileReader->Length());
```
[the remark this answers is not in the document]
Brett208: I'll fix this in the current open branch. I also noticed that ClmFile isn't even setting m_ArchiveFileSize right now so I'll update that as well.

Archives/VolFile.cpp
```diff
@@ -89,7 +85,7 @@ namespace Archives
 	{
 		CheckPackedFileIndexBounds(fileIndex);
 
-		char* offset = (char*)m_BaseOfFile + m_Index[fileIndex].dataBlockOffset;
+		char* offset = (char*)m_IndexEntries[fileIndex].dataBlockOffset;
```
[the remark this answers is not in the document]
Brett208: This is addressed in a different branch. Thanks for pointing out though. A pretty big blunder on my part to get into the mess in the first place.

Archives/VolFile.cpp
```diff
@@ -354,7 +347,7 @@ namespace Archives
 		return true;
 	}
 
-	void VolFile::CopyFileIntoVolume(StreamWriter& volWriter, HANDLE inputFile, int size)
+	void VolFile::CopyFileIntoVolume(StreamWriter& volWriter, HANDLE inputFile, int32_t size)
```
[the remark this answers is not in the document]
Brett208: I had planned to delete the size argument when refactoring, but it got away from me. I'll delete it in the current branch.

Archives/VolFile.cpp
```diff
@@ -389,103 +382,102 @@ namespace Archives
...
-	// invalid then this function returns -1.
-	// Note: tagText is a 4 byte field
-	int VolFile::ReadTag(int offset, const char *tagText)
+	// invalid then an error is thrown.
+	uint32_t VolFile::ReadTag(const std::array<char, 4>& tagName)
 	{
-		char *addr = (char*)m_BaseOfFile + offset;
+		// Tags are 4 chars in length and do not include a null terminator
+		std::array<char, 4> tagFromFile;
+		archiveFileReader->Read(tagFromFile.data(), tagFromFile.size());
```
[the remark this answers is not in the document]
Brett208: Removing the .data function call allows the function to work without specifying a specific size. Will be updated in the current branch. I was wondering how to get that to work. In the discussion about new template functions, we can forget about the std::array version and just focus on vector and string.

Thanks.
Brett

Archives/VolFile.cpp
```diff
@@ -389,103 +382,102 @@ namespace Archives
...
 
-		int length = *(int*)(addr + 4);			   // Get the length (and length tag)
-		if ((length & 0x80000000) != 0x80000000) { // Check for the tag (MSB set)
-			return -1;
+		uint32_t length = 0;
+		archiveFileReader->Read(&length, sizeof(length));
+
+		// Check for the tag (MSB set)
+		if ((length & 0x80000000) != 0x80000000) {
+			throw std::runtime_error("Tag " + std::string(tagName.data()) + " contains an invalid length.");
```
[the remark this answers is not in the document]
Brett208: Okay, so I'm looking at the bitwise operation here. 

0x8000 0000 is equivalent to b1000 0000 0000 0000 0000 0000 0000 0000

& will mask out all values not set, so we are checking if the highest order bit is set in a 32 bit (4 byte) field is set to 0 or 1. In this case, if the highest order bit is set to 1, it indicates 2 byte padding is used. While 2 byte padding is mentioned in the volume documentation, it is not actually available for use?

Thanks,
Brett‬

Archives/VolFile.cpp
Brett208: Will update to read sizeof(SectionHeader) + sizeof(m_StringTableLength) + sizeof(SectionHeader)

Archives/VolFile.cpp
```diff
@@ -389,103 +382,102 @@ namespace Archives
...
-			return false;
+		uint32_t volhSize = ReadTag(std::array<char, 4> { 'v', 'o', 'l', 'h' });
+		if (volhSize != 0) {
+			return false; //the size of this section must be 0
 		}
 
-		// Make sure the file is large enough to contain the header
-		if (m_MappedFileSize < m_HeaderLength + 8) {
-			return false;
-		}
```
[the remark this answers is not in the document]
Brett208: Check will be restored. I'll label 8 as sizeof(HeaderSection) * 2.

Archives/VolFile.cpp
Brett208: The code is adding an empty string whenever a null terminator is encounter. Then it fills the string with characters until the next null terminator is found. 

The final null terminator will cause an empty string to be added to the vector, which must be deleted when the for loop is completed.

There is probably a better way to accomplish this in terms of efficiency and readbility, although the provided code doesn't seem to have any bugs that I've seen.

Archives/VolFile.cpp
Brett208: I think you are referinng to the empty push_back call. There needs to be a string present to add characters to for when the loop starts.

Archives/VolFile.cpp
Brett208: What do you think of eliminating m_NumberOfPackedFiles altogether and just using m_numberOfIndexEntries to represent both? I'm guessing they are separate values to support deleting a file from a volume without recreating the volume? However, in practice, I don't think this is ever done, so maybe not a valid use case?

Thanks,
Brett

Archives/VolFile.cpp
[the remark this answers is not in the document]
DanRStevens: The existing VOL files never used 2-byte padding, and my writing code always output 4-byte padding, so I never bothered to support 2-byte padding. I only know about it because the game has code paths to support 2-byte padding.

The length is stored in the lower 31 bits, so technically it's not the length that's invalid. It's just the padding bit setting that is unsupported.

Archives/VolFile.cpp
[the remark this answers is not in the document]
DanRStevens: Err, `sizeof(HeaderSection) == 8`. It's just a single header section.

Archives/VolFile.cpp
DanRStevens: Oh, I get it now. That is a bit obscure. And I had misread the `+` as a `,` in the `erase` call, making it look like it was erasing the whole table just after it had been constructed.

Maybe we can flag this for review at a later point once we've got other higher priority items out of the way.

I think one way to make the code easier to read, is using the null terminated string constructor for `std::string`, rather than scanning and appending characters one at a time. The address of the next string can then be calculated using the length of the previous string plus the null terminator. Then you don't need to append empty placeholder string objects.

Archives/VolFile.cpp
DanRStevens: You may have a point here. We don't repack in-place.

The point of having that value separate is to track empty index entries. The empty entries could be used to add new files without needing to rewrite and shift the contents to make room for extra index entries. The string table also has a similar pair of values for used versus allocated size.

The other side of this is, you can delete files from a VOL and not need to repack the whole thing. Just mark the index as unused. Even removing the entry from the string table and the file contents would be optional, though probably desired. Bonus, if the packed file was at the end of the VOL, it would be easy to truncate the file in constant time, with no copying of data to fill in gaps to save space.

I guess the question is, do we want to ever support repacking in place? My thought is probably not, since it's kind of messy, and supporting it means having extra wasted allocated space.

Brett208: I think the answer is no repacking in place by either removing or adding a file. If I'm deleting a file, I want to reduce the disk size of the volume, which probably means repacking. If I'm adding a file, it should be added so as to support the binary search for the filename, which is probably easiest to support by repacking. 

The other reason I'm thinking of not supporting this in maintainability. This feature is fully untested and I'd rather not bother with trying out and supporting the different use cases.

For now, I'll just open an issue on this one so we can merge the branch.

-Brett

DanRStevens: Good call. I'm eager to see these new changes merged in.

PR #67, opened 2018-06-18
Fix Bugs When Reading and extracting from volume files

Brett208:
A bug remains when attempting to extract a file or open a memory stream to a file's location in a volume.

 - Account for difference in padding and actual size of volume header StringTable
 - Improve error messaging when reading a volume header into memory
 - Remove name of exception when an exception is caught but the exception is not referenced in Archive code.

### DIFF
--- a/Archives/VolFile.cpp
+++ b/Archives/VolFile.cpp
@@ -15,9 +15,7 @@ namespace Archives
 		// Archive file size is limited to unsigned 4 bytes in length
 		m_ArchiveFileSize = static_cast<uint32_t>(archiveFileReader->Length());
 
-		if (!ReadVolHeader()) {
-			throw std::runtime_error("Invalid vol header in " + std::string(fileName) + ".");
-		}
+		ReadVolHeader();
 	}
 
 	VolFile::~VolFile() { }
@@ -115,11 +113,11 @@ namespace Archives
 		try
 		{
 			char* offset = (char*)m_IndexEntries[fileIndex].dataBlockOffset;
-			int length = *(int*)(offset + 4) & 0x7FFFFFFF;
+			std::size_t length = *(int*)(offset + 4) & 0x7FFFFFFF;
 			offset += 8;
 			
 			FileStreamWriter fileStreamWriter(pathOut);
-			fileStreamWriter.Write(offset, length);
+			fileStreamWriter.Write(&offset, length);
 		}
 		catch (std::exception e)
 		{
@@ -174,7 +172,7 @@ namespace Archives
 			XFile::RenameFile(tempFileName, m_ArchiveFileName);
 			return true;
 		}
-		catch (std::exception& e) {
+		catch (std::exception&) {
 			return false;
 		}
 	}
@@ -201,7 +199,7 @@ namespace Archives
 		try {
 			WriteVolume(volumeFileName, volInfo);
 		}
-		catch (std::exception& e) {
+		catch (std::exception&) {
 			CleanUpVolumeCreate(volInfo);
 			return false;
 		}
@@ -392,7 +390,8 @@ namespace Archives
 		archiveFileReader->Read(tagFromFile.data(), tagFromFile.size());
 
 		if (tagFromFile != tagName) {
-			throw std::runtime_error("Tag " + std::string(tagFromFile.data(), tagFromFile.size()) + " does not match requested tag of " + std::string(tagName.data(), tagName.size()));
+			throw std::runtime_error("The tag " + std::string(tagName.data(), tagName.size()) + 
+				" was not found in the proper position in volume " + m_ArchiveFileName);
 		}
 
 		uint32_t length = 0;
@@ -400,7 +399,7 @@ namespace Archives
 
 		// Check for the tag (MSB set)
 		if ((length & 0x80000000) != 0x80000000) {
-			throw std::runtime_error("Tag " + std::string(tagName.data()) + " contains an invalid length.");
+			throw std::runtime_error("The tag " + std::string(tagName.data(), tagName.size()) + " from volume " + m_ArchiveFileName + " contains an invalid length.");
 		}
 
 		// Mask out the tag and return the length.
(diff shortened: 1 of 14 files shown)


Archives/VolFile.cpp
```diff
@@ -115,11 +113,11 @@ namespace Archives
 		try
 		{
 			char* offset = (char*)m_IndexEntries[fileIndex].dataBlockOffset;
-			int length = *(int*)(offset + 4) & 0x7FFFFFFF;
+			std::size_t length = *(int*)(offset + 4) & 0x7FFFFFFF;
```
DanRStevens: File access should use fixed sized fields.

The original code was de-referencing a pointer into the mapped view of the file. Hence it used to be `base + offset`, where base was the base address of the window into the memory mapped file, and offset was the offset into the file.

This data needs to be actually read into memory now.

Archives/VolFile.cpp
```diff
@@ -465,6 +461,9 @@ namespace Archives
 		}
 
 		m_StringTable.erase(m_StringTable.begin() + m_StringTable.size() - 1);
+
+		// Seek to the end of padding at end of StringTable
+		archiveFileReader->SeekRelative(m_StringTableLength - actualStringTableLength - 4);
```
DanRStevens: Ahh, right. I missed that.

DanRStevens:
This change does fix the one stated bug, though other more serious bugs remain.

It may not be easy or quick to get master back into a working state. I'm not sure we have good infrastructure in place to support replacing the memory mapped I/O yet. In particular, being able to take a slice of another file stream would help. A possibly easier short term solution might be a memory stream that owns it's own buffer, which is cleaned up in its destructor.

Archives/VolFile.cpp
```diff
@@ -114,12 +128,14 @@ namespace Archives
...
 		{
-			char* offset = (char*)m_IndexEntries[fileIndex].dataBlockOffset;
-			int length = *(int*)(offset + 4) & 0x7FFFFFFF;
-			offset += 8;
+			SectionHeader sectionHeader = GetSectionHeader(fileIndex);
 			
+			std::vector<void*> packedFileBuffer;
+			packedFileBuffer.resize(sectionHeader.length);
+			archiveFileReader->Read(packedFileBuffer.data(), sectionHeader.length);
+
```
Brett208: This should be refactored to read files in a standard chunk size instead of whole files to improve performance on very large files. Maybe use the constant VOL_WRITE_SIZE?

Brett208:
Spent a lot of time today getting this all working. Ran out of time refactoring, so there are some rough edges I'm sure. 

The FileSliceReader successfully read the map Ashes.map from the maps volume without actually extracting it to disk and allowed OP2MapImager to image it. Additionally, I was able to extract the map to disk after this, so both use cases are working.

I think the last known bug to me is dealing with padding if it exists at the end of IndexEntries on volume file. Open for suggestions on how to handle it.

Probably next step is just to go through the comments that @DanRStevens  already made and address/incorporate them. Then go through and make another pass on the new code. 

Thanks,
Brett

Archives/VolFile.cpp
DanRStevens: You're allocating 4x-8x more memory than you probably think you are ;)

Agreed on using a chunked write. It's a fairly generic operation to copy from one stream to another. Perhaps this can be abstracted to a utility function. For now though, this should be sufficient to fix the current issues. Optimization can always be done later.

The `void*` should of course be changed to a `uint8_t` or `char`-type.

Brett208:
Sounds good, I'll work at least one more update to this branch today before merging and keep addressing comments. Ran out of time right now though.

-Brett

Streams/FileSliceReader.h
```diff
@@ -1,24 +1,31 @@
...
+
 protected:
 	void ReadImplementation(void* buffer, std::size_t size) override;
 
 private:
 	FileStreamReader fileStreamReader;
-	uint64_t startingOffset;
-	uint64_t sliceLength;
+	const uint64_t startingOffset;
+	const uint64_t sliceLength;
```
DanRStevens: Very interesting, and probably a good idea.

Streams/FileStreamWriter.h
```diff
@@ -17,9 +18,14 @@ class FileStreamWriter : public SeekableStreamWriter
 	void Seek(uint64_t offset) override;
 	void SeekRelative(int64_t offset) override;
 
+	inline std::string GetFilename() const {
```
DanRStevens: Maybe this should return a const reference. It's likely to be used in a read-only manner, so that would avoiding copying to a new `std::string` object, which this currently does.

Archives/VolFile.cpp
```diff
@@ -132,7 +129,7 @@ namespace Archives
 			
 			std::vector<void*> packedFileBuffer;
```
DanRStevens: This data type still needs fixing. Should by a byte sized data type. Pointers are not byte sized. When the buffer is resized, it allocates `sectionHeader.length * sizeof(void*)` bytes. For a 32-bit build, pointers are 4 bytes, so that is 4x more memory than needed, and for a 64-bit build, that is 8x more memory than needed. Plus the type just doesn't make sense in this context.

The vector should contain a hidden internal pointer to an array of bytes, not a pointer to an array of pointers to unknown type. You have too much indirection here. :wink:

Archives/VolFile.cpp
```diff
@@ -430,25 +429,30 @@ namespace Archives
 	{
 		m_HeaderLength = ReadTag(std::array<char, 4> { 'V', 'O', 'L', ' ' });
 
+		// Make sure the file is large enough to contain the header
+		if (archiveFileReader.Length() < m_HeaderLength + sizeof(SectionHeader) * 2) {
```
DanRStevens: This check was originally meant to verify minimum size was at least 1 `SectionHeader`.

Streams/FileStreamWriter.h
Brett208: Agreed. I hadn't thought of this.

Archives/VolFile.cpp
Brett208: Thanks for re-pointing this out. I was going to research what size of void* was after you last pointer. I'll plan to update to uint8_t as this makes more since to me than char since it isn't supposed to be text.

Archives/VolFile.cpp
Brett208: Good point. I'll fix this bug and read the check to ensure the volume is big enough for the initial 'VOL ' header as well.

Archives/VolFile.cpp
DanRStevens: Not sure if it was intended, but I appreciate your puns. :smiley:

Brett208:
I went ahead and merged this to get the master back in working order. I think the current list of untracked bugs/features are:

1. Whether to allow deleting/adding files to an existing archive & cleaning up VolFile::ReadPackedFileCount()
2. Extracting files from volumes in chunks instead of entire file at once. Possible creation of standardized function for this either in ArchiveUnpacker or maybe attached to FileStreamReader.
3. Handling possible extra padding on the end of volume IndexEntries

After these are pushed into the issue tracker, I'd like to either whittle down some of the easier issues or proceed with removing the Windows HANDLE from the CreateVolumeInfo struct.

Thanks,
Brett

DanRStevens:
Awesome that this is done. I think that was the hardest part of the conversion process. Now we just have to handle reading of input files with standard's based methods, and we can get rid of the `<windows.h>` dependency. We're actually really close now.

I understand if you want to do some easy stuff after getting that memory mapped I/O part out of the way.

Issue #58, opened 2018-06-10
Represent VOL file tag names as constants


Brett208:
Besides just being a loose set of constants, they could be organized as an enum to reduce clutter in local scope. Would need to calculate the uint32_t value of each tag name.

```
#include <cstdint>

const uint32_t VOL_ = 0xX; // "VOL "
const uint32_t VOLH = 0xX; // "volh"
const uint32_t VOLS = 0xX; // "vols"
const uint32_t VOLI = 0xX; // "voli" 
const uint32_t VBLK = 0xX; // "VBLK"
```

Brett208:
I would like to see the definition of each tag consolidated to reduce repetition in the codebase. 

I agree using some form of std::array<char, 4> would be more intuitive than uint32_t. I was sort of following the pattern for the WAV constants in CLM files. This pattern confused me in the beginning so probably better to try some sort of char array.

However, this is a pretty minor issue and I'd be okay just burying it for now while working the bigger issues like getting windows.h out of the VolFile class.

PR #70, opened 2018-06-21
Improve section tag constants in archives

Brett208:
See commit messages for details. This should close out issue #58. 

Tested by listing contents of a volume, and packing a new clm file.

### DIFF
--- a/Archives/VolFile.cpp
+++ b/Archives/VolFile.cpp
@@ -19,8 +19,6 @@ namespace Archives
 
 	VolFile::VolFile(const char *fileName) : ArchiveFile(fileName), archiveFileReader(fileName)
 	{
-		static_assert(sizeof(SectionHeader) == 8, "SectionHeader not of required size");
-
 		m_ArchiveFileSize = archiveFileReader.Length();
 
 		ReadVolHeader();
--- a/Archives/VolFile.cpp
+++ b/Archives/VolFile.cpp
@@ -19,6 +19,8 @@ namespace Archives
 
 	VolFile::VolFile(const char *fileName) : ArchiveFile(fileName), archiveFileReader(fileName)
 	{
+		static_assert(sizeof(SectionHeader) == 8, "SectionHeader not of required size");
+
 		m_ArchiveFileSize = archiveFileReader.Length();
 
 		ReadVolHeader();
@@ -253,7 +255,7 @@ namespace Archives
 		// Write each file header and contents
 		for (std::size_t i = 0; i < volInfo.fileCount(); i++)
 		{
-			WriteTag(volWriter, volInfo.indexEntries[i].fileSize, TagVBLK);
+			volWriter.Write(SectionHeader(TagVBLK, volInfo.indexEntries[i].fileSize));
 
 			try {
 				CopyFileIntoVolume(volWriter, volInfo.fileHandles[i]);
@@ -272,13 +274,12 @@ namespace Archives
 	void VolFile::WriteHeader(StreamWriter& volWriter, const CreateVolumeInfo &volInfo)
 	{
 		// Write the header
-		WriteTag(volWriter, volInfo.paddedStringTableLength + volInfo.paddedIndexTableLength + 24, TagVOL_);
+		volWriter.Write(SectionHeader(TagVOL_, volInfo.paddedStringTableLength + volInfo.paddedIndexTableLength + 24));
 
-		WriteTag(volWriter, 0, TagVOLH);
-		//WriteTag(volWriter, 0, "volh");
+		volWriter.Write(SectionHeader(TagVOLH, 0));
 
 		// Write the string table
-		WriteTag(volWriter, volInfo.paddedStringTableLength, TagVOLS);
+		volWriter.Write(SectionHeader(TagVOLS, volInfo.paddedStringTableLength));
 
 		volWriter.Write(&volInfo.stringTableLength, sizeof(volInfo.stringTableLength));
 
@@ -292,7 +293,7 @@ namespace Archives
 		volWriter.Write(&padding, volInfo.paddedStringTableLength - (volInfo.stringTableLength + 4));
 
 		// Write the index table
-		WriteTag(volWriter, volInfo.indexTableLength, TagVOLI);
+		volWriter.Write(SectionHeader(TagVOLI, volInfo.indexTableLength));
 
 		volWriter.Write(volInfo.indexEntries.data(), volInfo.indexTableLength);
 
@@ -386,22 +387,9 @@ namespace Archives
 		} while (numBytesRead != 0);
 	}
 
-	// Writes a section tag to the open output file.
-	void VolFile::WriteTag(StreamWriter& volWriter, uint32_t length, const std::array<char, 4>& tagName)
-	{
-		SectionHeader sectionHeader(tagName, length);
-
-		try {
-			volWriter.Write(sectionHeader);
-		}
-		catch (std::exception& e) {
-			throw std::runtime_error("Unable to write tag " + std::string(tagName.data(), tagName.size()) + ". Internal Error: " + e.what());
-		}
-	}
-
 	// Reads a tag in the .vol file and returns the length of that section.
 	// If tag does not match what is in the file or if the length is invalid then an error is thrown.
-	uint32_t VolFile::ReadTag(const std::array<char, 4>& tagName)
+	uint32_t VolFile::ReadTag(std::array<char, 4> tagName)
 	{
 		SectionHeader tag;
 		archiveFileReader.Read(tag);
--- a/Archives/VolFile.cpp
+++ b/Archives/VolFile.cpp
@@ -9,6 +9,14 @@
 
 namespace Archives
 {
+	// Volume section header tags
+	const std::array<char, 4> TagVOL_{ 'V', 'O', 'L', ' ' }; // Volume file tag
+	const std::array<char, 4> TagVOLH{ 'v', 'o', 'l', 'h' }; // Header tag
+	const std::array<char, 4> TagVOLS{ 'v', 'o', 'l', 's' }; // Filename table tag
+	const std::array<char, 4> TagVOLI{ 'v', 'o', 'l', 'i' }; // Index table tag
+	const std::array<char, 4> TagVBLK{ 'V', 'B', 'L', 'K' }; // Packed file tag
+
+
 	VolFile::VolFile(const char *fileName) : ArchiveFile(fileName), archiveFileReader(fileName)
 	{
 		m_ArchiveFileSize = archiveFileReader.Length();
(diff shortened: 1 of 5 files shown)


Archives/VolFile.cpp
```diff
@@ -378,49 +387,37 @@ namespace Archives
...
 			throw std::runtime_error("The tag " + std::string(tagName.data(), tagName.size()) + 
 				" was not found in the proper position in volume " + m_ArchiveFileName);
 		}
 
-		uint32_t length = 0;
-		archiveFileReader.Read(&length, sizeof(length));
-
-		// Check for the tag (MSB set)
-		if ((length & 0x80000000) != 0x80000000) {
+		if (tag.padding == VolPadding::TwoByte) {
```
DanRStevens: This is wonderful syntax

Archives/VolFile.cpp
```diff
@@ -506,4 +503,8 @@ namespace Archives
 		}
 		m_NumberOfPackedFiles = packedFileCount;
 	}
+
+	VolFile::SectionHeader::SectionHeader() {}
+	VolFile::SectionHeader::SectionHeader(std::array<char, 4> tag, uint32_t length, VolPadding padding) 
+		: tag(tag), length(length), padding(padding) {}
```
DanRStevens: Good, though I'm not sure if this needs to be explicit. Public struct fields can be initialized with brace syntax. There may be some benefits to using constructor notation though. Maybe experiment with using `= default`, at least for the empty constructor.

Archives/VolFile.h
```diff
@@ -53,11 +53,20 @@ namespace Archives
 			CompressionType compressionType;
 		};
 
+		// Specify boundary padding for a volume file section
+		enum class VolPadding : uint32_t
+		{
+			TwoByte = 0,
+			FourByte = 1 // Default value
```
DanRStevens: Rather than a comment, perhaps this can be written as:
```
enum class VolPadding : uint32_t
{
	TwoByte = 0,
	FourByte = 1,
	Default = FourByte
}
```

Archives/VolFile.h
```diff
@@ -53,11 +53,20 @@ namespace Archives
...
+		};
+
 		struct SectionHeader
 		{
+			SectionHeader();
+			SectionHeader(std::array<char, 4> tag, uint32_t length, VolPadding padding = VolPadding::FourByte);
 			std::array<char, 4> tag;
 			uint32_t length : 31;
-			uint32_t padding : 1; // 0 = 4-byte boundary padding, 1 = 2-byte boundary padding
+			VolPadding padding : 1;
```
DanRStevens: You may want to verify `sizeof(SectionHeader)` here. I believe in some cases, changing the type can break apart bit fields into separate storage locations. In particular, `uint32_t` is not `VolPadding`. It is close though, in that it's an `enum` based on `uint32_t`, so it might still work.

I'd recommend adding a `static_assert` check to ensure the size of the struct is and always remains the size required by the file format:
```
static_assert(sizeof(SectionHeader) == 8, "SectionHeader not of required size");
```

Archives/VolFile.h
```diff
@@ -79,7 +88,8 @@ namespace Archives
 		};
 
 		uint32_t ReadTag(const std::array<char, 4>& tagName);
-		void WriteTag(StreamWriter& volWriter, uint32_t length, const char *tagText);
+		//void WriteTag(StreamWriter& volWriter, uint32_t length, const char *tagText);
+		void VolFile::WriteTag(StreamWriter& volWriter, uint32_t length, const std::array<char, 4>& tagName);
```
DanRStevens: You have a `VolFile::` prefix here.

Should also remove the old commented out code so it doesn't confuse people in the future.

Or maybe consider removing this function entirely, and just passing a constructed `StreamHeader` directly to `volWriter.Write`. It should be possible to even pass an unnamed temporary value. Though that may lose out on the nice exception message you provided in the function.

Archives/VolFile.cpp
```diff
@@ -9,6 +9,14 @@
 
 namespace Archives
 {
+	// Volume section header tags
+	const std::array<char, 4> TagVOL_{ 'V', 'O', 'L', ' ' }; // Volume file tag
+	const std::array<char, 4> TagVOLH{ 'v', 'o', 'l', 'h' }; // Header tag
+	const std::array<char, 4> TagVOLS{ 'v', 'o', 'l', 's' }; // Filename table tag
+	const std::array<char, 4> TagVOLI{ 'v', 'o', 'l', 'i' }; // Index table tag
+	const std::array<char, 4> TagVBLK{ 'V', 'B', 'L', 'K' }; // Packed file tag
```
DanRStevens: I can live with this. The tags are being initialized by the actual string data. Uses are name checked so there are no typos.

I wonder if there might be a way to improve the intialization syntax using the StackOverflow answer from [How to initializez a `std::array<char, n>` with a string literal, omitting the trailing zero](https://stackoverflow.com/questions/33484233/how-to-initialize-a-stdarraychar-n-with-a-string-literal-omitting-the-trail). Admittedly I don't entirely understand the answer, though it looks short enough to study and figure out.

Archives/WaveFile.h
```diff
@@ -1,13 +1,14 @@
...
 namespace Archives
 {
-	const uint32_t RIFF = 0x46464952;	// "RIFF" Resource Interchange File Format
-	const uint32_t WAVE = 0x45564157;	// "WAVE"
-	const uint32_t FMT = 0x20746D66;	// "fmt "
-	const uint32_t DATA = 0x61746164;	// "data"
+	const std::array<char, 4> tagRIFF{ 'R', 'I', 'F', 'F' };
+	const std::array<char, 4> tagWAVE{ 'W', 'A', 'V', 'E' };
+	const std::array<char, 4> tagFMT_{ 'f', 'm', 't', ' ' };
+	const std::array<char, 4> tagDATA{ 'd', 'a', 't', 'a' };
```
DanRStevens: This is a little easier to verify for correctness. I like that.

Archives/VolFile.cpp
```diff
@@ -264,12 +272,13 @@ namespace Archives
 	void VolFile::WriteHeader(StreamWriter& volWriter, const CreateVolumeInfo &volInfo)
 	{
 		// Write the header
-		WriteTag(volWriter, volInfo.paddedStringTableLength + volInfo.paddedIndexTableLength + 24, "VOL ");
+		WriteTag(volWriter, volInfo.paddedStringTableLength + volInfo.paddedIndexTableLength + 24, TagVOL_);
 
-		WriteTag(volWriter, 0, "volh");
+		WriteTag(volWriter, 0, TagVOLH);
+		//WriteTag(volWriter, 0, "volh");
```
DanRStevens: Let's clear out this old line of code.

Brett208: Sounds good.

Archives/VolFile.cpp
```diff
@@ -378,49 +387,37 @@ namespace Archives
 	}
 
 	// Writes a section tag to the open output file.
-	void VolFile::WriteTag(StreamWriter& volWriter, uint32_t length, const char *tagText)
+	void VolFile::WriteTag(StreamWriter& volWriter, uint32_t length, const std::array<char, 4>& tagName)
```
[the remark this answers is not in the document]
Brett208: Agreed, I started not passing by reference later in the changes but never made it back to pull the reference out here. Thanks for catching.

Archives/VolFile.cpp
```diff
@@ -378,49 +387,37 @@ namespace Archives
...
 	{
-		int buffer[2];
-
-		buffer[0] = *(int*)tagText;
-		// Set padding bit to indicate 4 byte padding
-		buffer[1] = length | 0x80000000;
+		SectionHeader sectionHeader(tagName, length);
```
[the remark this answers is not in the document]
Brett208: I set the constructor so if the padding value is not added then it will automatically be set to 1. 

What do you think of this format, where an argument has a default value? Maybe it isn't appropriate here?

Archives/VolFile.cpp
Brett208: Thanks, yeah I think it worked well.

Archives/VolFile.h
Brett208: I didn't actually know you could use default like this on a c++ enum. I think that it might be a good idea to put the default in so it is clear to maybe a less astute reader who maybe doesn't notice/understand the optional argument syntax in the constructor.

Archives/VolFile.cpp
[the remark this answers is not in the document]
DanRStevens: You're right. I didn't see that until after I submitted the review. I was focused on the implementation of the constructor, where the default value isn't listed.

The code is fine as is.

Archives/VolFile.h
DanRStevens: If the constructor then sets the default value using `Default`, it makes the constructor less self-documenting.

If another class is written, where someone wanted to default to 2-byte padding, it would be weird to reuse the existing enum.

Archives/VolFile.h
Brett208: This aspect of bit fields is certainly beyond my knowledge of the language. I placed the static_assert in the constructor of VolFile. It didn't give me any errors, so I guess on MSVC it is okay. 

Let me know if it is more appropriate to place it somewhere else though. I've never actually used static_assert before.

Archives/VolFile.h
Brett208: On further review, I think this function should be removed entirely. The error reporting currently in write tag will only catch problems with the actual write process that are not specific to the tag. It would just catch if there was a generic write problem that happened to occur during the write process, if that makes sense. 

I should be better about deleting the class prefix in the header files. I tend to create the function first, then paste the prototype into the header file and get sloppy about cleaning it up.

Archives/VolFile.cpp
Brett208: I have split feelings about this. I wish there was a built in language solution to this problem. For these short arrays, it isn't a big deal though it would be nice to improve the syntax.

Archives/VolFile.h
DanRStevens: I would place it right after the struct definition. It is a compile time check, so it doesn't need to be placed within a function.

Archives/VolFile.cpp
DanRStevens: Let's leave this for another day then. It may involve some learning and experimentation to come up with a satisfactory solution. I don't want to hold up the other changes on this.

Archives/VolFile.h
```diff
@@ -57,7 +57,8 @@ namespace Archives
 		enum class VolPadding : uint32_t
 		{
 			TwoByte = 0,
-			FourByte = 1 // Default value
+			FourByte = 1,
+			default = FourByte
```
DanRStevens: Err, `default` is a reserved word. This might cause problems. The syntax highlighting on GitHub kind of points that out here.

Archives/VolFile.cpp
```diff
@@ -19,6 +19,8 @@ namespace Archives
 
 	VolFile::VolFile(const char *fileName) : ArchiveFile(fileName), archiveFileReader(fileName)
 	{
+		static_assert(sizeof(SectionHeader) == 8, "SectionHeader not of required size");
```
DanRStevens: This should be placed right after the `struct SectionHeader` definition. It doesn't need to be in a function, so it makes sense to put it right after the struct it applies to.

Brett208:
Okay, I thought it involved using the default keyword in enum, not defining a new member that was Default, but not the keyword. I deleted it. Also moved the static_assert to the header file.

PR #71, opened 2018-06-21
Improve stream classes member variable initialization

Brett208:
This should improve error checking in case we ever try to set what should be a const variable after initialization. I'm not sure, but there may be some minor efficiency gains by using const and constructor initialization.

Also took this opportunity to standardize use of size_t in stream code.

### DIFF
--- a/Streams/MemoryStreamReader.h
+++ b/Streams/MemoryStreamReader.h
@@ -18,7 +18,7 @@ protected:
 	void ReadImplementation(void* buffer, std::size_t size) override;
 
 private:
-	char* streamBuffer;
-	size_t streamSize;
+	const char* streamBuffer;
+	const size_t streamSize;
 	size_t position;
 };
--- a/Streams/MemoryStreamReader.h
+++ b/Streams/MemoryStreamReader.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "SeekableStreamReader.h"
+#include <cstddef>
 #include <cstdint>
 
 class MemoryStreamReader : public SeekableStreamReader {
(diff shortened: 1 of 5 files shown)


Streams/MemoryStreamReader.h
```diff
@@ -18,7 +19,7 @@ class MemoryStreamReader : public SeekableStreamReader {
 	void ReadImplementation(void* buffer, std::size_t size) override;
 
 private:
-	char* streamBuffer;
-	size_t streamSize;
-	size_t position;
+	const char* streamBuffer;
```
DanRStevens: If you want to be really pedantic here, you can use:
```
const char* const streamBuffer;
```
That makes the pointer itself `const`, not just the data being pointed to.

Streams/MemoryStreamWriter.h
```diff
@@ -24,8 +25,8 @@ class MemoryStreamWriter : public SeekableStreamWriter
 	char* streamBuffer;
```
DanRStevens: Similarly, this can be declared as:
```
char* const streamBuffer;
```
The data is still writable, but the pointer itself is then `const`.

DanRStevens:
Agreed on the better error checking. I don't normally consider `const` in places like that, but it is a good improvement.

In terms of efficiency, I doubt there will be any difference from using `const`. Const seems to be more about correctness than efficiency. In some cases guarantees about correctness can enable certain optimizations that would otherwise be a compile error. I'm not aware of any optimizations that would be silently enabled for good code by adding const. Very interesting to think about though.

Brett208:
Good call on making the pointers const. I had forgotten about that feature. I'll apply them and then merge. 

I think the next step will be applying the comments made on the already merged MapReader/Writer branch from a few days ago.

I would never have thought about destruction occuring as you mentioned. Interesting pitfall that I'm certain to find the hard was someday in the future. :)

Thanks,
Brett

PR #73, opened 2018-06-22
Refactor MapReader/Writer

Brett208:
- Remove single line functions from both files
 - Switch from using a c-style array to a std::array in function ReadTilesetHeader
 - When Reading/Writing containers, standardize placing size() before sizeof(X)

I mistakenly committed the first changeset straight to the master. The changeset was standardizing size_t calls in the Map code. See https://github.com/OutpostUniverse/OP2Utility/commit/5b8b693aa8a715a2d6161fac9a00a38b3278e650

### DIFF
--- a/Maps/MapReader.cpp
+++ b/Maps/MapReader.cpp
@@ -19,6 +19,8 @@ namespace MapReader {
 		void ReadTileGroups(SeekableStreamReader& streamReader, MapData& mapData);
 		TileGroup ReadTileGroup(StreamReader& streamReader);
 		std::string ReadString(StreamReader& streamReader);
+
+		const std::array<char, 10> tilesetHeader{ "TILE SET\x1a" };
 	}
 
 
@@ -73,7 +75,7 @@ namespace MapReader {
 			std::array<char, 10> buffer;
 			streamReader.Read(buffer);
 
-			if (buffer != std::array<char, 10>{ "TILE SET\x1a" }) {
+			if (buffer != tilesetHeader) {
 				throw std::runtime_error("'TILE SET' string not found.");
 			}
 		}
--- a/Maps/MapReader.cpp
+++ b/Maps/MapReader.cpp
@@ -3,6 +3,7 @@
 #include <iostream>
 #include <stdexcept>
 #include <vector>
+#include <array>
 #include <cstring>
 
 
@@ -10,9 +11,7 @@ namespace MapReader {
 	// Anonymous namespace to hold private methods
 	namespace {
 		void SkipSaveGameHeader(SeekableStreamReader& streamReader);
-		void ReadHeader(StreamReader& streamReader, MapData& mapData);
 		void ReadTiles(StreamReader& streamReader, MapData& mapData);
-		void ReadClipRect(StreamReader& streamReader, ClipRect& clipRect);
 		void ReadTilesetSources(StreamReader& streamReader, MapData& mapData);
 		void ReadTilesetHeader(StreamReader& streamReader);
 		void ReadTileInfo(StreamReader& streamReader, MapData& mapData);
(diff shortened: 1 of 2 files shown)


Maps/MapReader.cpp
```diff
@@ -63,28 +62,18 @@ namespace MapReader {
...
-
 		void ReadTilesetHeader(StreamReader& streamReader)
 		{
-			char buffer[10];
-			streamReader.Read(buffer, sizeof(buffer));
+			std::array<char, 10> buffer;
+			streamReader.Read(buffer);
 
-			if (std::strncmp(buffer, "TILE SET\x1a", sizeof(buffer))) {
+			if (buffer != std::array<char, 10>{ "TILE SET\x1a" }) {
```
DanRStevens: I like using `array<char, 10>` here, definitely cleaner.

Maybe we should declare a global/static `const` for the `std::array<char, 10>{ "TILE SET\x1a" }` at the top of the .cpp file. That should move the initialization cost from run time to compile time. Basically the same as what was done for the VOL and CLM tags.

Maps/MapWriter.cpp
```diff
@@ -91,23 +71,18 @@ namespace MapWriter {
...
-			streamWriter.Write(tileInfos.data(), sizeof(TileInfo) * tileInfos.size());
+			streamWriter.Write(tileInfos.data(), tileInfos.size() * sizeof(TileInfo));
 		}
 
 		void WriteTerrainType(StreamWriter& streamWriter, const std::vector<TerrainType>& terrainTypes)
 		{
 			WriteContainerSize(streamWriter, terrainTypes.size());
 
-			streamWriter.Write(terrainTypes.data(), sizeof(TerrainType) * terrainTypes.size());
+			streamWriter.Write(terrainTypes.data(), terrainTypes.size() * sizeof(TerrainType));
```
DanRStevens: =)

These two changes did not go unnoticed

Maps/MapWriter.cpp
```diff
@@ -122,7 +97,7 @@ namespace MapWriter {
 				streamWriter.Write(tileGroup.tileHeight);
 
 				streamWriter.Write(tileGroup.mappingIndices.data(), 
-					sizeof(int) * tileGroup.mappingIndices.size());
+					tileGroup.mappingIndices.size()* sizeof(int));
```
DanRStevens: ... these 3 changes :wink:

DanRStevens:
The changes on master look fine.

----

It's kind of easy to commit things to the wrong branch in Git. If you forgot to create a branch, you can branch from the current commit, and you'll have a new branch with the same current history:
```
git checkout b new_branch_name
```

To fix master, you can switch back to master and reset the HEAD so the tip of the branch points to a previous commit:
```
git reset id_of_new_head_commit
```

As a shortcut, you can use `HEAD^` or `HEAD~` as the id for the parent of the current commit. Use `HEAD~n` to go back `n` generations, favouring first parents in the case of any merge commits along the way. On a merge of `m` branches simultaneously, the syntax `HEAD^n`, where `n <= m`, means the `n`-th parent of a merge commit, which is always 1 generation back. That is, merge commits have more than one parent.

The reset just moves the branch tip pointer. It doesn't change files. After the reset, you'll still have the changes present in the working copy, but they will now be marked as changed since the branch tip no longer includes the changes. You can recommit at this point, or use `git checkout .` to discard them.

Side note: With Git default settings, unreferenced commits are not garbage collected for 30 days. So if you reset a branch, and then discard local changes, but never created a new branch to hold the changes, they're actually still there in the repo. You can find them using the ref log. Basically, it's pretty safe to delete stuff that was previously committed without really losing it. At least assuming you've realized your mistake and learn how to use the ref log within 30 days.

Maps/MapReader.cpp
Brett208: Sounds good.

Issue #62, opened 2018-06-11
Consider using `std::size_t` instead of `size_t`


DanRStevens:
There is a subtle [difference between `size_t` and `std::size_t`](https://stackoverflow.com/questions/5813700/difference-between-size-t-and-stdsize-t).

Standard C types were imported into C++ and added to the `std` namespace. The standard guarantees the global C type (if it is defined) will match the corresponding `std` type. I've noticed the standard doesn't seem to require the types be available by default, but allows for `#include` of any standard library to define them.

As this is a C++ project, it might be better if we use `std::size_t`. We might also want to pay attention to any required header files. On cppreference.com, the [`size_t`](https://en.cppreference.com/w/cpp/types/size_t) page lists a few headers that define the type. A common one I see listed for the type is `<cstddef>`.

I tried running the new version of Clang on the project, and it barfed all over the `size_t` types, saying they were undefined.

----

A related but slightly different matter, is I think we are using `size_t` more often than we should. For-loop variables can and probably should be declared with auto when possible, at least for the new style for-loops. We also have `size_t` in a few places where a fixed size type would be more appropriate.

I recommend we update the code to reduce the number of explicit uses of `size_t`, and change the type in a few instances where another type would be more suitable.

I also recommend we use `std::size_t` for the instances that remain.

Brett208:
MSVC seems quite happy with the current size_t code, but of course we need it to work on the Linux compiler as well. @DanRStevens, is the primary compiler we want to support for Linux builds Clang?

In the future, I'll use the cheat sheet below and make sure at least one of these headers is defined in a file when using size_t. If none of the headers are present, I'll use cstddef by default. (https://en.cppreference.com/w/cpp/types/size_t):
 - header <cstddef>
 - header <cstdio>  
 - header <cstdlib>  
 - header <cstring>  
 - header <ctime>  
 - header <cwchar>

I'll start writing it as std::size_t. 

I've typically steered away from using the auto keyword as I prefer seeing the actual type when reading code instead of trying to determine it in context (hovering and using the tooltip also works in Visual Studio). This seems to be a bit against standard practice so I'm game trying to use auto more, especially in range based loops. 

I'm guessing that if the auto keyword resolves to size_t, one of the headers containing size_t must be present to satisfy the Clang compiler? I'm guessing since vector and array use size_t for their length, they also define size_t?

Thanks for pointing this out,
Brett

DanRStevens:
[...]

PR #75, opened 2018-06-22
Refactor utility classes

Brett208:
- Standardize use of size_t
 - Remove std using statements
 - Use the fs alias for referencing experimental filesystem namespace
 - Increase use of auto keyword on range based loops

I have no problem including the standard namespace in .cpp files. However, to match the rest of the code base, I thought it would be better to remove it. It also seems to be the prevalent way to write C++ right now. 

With this and the 2 previous branches, we should be able to close out issue #62. 

There is significant room for improvement in these classes, especially `ResourceManager`. I would be okay ignoring the needed work for now in the interest of continuing with other goals.

### DIFF
--- a/ResourceManager.cpp
+++ b/ResourceManager.cpp
@@ -138,7 +138,7 @@ void ResourceManager::ExtractAllOfFileType(const std::string& directory, const s
 
 bool ResourceManager::DuplicateFilename(std::vector<std::string>& currentFilenames, std::string pathToCheck)
 {
-	// Brett: When called on a large loop of filenames (60 more more) this function, this will create a bottleneck.
+	// Brett: When called on a large loop of filenames (60 or more), this function will create a bottleneck.
 
 	std::string filename = XFile::GetFilename(pathToCheck);
 
--- a/ResourceManager.cpp
+++ b/ResourceManager.cpp
@@ -10,13 +10,13 @@ ResourceManager::ResourceManager(const std::string& archiveDirectory)
 {
 	std::vector<std::string> volFilenames = XFile::GetFilesFromDirectory(archiveDirectory, ".vol");
 
-	for (const std::string& volFilename : volFilenames) {
+	for (const auto& volFilename : volFilenames) {
 		ArchiveFiles.push_back(std::make_unique<VolFile>(volFilename.c_str()));
 	}
 
 	std::vector<std::string> clmFilenames = XFile::GetFilesFromDirectory(archiveDirectory, ".clm");
 
-	for (const std::string& clmFilename : clmFilenames) {
+	for (const auto& clmFilename : clmFilenames) {
 		ArchiveFiles.push_back(std::make_unique<ClmFile>(clmFilename.c_str()));
 	}
 }
@@ -33,7 +33,7 @@ std::unique_ptr<SeekableStreamReader> ResourceManager::GetResourceStream(const s
 		return nullptr;
 	}
 
-	for (std::unique_ptr<ArchiveFile>& archiveFile : ArchiveFiles)
+	for (auto& archiveFile : ArchiveFiles)
 	{
 		std::string internalArchiveFilename = XFile::GetFilename(filename);
 		int internalArchiveIndex = archiveFile->GetInternalFileIndex(internalArchiveFilename.c_str());
@@ -52,7 +52,7 @@ std::vector<std::string> ResourceManager::GetAllFilenames(const std::string& dir
 
 	std::vector<std::string> filenames = XFile::GetFilesFromDirectory(directory, filenameRegex);
 
-	for (std::unique_ptr<ArchiveFile>& archiveFile : ArchiveFiles)
+	for (auto& archiveFile : ArchiveFiles)
 	{
 		for (int i = 0; i < archiveFile->GetNumberOfPackedFiles(); ++i)
 		{
@@ -125,7 +125,7 @@ bool ResourceManager::ExtractSpecificFile(const std::string& filename, bool over
 
 void ResourceManager::ExtractAllOfFileType(const std::string& directory, const std::string& extension, bool overwrite)
 {
-	for (std::unique_ptr<ArchiveFile>& archiveFile : ArchiveFiles)
+	for (auto& archiveFile : ArchiveFiles)
 	{
 		for (int i = 0; i < archiveFile->GetNumberOfPackedFiles(); ++i)
 		{
@@ -153,7 +153,7 @@ bool ResourceManager::DuplicateFilename(std::vector<std::string>& currentFilenam
 
 std::string ResourceManager::FindContainingArchiveFile(const std::string& filename)
 {
-	for (std::unique_ptr<ArchiveFile>& archiveFile : ArchiveFiles)
+	for (auto& archiveFile : ArchiveFiles)
 	{
 		int internalFileIndex = archiveFile->GetInternalFileIndex(filename.c_str());
 
--- a/ResourceManager.cpp
+++ b/ResourceManager.cpp
@@ -8,13 +8,13 @@ using namespace Archives;
 
 ResourceManager::ResourceManager(const std::string& archiveDirectory)
 {
-	std::vector<std::string> volFilenames = XFile::GetFilesFromDirectory(archiveDirectory, ".vol");
+	auto volFilenames = XFile::GetFilesFromDirectory(archiveDirectory, ".vol");
 
 	for (const auto& volFilename : volFilenames) {
 		ArchiveFiles.push_back(std::make_unique<VolFile>(volFilename.c_str()));
 	}
 
-	std::vector<std::string> clmFilenames = XFile::GetFilesFromDirectory(archiveDirectory, ".clm");
+	auto clmFilenames = XFile::GetFilesFromDirectory(archiveDirectory, ".clm");
 
 	for (const auto& clmFilename : clmFilenames) {
 		ArchiveFiles.push_back(std::make_unique<ClmFile>(clmFilename.c_str()));
@@ -33,7 +33,7 @@ std::unique_ptr<SeekableStreamReader> ResourceManager::GetResourceStream(const s
 		return nullptr;
 	}
 
-	for (auto& archiveFile : ArchiveFiles)
+	for (const auto& archiveFile : ArchiveFiles)
 	{
 		std::string internalArchiveFilename = XFile::GetFilename(filename);
 		int internalArchiveIndex = archiveFile->GetInternalFileIndex(internalArchiveFilename.c_str());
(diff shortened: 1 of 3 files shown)


ResourceManager.cpp
```diff
@@ -1,42 +1,41 @@
...
 #include "XFile.h"
 
-using namespace std;
 using namespace Archives;
 
-ResourceManager::ResourceManager(const string& archiveDirectory)
+ResourceManager::ResourceManager(const std::string& archiveDirectory)
 {
-	vector<string> volFilenames = XFile::GetFilesFromDirectory(archiveDirectory, ".vol");
+	std::vector<std::string> volFilenames = XFile::GetFilesFromDirectory(archiveDirectory, ".vol");
```
DanRStevens: This could be a good place to use `auto`. It saves typing, and really, we are unlikely to care if these are returned as a `vector` or as a `list` or as something else, so long as we can iterate over the filenames in a standard way.

Effectively, we are over-specifying by not using `auto` here. Kind of like specifying `FileStreamReader&` when any `StreamReader&` would do.

ResourceManager.cpp
```diff
@@ -1,42 +1,41 @@
...
 
-	for (const string& volFilename : volFilenames) {
-		ArchiveFiles.push_back(make_unique<VolFile>(volFilename.c_str()));
+	for (const auto& volFilename : volFilenames) {
+		ArchiveFiles.push_back(std::make_unique<VolFile>(volFilename.c_str()));
 	}
 
-	vector<string> clmFilenames = XFile::GetFilesFromDirectory(archiveDirectory, ".clm");
+	std::vector<std::string> clmFilenames = XFile::GetFilesFromDirectory(archiveDirectory, ".clm");
```
DanRStevens: Ditto.

ResourceManager.cpp
```diff
@@ -1,42 +1,41 @@
...
-		return make_unique<FileStreamReader>(filename);
+		return std::make_unique<FileStreamReader>(filename);
 	}
 
 	if (!accessArchives) {
 		return nullptr;
 	}
 
-	for (unique_ptr<ArchiveFile>& archiveFile : ArchiveFiles)
+	for (auto& archiveFile : ArchiveFiles)
```
DanRStevens: We can probably use const here. It would apply to the unique_ptr itself here.

Come to think of it though, the archive objects themselves can probably be marked const, since they are not modified after being opened, and they are opened in the constructor. Modifications are only done during re-packing, which is not handled by members of this class.

ResourceManager.cpp
```diff
@@ -1,42 +1,41 @@
...
 	{
-		string internalArchiveFilename = XFile::GetFilename(filename);
+		std::string internalArchiveFilename = XFile::GetFilename(filename);
```
DanRStevens: I suppose this could be `auto` as well, though I don't think it matters which way we go for simple cases like this.

ResourceManager.cpp
```diff
@@ -47,17 +46,17 @@ unique_ptr<SeekableStreamReader> ResourceManager::GetResourceStream(const string
...
 }
 
-vector<string> ResourceManager::GetAllFilenames(const string& directory, const string& filenameRegexStr, bool accessArcives)
+std::vector<std::string> ResourceManager::GetAllFilenames(const std::string& directory, const std::string& filenameRegexStr, bool accessArcives)
 {
-	regex filenameRegex(filenameRegexStr, regex_constants::icase);
+	std::regex filenameRegex(filenameRegexStr, std::regex_constants::icase);
 
-	vector<string> filenames = XFile::GetFilesFromDirectory(directory, filenameRegex);
+	std::vector<std::string> filenames = XFile::GetFilesFromDirectory(directory, filenameRegex);
```
DanRStevens: This should probably be `auto`

ResourceManager.cpp
```diff
@@ -47,17 +46,17 @@ unique_ptr<SeekableStreamReader> ResourceManager::GetResourceStream(const string
...
 
-	for (unique_ptr<ArchiveFile>& archiveFile : ArchiveFiles)
+	for (auto& archiveFile : ArchiveFiles)
```
DanRStevens: I think this can be `const`

ResourceManager.cpp
```diff
@@ -66,19 +65,19 @@ vector<string> ResourceManager::GetAllFilenames(const string& directory, const s
 	return filenames;
 }
 
-vector<string> ResourceManager::GetAllFilenamesOfType(const string& directory, const string& extension, bool accessArchives)
+std::vector<std::string> ResourceManager::GetAllFilenamesOfType(const std::string& directory, const std::string& extension, bool accessArchives)
 {
-	vector<string> filenames = XFile::GetFilesFromDirectory(directory, extension);
+	std::vector<std::string> filenames = XFile::GetFilesFromDirectory(directory, extension);
```
DanRStevens: This can be `auto`

ResourceManager.cpp
```diff
@@ -66,19 +65,19 @@ vector<string> ResourceManager::GetAllFilenames(const string& directory, const s
...
 
 	if (!accessArchives) {
 		return filenames;
 	}
 
-	for (unique_ptr<ArchiveFile>& archiveFile : ArchiveFiles)
+	for (std::unique_ptr<ArchiveFile>& archiveFile : ArchiveFiles)
```
DanRStevens: This can probably be `const auto&`

ResourceManager.cpp
```diff
@@ -66,19 +65,19 @@ vector<string> ResourceManager::GetAllFilenames(const string& directory, const s
...
 	{
 		for (int i = 0; i < archiveFile->GetNumberOfPackedFiles(); ++i)
 		{
-			string internalFilename = archiveFile->GetInternalFileName(i);
+			std::string internalFilename = archiveFile->GetInternalFileName(i);
```
DanRStevens: Could be `auto`, but I don't think it matters in this case

ResourceManager.cpp
```diff
@@ -124,9 +123,9 @@ bool ResourceManager::ExtractSpecificFile(const string& filename, bool overwrite
 	return false;
 }
 
-void ResourceManager::ExtractAllOfFileType(const string& directory, const string& extension, bool overwrite)
+void ResourceManager::ExtractAllOfFileType(const std::string& directory, const std::string& extension, bool overwrite)
 {
-	for (unique_ptr<ArchiveFile>& archiveFile : ArchiveFiles)
+	for (auto& archiveFile : ArchiveFiles)
```
DanRStevens: Can probably be `const` as well

ResourceManager.cpp
```diff
@@ -137,13 +136,13 @@ void ResourceManager::ExtractAllOfFileType(const string& directory, const string
...
 }
 
-bool ResourceManager::DuplicateFilename(vector<string>& currentFilenames, string pathToCheck)
+bool ResourceManager::DuplicateFilename(std::vector<std::string>& currentFilenames, std::string pathToCheck)
 {
-	// Brett: When called on a large loop of filenames (60 more more) this function, this will create a bottleneck.
+	// Brett: When called on a large loop of filenames (60 or more), this function will create a bottleneck.
 
-	string filename = XFile::GetFilename(pathToCheck);
+	std::string filename = XFile::GetFilename(pathToCheck);
```
DanRStevens: Could be `auto`, but I don't think it matters here

ResourceManager.cpp
```diff
@@ -152,9 +151,9 @@ bool ResourceManager::DuplicateFilename(vector<string>& currentFilenames, string
 	return false;
 }
 
-string ResourceManager::FindContainingArchiveFile(const string& filename)
+std::string ResourceManager::FindContainingArchiveFile(const std::string& filename)
 {
-	for (unique_ptr<ArchiveFile>& archiveFile : ArchiveFiles)
+	for (auto& archiveFile : ArchiveFiles)
```
DanRStevens: Could probably be `const`

XFile.cpp
```diff
@@ -40,53 +40,53 @@ bool XFile::ExtensionMatches(const string& pathStr, const string& extension)
...
 	}
 
 	return filenames;
 }
 
-vector<string> XFile::GetFilesFromDirectory(const string& directory, const regex& filenameRegex)
+std::vector<std::string> XFile::GetFilesFromDirectory(const std::string& directory, const std::regex& filenameRegex)
 {
-	vector<string> filenames = GetFilesFromDirectory(directory);
+	std::vector<std::string> filenames = GetFilesFromDirectory(directory);
```
DanRStevens: This could be `auto`

XFile.cpp
```diff
@@ -98,17 +98,17 @@ vector<string> XFile::GetFilesFromDirectory(const string& directory, const regex
 	return filenames;
 }
 
-vector<string> XFile::GetFilesFromDirectory(const string& directory, const string& fileType)
+std::vector<std::string> XFile::GetFilesFromDirectory(const std::string& directory, const std::string& fileType)
 {
-	vector<string> filenames = GetFilesFromDirectory(directory);
+	std::vector<std::string> filenames = GetFilesFromDirectory(directory);
```
DanRStevens: This could be `auto`

ResourceManager.cpp
Brett208: ok

ResourceManager.cpp
Brett208: I don't know that I am ready to pepper the code with this many autos yet.

ResourceManager.cpp
Brett208: sounds good

ResourceManager.cpp
Brett208: agreed, I missed marking it auto.

ResourceManager.cpp
Brett208: agreed

ResourceManager.cpp
Brett208: agreed

Brett208:
I'll have to read up on GitHub keywords stuff. 

I applied the const on the appropriate ranged base loop variables. 

I'll have to think more about where applying auto outside of the ranged based loops is beneficial vs detrimental to the readability of the code base. I checked out a few stack overflow questions about it.

Thanks,
Brett

DanRStevens:
Fair enough. Like I said, there are some elements of style to it, and not everyone agrees where to draw the line.

In the case of simple strings, there is often little chance the data type will change, and it's not very long to type, so it shouldn't matter. When it comes to generic containers I tend to prefer auto since containers types might be refactored for efficiency reasons, while not actually impacting code correctness.

Selecting a container type often comes with specific assumptions, which relate to efficiency more so than it being a collection of objects. For instance, does the iteration order matter. Are elements only ever appended to the end, or can they be inserted in the middle. Is it important that a list of file names is stored contiguously in memory. Does adding a new element potentially require all previously added elements to be copied to a new area in memory. These are all very much to do with efficiency of how the container is used, rather than correctness of it being a container. If you don't rely on those assumptions, you can swap out the container type and still have correct code. Assuming of course the code isn't peppered with the specific type everywhere, making the change more difficult.

PR #79, opened 2018-06-23
Unpack From Volume Files in Chunks

### LINKED ISSUE #76: Extract files from volumes in chunks instead of entire file at once
Consider creation of standardized function for this either that either lives in ArchiveUnpacker or maybe attached to FileStreamWriter.

Brett208:
- Standardize unpack writing to a base function in ArchiveUnpacker.

Closes #76.

### DIFF
--- a/Archives/ArchiveUnpacker.cpp
+++ b/Archives/ArchiveUnpacker.cpp
@@ -3,10 +3,11 @@
 #include "../Streams/SeekableStreamReader.h"
 #include "../Streams/FileStreamWriter.h"
 #include <array>
+#include <cstddef>
 
 namespace Archives
 {
-	const uint32_t ARCHIVE_WRITE_SIZE = 0x00020000;
+	const std::size_t ARCHIVE_WRITE_SIZE = 0x00020000;
 
 	ArchiveUnpacker::ArchiveUnpacker(const std::string& fileName) :
 		m_ArchiveFileName(fileName), m_NumberOfPackedFiles(0), m_ArchiveFileSize(0) { }
--- a/Archives/ArchiveUnpacker.cpp
+++ b/Archives/ArchiveUnpacker.cpp
@@ -1,9 +1,14 @@
 #include "ArchiveUnpacker.h"
 #include "../XFile.h"
+#include "../Streams/SeekableStreamReader.h"
+#include "../Streams/FileStreamWriter.h"
+#include <array>
 
 namespace Archives
 {
-	ArchiveUnpacker::ArchiveUnpacker(const std::string& fileName) : 
+	const uint32_t ARCHIVE_WRITE_SIZE = 0x00020000;
+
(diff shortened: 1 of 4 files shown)


Brett208:
@DanRStevens, what do you think of refactoring the new function in ArchiveUnpacker to live in StreamWriter with the following signature:

```
// Writes from the current position of the passed stream until the end of the stream.
// Takes an optional bufferSize to specify how large of chunks to read each pass.
void WriteFromStream(SeekableStreamReader& seekableStreamReader, size_t bufferSize =  0x00020000);

// Writes from the current position of the passed stream through the specified writeLength.
// Takes an optional bufferSize to specify how large of chunks to read each pass.
void WriteFromStream(StreamReader& streamReader, uint64_t writeLength, size_t bufferSize = 0x00020000);
```

It would require adding a SeekableStreamReader include to StreamWriter, which I'm not necessarily fond of. These 2 functions would both be used in ClmFile and VolFile, so I'm guessing their application would apply to other applications as well.

I just chose the default bufferSize because it is used in other parts of Archives and am not sure if a different default might be better.

Thanks,
Brett

Archives/ArchiveUnpacker.cpp
```diff
@@ -1,9 +1,14 @@
 #include "ArchiveUnpacker.h"
 #include "../XFile.h"
+#include "../Streams/SeekableStreamReader.h"
+#include "../Streams/FileStreamWriter.h"
+#include <array>
 
 namespace Archives
 {
-	ArchiveUnpacker::ArchiveUnpacker(const std::string& fileName) : 
+	const uint32_t ARCHIVE_WRITE_SIZE = 0x00020000;
```
DanRStevens: Should be `std::size_t`

Brett208: Okay, I'll update.

Brett208:
Agreed we can open an issue for the WriteFromStream function. 

What do you mean on the comment "I don't believe seekable properties are needed at all for the copy." I'm not following which part of the source code this is coming from.

Thanks,
Brett

DanRStevens:
I was referring to:
```
void WriteFromStream(SeekableStreamReader& seekableStreamReader, size_t bufferSize =  0x00020000);
```
Curiously, the other method you listed uses just a `StreamReader`.

Brett208:
@DanRStevens, in order to read to the end of the stream without passing a length, SeekableStreamReader would be required here. Where a length to read is specified in the other prototype so the stream could be read just using the StreamReader format.

We can talk about it more in an issue tracker, just wanted to make sure you were not referring to the source code before merging.

-Brett

DanRStevens:
Right, because we have no way to do a partial read without causing an exception, and there are no promises about buffer contents in the case of an exception. Ok, that seems to indicate a weakness in the streams design. You're right, we should address that in the issue tracker.

In an earlier design, I had both a exception throwing read, and a length returning non-exception variant. That would be one approach to the issue.

PR #80, opened 2018-06-23
Remove Final Windows.h Include

### LINKED ISSUE #30: Make Archives codebase cross-platform compatible
Currently Archives codebase contains windows specific implementation details and relies on Windows.h.

Windows specific dependencies are mostly related to file handles and associated code.

### LINKED ISSUE #56: Refactor variables in struct VolFile::CreateVolumeInfo
Refactor VolFile::CreateVolumeInfo and VolFile::IndexEntry to use C++11 containers to reduce custom memory management. Use fixed size type fields to improve compatibility of reading/writing data with different compilers. 

Within CreateVolumeInfo, switch the following fields to std::vector: 
 - IndexEntry *indexEntry;
 - int *fileNameLength;
 - HANDLE *fileHandle;

Within CreateVolumeInfo, make the following fields an unsigned fixed size type: 
 - int stringTableLength;
 - int indexTableLength;
 - int paddedStringTableLength;
 - int paddedIndexTableLength;

Within VolFile::IndexEntry, make the following fields an unsigned fixed size type: 
 - unsigned int fileNameOffset;
 - unsigned int dataBlockOffset;
 - int fileSize;

Brett208:
- Change Volume CreateVolInfo HANDLE out with a FileStreamReader
 - Standardize packing of files to disk between VolFile and ClmFile

And there may be much rejoicing, if it works on Linux...

Closes #56.
Closes #30.

Tested by packing and unpacking both a volume and clm file.

### DIFF
--- a/Archives/ArchivePacker.h
+++ b/Archives/ArchivePacker.h
@@ -1,5 +1,7 @@
 #pragma once
 
+#include "../Streams/StreamReader.h"
+#include "../Streams/StreamWriter.h"
 #include <string>
 #include <vector>
 
@@ -29,5 +31,7 @@ namespace Archives
 		// Compares 2 filenames case insensitive to determine which comes first alphabetically.
 		// Does not compare the entire path, but only the filename.
 		static bool ComparePathFilenames(const std::string path1, const std::string path2);
+
+		void ArchivePacker::PackFile(StreamWriter& streamWriter, StreamReader& fileToPackReader, const uint64_t fileToPackSize);
 	};
 }
--- a/Archives/VolFile.cpp
+++ b/Archives/VolFile.cpp
@@ -1,6 +1,4 @@
 #include "VolFile.h"
-#include "../Streams/MemoryStreamReader.h"
-#include "../Streams/FileStreamWriter.h"
 #include "../Streams/FileSliceReader.h"
 #include "../XFile.h"
 #include <stdexcept>
@@ -214,25 +212,22 @@ namespace Archives
 		// Allowing duplicate names when packing may cause unintended results during binary search and file extraction.
 		CheckSortedContainerForDuplicateNames(volInfo.internalNames);
 
+		volInfo.stringTableLength = 0;
+
 		// Open input files and prepare header and indexing info
-		if (!PrepareHeader(volInfo)) {
-			return false;
-		}
+		PrepareHeader(volInfo);
 
 		try {
 			WriteVolume(volumeFileName, volInfo);
 		}
 		catch (std::exception&) {
-			CleanUpVolumeCreate(volInfo);
 			return false;
 		}
-
-		CleanUpVolumeCreate(volInfo);
 
 		return true;
 	}
 
-	void VolFile::WriteVolume(const std::string& fileName, const CreateVolumeInfo& volInfo) 
+	void VolFile::WriteVolume(const std::string& fileName, CreateVolumeInfo& volInfo) 
 	{
 		FileStreamWriter volWriter(fileName);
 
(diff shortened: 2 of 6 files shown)


Archives/VolFile.cpp
```diff
@@ -214,49 +212,38 @@ namespace Archives
 		// Allowing duplicate names when packing may cause unintended results during binary search and file extraction.
 		CheckSortedContainerForDuplicateNames(volInfo.internalNames);
 
+		volInfo.stringTableLength = 0;
```
DanRStevens: This seems like it would be better placed in `PrepareHeader`.

Archives/VolFile.cpp
```diff
@@ -298,50 +285,30 @@ namespace Archives
...
-			{
-				// Error opening file. Close already opened files and return error
-				for (i--; i >= 0; i--) {
-					CloseHandle(volInfo.fileHandles[i]);
-				}
-
-				return false;
+		for (const auto& filename : volInfo.filesToPack) {
+			try {
+				volInfo.fileStreamReaders.push_back(std::make_unique<FileStreamReader>(filename));
```
DanRStevens: Hmm, unrelated to the current changes, but I just noticed this and wanted to point it out.

There's a bit of a style issue here. Perhaps we should we clear the `fileStreamReaders` collection before the start of this loop. Clearing it does seem a bit silly and redundant, since it will be initialized to empty, and this function *should* only be called once.

I guess what we really want, and to be obvious in a verifiable way, is that the collection of open files corresponds with the list of file names. If this function was called twice, that wouldn't be the case.

What this really calls for is a [`map`](https://en.wikipedia.org/wiki/Map_(higher-order_function)), as exists in most functional programming languages. The concept is implemented in C++ as [`std::transform`](https://en.cppreference.com/w/cpp/algorithm/transform?), however, the syntax is a bit ugly. The examples in the documentation also don't make it look very compelling. If we mapped the filenames to FileStreamReader objects, it would clearly be correct without having to analyse outside of the function.

Anyway, no need to change anything now. If we refactor to use static methods for packing, it will likely touch or eliminate this issue then. Likewise, if we delay opening the files until the point they are being copied, it will also eliminate this issue. The issue being one of style and verifiability, not of correctness.

Archives/VolFile.cpp
```diff
@@ -298,50 +285,30 @@ namespace Archives
...
+		OpenAllInputFiles(volInfo);
 
 		// Get file sizes and calculate length of string table
 		for (std::size_t i = 0; i < volInfo.fileCount(); i++)
 		{
 			IndexEntry indexEntry;
 
-			indexEntry.fileSize = GetFileSize(volInfo.fileHandles[i], 0);
+			// A packed volume cannot be larger than uint32_t
+			indexEntry.fileSize = static_cast<uint32_t>(volInfo.fileStreamReaders[i]->Length());
```
DanRStevens: We should probably check this and throw an exception if it's not true. There's no way to know for certain that a file won't exceed 4GB in size until run time.

Archives/ArchivePacker.h
```diff
@@ -29,5 +31,7 @@ namespace Archives
 		// Compares 2 filenames case insensitive to determine which comes first alphabetically.
 		// Does not compare the entire path, but only the filename.
 		static bool ComparePathFilenames(const std::string path1, const std::string path2);
+
+		void ArchivePacker::PackFile(StreamWriter& streamWriter, StreamReader& fileToPackReader, const uint64_t fileToPackSize);
```
DanRStevens: Prefix `ArchivePacker::` should be removed.

Brett208: I am terrible about this. Thanks.

Archives/VolFile.cpp
Brett208: I'll move it in there. The PrepareHeader function needs to be broken into smaller chunks at some point but I'm having trouble figuring out a good way to break it up since all the calculations are done sort of intermingled.

Archives/VolFile.cpp
Brett208: In the long run I would like to see the write specific portion of VolumeInfo stripped out and use the VolumeInfo struct both when reading and writing volume files. This would probably also affect the implementation. For now I'll just clear the stream container in the function to be explicit.

Archives/VolFile.cpp
Brett208: Good call. We should look at mirroring this in clmFile after we replace all the stream related variables to fixed width in clmFile.

Brett208:
I forgot to push the last changeset before merging. you will see that changeset in the next branch that was merged in.

Additionally, I created a naming collision by creating a WriteFromStream function in both ArchiveUnpacker and ArchivePacker. This collision is fixed by a commit on the master branch.

If we do not move forward with pushing WriteFromStream into streams in the near future, these 2 functions could be merged in some sort of ArchiveHelper.h/.cpp or something.

-Brett

Archives/VolFile.cpp
DanRStevens: Agreed. The write specific portions for VOL creation are probably better as local variables.

Issue #86, opened 2018-06-24
Chained calls to `append` produce compile errors on Linux


DanRStevens:
On Linux, both g++ and clang give errors for chained calls to `append`:
https://github.com/OutpostUniverse/OP2Utility/blob/eb764f239c8f6b8375a5b9697c0d22b36ad622e3/XFile.cpp#L144-L148

The error is:
```
XFile.cpp:144:42: error: no matching member function for call to 'append'
                return fs::path().append(subDirectory).append(filename)....
                       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~
/usr/bin/../lib/gcc/x86_64-linux-gnu/5.4.0/../../../../include/c++/5.4.0/experimental/fs_path.h:223:7: note: 
      candidate template ignored: requirement '__and_<__not_<is_same<path, path>
      >, __constructible_from<path, void> >::value' was not satisfied [with
      _Source = std::experimental::filesystem::v1::__cxx11::path]
      append(_Source const& __source)
      ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/5.4.0/../../../../include/c++/5.4.0/experimental/fs_path.h:231:7: note: 
      candidate function template not viable: requires 2 arguments, but 1 was
      provided
      append(_InputIterator __first, _InputIterator __last)
      ^
XFile.cpp:148:32: error: no matching member function for call to 'append'
        return p.append(subDirectory).append(filename).string();
               ~~~~~~~~~~~~~~~~~~~~~~~^~~~~~
/usr/bin/../lib/gcc/x86_64-linux-gnu/5.4.0/../../../../include/c++/5.4.0/experimental/fs_path.h:223:7: note: 
      candidate template ignored: requirement '__and_<__not_<is_same<path, path>
      >, __constructible_from<path, void> >::value' was not satisfied [with
      _Source = std::experimental::filesystem::v1::__cxx11::path]
      append(_Source const& __source)
      ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/5.4.0/../../../../include/c++/5.4.0/experimental/fs_path.h:231:7: note: 
      candidate function template not viable: requires 2 arguments, but 1 was
      provided
      append(_InputIterator __first, _InputIterator __last)
      ^
```

[...]

Issue #83, opened 2018-06-23
Allow for including C++17 filesystem on both Windows & Linux


Brett208:
VS2017's version of MSVC requires the experimental directory in the namespace to use filesystem.

```
#include <filesystem>

namespace fs = std::experimental::filesystem;
```

Originally the codebase had `#include <experimental/filesystem>`. I just checked today and deleting this line does not affect the code's ability to compile on MSVC. I'm not sure if this is a recent change or if it was never required, although it doesn't throw an error either way.

Perhaps using a preprocessor directive to change how the namespace alias is defined on MSVC verses Clang/the other typical Linux compilers?

Brett208:
@DanRStevens, maybe this one is better for you to tackle since I don't have access to a Linux machine right now.

Brett208:
Just checked, and the following compiles fine on my version of MSVC.

```
#include <experimental/filesystem>

namespace fs = std::experimental::filesystem;
```

So I think we can just use the experimental include and it should work on both MSVC and Clang.

We were including both because I was having trouble getting it to work initially and saw it on a tutorial somewhere. I don't remember the details on why I settled for both now though. It may have been a dumb call on my part or MSVC may have changed since then as I've updated VS a couple of times.

PR #87, opened 2018-06-24
Linux filesystem fixes

DanRStevens:
This gets the code compiling on Linux.

----

Not part of the PR, but as it relates to these changes:

I don't think we need to `.string()` calls for the return values. At least, it compiles on Linux without them. Not sure how portable that is though to rely on automatic conversion to string, so I left out the changes from the PR. If you're curious to try:
```
	if (p == filename) {
		return fs::path() / subDirectory / filename;
	}

	p = p.remove_filename();
	return p / subDirectory / filename;
```

Additionally, the first `return` I think could be simplified to:
```
		return fs::path(subDirectory) / filename;
```

### DIFF
--- a/XFile.cpp
+++ b/XFile.cpp
@@ -2,7 +2,7 @@
 #include "StringHelper.h"
 #include <cstddef>
 
-#include <filesystem>
+#include <experimental/filesystem>
 
 namespace fs = std::experimental::filesystem;
 
--- a/XFile.cpp
+++ b/XFile.cpp
@@ -141,11 +141,11 @@ std::string XFile::AppendSubDirectory(const std::string& pathStr, const std::str
 	fs::path filename(p.filename());
 
 	if (p == filename) {
-		return fs::path().append(subDirectory).append(filename).string();
+		return (fs::path() / subDirectory / filename).string();
 	}
 
 	p = p.remove_filename();
-	return p.append(subDirectory).append(filename).string();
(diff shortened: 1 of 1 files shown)


DanRStevens:
This fixes #83

DanRStevens:
This fixes #86. (Though doesn't explain what the original problem was).

Brett208:
@DanRStevens you might already be aware of this, but for fixes #X to work, it has to be placed in the initial pull request message or a commit. So, when this is merged, it will not close #86 or #83. Check out https://github.com/OutpostUniverse/OP2Utility/pull/79. See how closes in the initial message has the dotted line under it. When you see that, it will close the issue when the merge is made.

---

These changes test without issue on MSVC. I wasn't aware that you could use the division operator this way in the filesystem library. Good thing I read the documentation so closely.

Anyways, I tested removing the string() function as you mentioned. It appears to be required on MSVC. It gives the following compile time error when removed.

Error (active) E0312 no suitable user-defined conversion from "std::experimental::filesystem::v1::path" to "std::string" exists OP2Utility	c:\ . . . OP2Archive\OP2Utility\XFile.cpp 144	

Perhaps make a note that it is required by MSVC and not Clang?

As another side note, when I was testing this function, the following input arguments give the results .\\.\\art.vol on Windows. This works fine in the program, but seems sloppy.

pathStr = .\\art.vol
subDirectory = ./

Thanks,
Brett

PR #90, opened 2018-06-24
Fix LZH reading code

### LINKED ISSUE #85: Extracting from LZH compressed volumes is currently broken
It still assumes the file is memory mapped. 

```
void VolFile::ExtractFileLzh(std::size_t fileIndex, const std::string& pathOut)
{
	try
	{
		char* offset = (char*)m_IndexEntries[fileIndex].dataBlockOffset;
		int length = *(int*)(offset + 4) & 0x7FFFFFFF;
		offset += 8;

		HuffLZ decompressor(length, offset);
		const char *buffer = 0;

		FileStreamWriter fileStreamWriter(pathOut);

		do
		{
			buffer = decompressor.GetInternalBuffer(&length);
			fileStreamWriter.Write(buffer, length);
		} while (length);
	}
	catch (std::exception e)
	{
		throw std::runtime_error("Error attempting to extract LZH compressed file " + pathOut + ". Internal Error Message: " + e.what());
	}
}
```

DanRStevens:
This is untested, but does compile. It's a minimal change to update the
code to more recent conventions, hopefully fixing the break.

This hopefully fixes #85.

### DIFF
--- a/Archives/AdaptHuffTree.cpp
+++ b/Archives/AdaptHuffTree.cpp
@@ -80,7 +80,7 @@ namespace Archives
 			throw std::runtime_error("Index out of range");
 		}
 
-		// Return weather or not this is a terminal node
+		// Return whether or not this is a terminal node
 		return m_Data[nodeIndex] >= m_NumNodes;
 	}
 
@@ -93,7 +93,8 @@ namespace Archives
 			throw std::runtime_error("Index out of range");
 		}
 
-		// Return weather or not this is a terminal node
+		// Return data stored in node translated back to normal form
+		// Note: This assumes the node is a terminal node
 		return m_Data[nodeIndex] - m_NumNodes;
 	}
 
--- a/Archives/BitStream.cpp
+++ b/Archives/BitStream.cpp
@@ -15,7 +15,7 @@ namespace Archives
 		m_WriteBuff = 0;
 	}
 
-	BitStream::BitStream(int bufferSize, char *buffer)
+	BitStream::BitStream(int bufferSize, void *buffer)
 	{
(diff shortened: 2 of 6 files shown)


Brett208:
I pulled all the 1.2.0.5 volume files and extracted all files from them. Sheets.vol is the only one that appears to be compressed (which I think we have discussed before). 

This code does decompress them. The file appears decompressed properly until the last line(s) where there is garbage in the file. For the non tech trees, it Is random characters. For the tech trees it appears to be a bunch of extra empty lines.

I'm unsure if this is due to a problem with the file pass of the decompression? Could be a different recent change though as well.

I'll post the files in the forum for review.

-Brett

Brett208:
Spent a bit of time trying to debug this. Didn't make any progress.

I did notice the following weather comment typo in AdaptHuffTree.cpp

```
// Return weather or not this is a terminal node
return m_Data[nodeIndex] >= m_NumNodes;
```

DanRStevens:
I think this is normal expected behaviour. The game actually decompresses garbage at the end of those files, which results in parse errors, which halts processing of those files. I don't think the game slices the output of the decompressors. Also worth mentioning, as it's related, but the input streams are byte aligned, while the compression codes are only bit aligned. Hence there may be padding bits in the bytes at the end of a compressed stream. These extra bits do odd things when decompressing, resulting in extra output bytes, which are not properly sliced off. I think one of the reasons why they are not sliced off is the game stores the compressed size in the index table, not the uncompressed size, so it really has no idea how big the decompressed stream should be.

I think this is discussed on the forum somewhere, though I don't remember which thread off the top of my head.

PR #92, opened 2018-06-24
Linux makefile library settings

DanRStevens:
Makefile updates to produce a static library.

Previously the makefile was attempting to produce an executable file, which fails at the link stage due to a missing "main" method. Obviously that was silly, and so the makefile was updated to produce a library instead.

I also included a change for the compiler default. We don't need to use MinGW anymore now that the window.h dependency has been removed. Additionally, MinGW fails on the experimental/filesystem include. At this point g++ and clang are now better options.

### DIFF
--- a/makefile
+++ b/makefile
@@ -2,7 +2,7 @@
 # Set compiler default to mingw
 # Can still override from command line or environment variables
 ifeq ($(origin CXX),default)
-	CXX := i686-w64-mingw32-g++
+	CXX := clang-6.0
 endif
 
 SRCDIR := .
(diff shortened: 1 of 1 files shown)


DanRStevens:
@Brett208 I guess there isn't a whole lot for you to review, given this is a makefile only change. Might still be fun to look at though.

If you really want to go out of your way, you could probably install make and clang on Windows :stuck_out_tongue_winking_eye: 
(No, I don't expect you to do this)

Brett208:
Yeah, looks good to me. :)

PR #93, opened 2018-06-25
Use fixed width variables in clm file

### LINKED ISSUE #82: Switch to fixed width types on stream related variables in clmFile
The following changes should be made to ClmFile to improve stream read/write compatibility with different compilers.

Set IndexEntry::dataOffset and dataLength to uint32_t
```
struct IndexEntry
{
	std::array<char, 8> fileName;
	unsigned int dataOffset;
	unsigned int dataLength;

	std::string GetFileName() const;
};
```

int ClmFile::FindChunk should return a uint32_t and throw an error if Chunk is not found instead of returning -1

In ClmFile::ReadAlWaveHeaders, fmtChunkLength & dataChunkLength should be uint32_t instead of int.
```
// Read the format tag
int fmtChunkLength = FindChunk(tagFMT_, *filesToPackReaders[i]);
. . .
// Find the start of the data
int dataChunkLength = FindChunk(tagDATA, *filesToPackReaders[i]);
```

Variable CLM_WRITE_SIZE can be removed (no longer referenced)
```const uint32_t CLM_WRITE_SIZE = 0x00020000;```

### PR DESCRIPTION
**Pull request by Brett208**
Remove Unused Variabe CLM_WRITE_SIZE

Use Fixed Width Variables in stream related code in ClmFile … 
 - Switch functions ReadAllWaveHeaders & FindChunk to throw errors on failure instead of boolean return

Switch to catching exceptions by reference in VolFile.

Closes #82

### DIFF
--- a/Archives/ArchivePacker.cpp
+++ b/Archives/ArchivePacker.cpp
@@ -40,7 +40,7 @@ namespace Archives
 
 	void ArchivePacker::WriteFromStream(StreamWriter& streamWriter, StreamReader& streamReader, const uint64_t writeLength)
 	{
-		uint32_t numBytesToRead;
+		std::size_t numBytesToRead;
 		uint32_t offset = 0;
 		std::array<char, ARCHIVE_WRITE_SIZE> buffer;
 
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by Brett208** [comment]
Minor related question: Is there any reason to declare a std::exception as const when catching? EG

```
try {
}
catch (const std::exception&) {
}
```

There doesn't seem much gained from ensuring the exception class isn't modified in the catch block.

Brett

Brett208:
Thanks for the insights. I'll default to marking them as const for now. On the odd chance that one should be edited, it can be explicitly unmarked as const.

PR #94, opened 2018-06-25
Fix stream initializer list warnings

### LINKED ISSUE #91: FileStreamReader member initializer list does not match actual initialization order
On Linux, there is a warning with g++ and clang for the member initializer list in FileStreamReader constructor:
```
g++ -MT build/obj/Streams/MemoryStreamWriter.o -MMD -MP -MF build/deps/Streams/MemoryStreamWriter.Td -std=c++14 -g -Wall -Wno-unknown-pragmas  -c -o build/obj/Streams/MemoryStreamWriter.o Streams/MemoryStreamWriter.cpp
In file included from Streams/FileStreamReader.cpp:1:0:
Streams/FileStreamReader.h: In constructor ‘FileStreamReader::FileStreamReader(std::__cxx11::string)’:
Streams/FileStreamReader.h:29:20: warning: ‘FileStreamReader::filename’ will be initialized after [-Wreorder]
  const std::string filename;
                    ^
Streams/FileStreamReader.h:28:16: warning:   ‘std::ifstream FileStreamReader::file’ [-Wreorder]
  std::ifstream file;
                ^
Streams/FileStreamReader.cpp:5:1: warning:   when initialized here [-Wreorder]
 FileStreamReader::FileStreamReader(std::string filename) : 
 ^
```

One fix is to swap the declaration order of the fields:
```
private:
	const std::string filename;
	std::ifstream file;
```

The other alternative is to swap the order in the member initializer list:
```
	//filename(filename), file(filename, std::ios::in | std::ios::binary)
	file(filename, std::ios::in | std::ios::binary), filename(filename)
```

Either way, the declaration order should then match the listed initialization order. Or in other words, the listed initialization order will then match the actual initialization order.

Any preference?

DanRStevens:
This fixes #91.

I went with `file` for the member variable name, since it's largely irrelevant if the member variable provides file access through a stream interface or not. Perhaps at some point in the future it might change to a random access file class.

### DIFF
--- a/Streams/FileStreamReader.cpp
+++ b/Streams/FileStreamReader.cpp
@@ -3,7 +3,8 @@
 
 // Defers calls to C++ standard library methods
 FileStreamReader::FileStreamReader(std::string filename) : 
-	filename(filename), file(filename, std::ios::in | std::ios::binary)
+	filename(filename),
+	file(filename, std::ios::in | std::ios::binary)
 {
 	if (!file.is_open()) {
 		throw std::runtime_error("Could not open file: " + filename);
--- a/Streams/FileStreamReader.h
+++ b/Streams/FileStreamReader.h
@@ -25,6 +25,6 @@ protected:
 	void ReadImplementation(void* buffer, std::size_t size) override;
 
 private:
-	std::ifstream file;
 	const std::string filename;
+	std::ifstream file;
 };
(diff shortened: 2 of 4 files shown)


Brett208:
This compiles fine without warning on MSVC. Ran through an extraction of all files from art.vol without issue.

Issue #29, opened 2018-05-14
Allow creating an archive without using a template file


Brett208:
Currently, a template volume (vol) or CLM file is required when creating a new archive.

Brett208:
Two possible solutions are:

- Make The Packing functions in ClmFile and VolFile static so they do not rely on opening an existing ClmFile before packing a new one. Need to check/experiment if C++ supports static virtual functions for this solution.

 - Delete the virtual class ArchiveFile. This would require separating ClmFile and VolFile into 2 separate classes, ClmPacker, ClmUnPacker. Could use ClmReader/Writer instead as well.

I think there are pros and cons to both solutions.

DanRStevens:
I suggest dropping the `virtual` and just making the functions `static`. Packing archives doesn't really need a long lived archive object with access to memory variables. Besides, when you create the archives, you know what type of archive you are creating.

The existing code is probably closer in spirit to the first option.

I think the pro of the first option is it'd be the simplest change to make, and it would address the exact issue with have concerning object instantiation and requiring template files.

The pro of the second option would be if you needed to support long lived archive write operations, like a database that supports interleaved reading and writing operations. The downside is much more complicated code and likely much higher memory usage. I don't think we need the benefits of the second option, nor should we pay the costs of it.

Brett208:
I plan on resolving this issue next. OP2Archive currently creates archives by calling CreateArchive off of ArchiveFile.

To make CreateArchive static, we will need to remove the virtual base function. I don't think this will be too big of a deal, but it will have to be dealt with in OP2Archive by selecting the right type of Archive to be created based on the type.

PR #97, opened 2018-06-26
Remove bool return from create archive

### LINKED ISSUE #44: Rewrite ArchivePacker::CreateArchive to be exception based instead of returning bool on success
Will involve refactoring VolFile::CreateArchive and ClmFile::CreateArchive. 

rewrite:
```virtual bool CreateArchive(std::string volumeFileName, std::vector<std::string> filesToPack) = 0;```

to read:
```virtual void CreateArchive(const std::string& volumeFileName, std::vector<std::string> filesToPack) = 0;```

Brett208:
In addition to the title:

 - Set argument volumeFileName in CreateArchive as a const reference
 - Replace bool return from CompareWaveFormats to throw an exception if WaveFormats do not match
 - Move code to select proper WaveFormat when Writing a ClmFile into a subfunction
 - Provide filenames of offending files when WaveFormat does not match 

Closes #44.

### DIFF
--- a/Archives/ArchivePacker.h
+++ b/Archives/ArchivePacker.h
@@ -15,10 +15,11 @@ namespace Archives
 
 		// Repack is used to replace the old volume with a new volume created from the
 		// files (in the current directory) that match the internal file names
-		virtual bool Repack() = 0;
+		virtual void Repack() = 0;
+
 		// Create volume is used to create a new volume file with the files specified in filesToPack.
 		// Returns true if successful and false otherwise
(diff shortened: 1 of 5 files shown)


Brett208:
This should be the last external API change that I have planned in the near future that affects OP2MapImager and OP2Archive. After this, I can catch those projects up with these changes so OP2MapImager and OP2Archive will compile on their master branch with an OP2Utility version that does not contain Windows.h.

Brett208:
Thanks for the review.

I forgot, I also want to resolve issue #29 before starting in on the other projects as it will allow removing the template file requirement.

PR #100, opened 2018-06-29
Standardize for loops

### LINKED ISSUE #96: Clang tautological-compare warning
The older stock install of Clang (but curiously not the newer version) flagged some code with a warning. It does rather look like an infinite loop due to the unsigned nature of `std::size_t`:
```
clang -MT build/obj/XFile.o -MMD -MP -MF build/deps/XFile.Td -std=c++14 -g -Wall -Wno-unknown-pragmas  -c -o build/obj/XFile.o XFile.cpp
XFile.cpp:86:47: warning: comparison of unsigned expression >= 0 is always true
      [-Wtautological-compare]
        for (std::size_t i = filenames.size() - 1; i >= 0; i--) 
                                                   ~ ^  ~
XFile.cpp:104:47: warning: comparison of unsigned expression >= 0 is always true
      [-Wtautological-compare]
        for (std::size_t i = filenames.size() - 1; i >= 0; --i)
                                                   ~ ^  ~
2 warnings generated.
```
Also of curious note is the inconsistency between `i--` and `--i`.

The functions in question are:
https://github.com/OutpostUniverse/OP2Utility/blob/41ec8f877e191e200998fdc9466cc63ad88258d2/XFile.cpp#L82-L121

The following StackOverflow may provide some guidance:
[Reverse iteration with an unsigned loop variable](https://stackoverflow.com/questions/3623263/reverse-iteration-with-an-unsigned-loop-variable)

Brett208:
- In For loops, use --/++i instead of i--/++
 - Change function ArchiveUnpacker::ExtractAllFiles argument to be a string instead of a char*
 - Remove unnessary c_str() call on string inside function ArchiveUnpacker::ExtractAllFiles
 - Change function ArchiveUnpacker::GetInternalFileIndex's argument to be a string instead of a char*
 - Increase use of range based loops and auto keyword in range based loops
 - Fix bug in StringHelper::RemoveStrings that only removed the first string marked for removal
 - Standardize use of loop for (std::size_t i = filenames.size(); i-- > 0; ) when counting down from size to 0

Tested by Extracting a Vol, and clm. Extracted a compressed sheets.vol. REMOVED 2 files from a clm file.

Closes #96

### DIFF
--- a/StringHelper.cpp
+++ b/StringHelper.cpp
@@ -18,21 +18,19 @@ std::string StringHelper::ConvertToUpper(const std::string& str)
 }
 
 // Returns a new vector with matching strings removed. Case insensitive.
-std::vector<std::string> StringHelper::RemoveStrings(const std::vector<std::string>& stringsToSearch, const std::vector<std::string>& stringsToRemove)
+std::vector<std::string> StringHelper::RemoveStrings(std::vector<std::string> stringsToSearch, const std::vector<std::string>& stringsForRemoval)
 {
-	std::vector<std::string> stringsToReturn(stringsToSearch);
-
-	// i will wrap around to SIZE_MAX  when loop is completed
-	for (std::size_t i = stringsToSearch.size() - 1; i < stringsToSearch.size(); ++i) {
-		for (const auto& stringToRemove : stringsToRemove) {
+	// Loop starts at index size - 1 and ends after index 0 executes
+	for (std::size_t i = stringsToSearch.size(); i-- > 0; ) {
+		for (const auto& stringToRemove : stringsForRemoval) {
 			if (CheckIfStringsAreEqual(stringsToSearch[i], stringToRemove)) {
-				stringsToReturn.erase(stringsToReturn.begin() + i);
+				stringsToSearch.erase(stringsToSearch.begin() + i);
 				break;
 			}
 		}
 	}
 
-	return stringsToReturn;
+	return stringsToSearch;
 }
 
 bool StringHelper::CheckIfStringsAreEqual(const std::string& string1, const std::string& string2)
--- a/XFile.cpp
+++ b/XFile.cpp
@@ -83,7 +83,8 @@ std::vector<std::string> XFile::GetFilesFromDirectory(const std::string& directo
 {
 	std::vector<std::string> filenames = GetFilesFromDirectory(directory);
 
-	for (std::size_t i = filenames.size() - 1; i >= 0; i--) 
+	// Loop starts at index size - 1 and ends after index 0 executes
+	for (std::size_t i = filenames.size(); i-- > 0; )
 	{
 		if (!std::regex_search(filenames[i], filenameRegex)) {
 			filenames.erase(filenames.begin() + i);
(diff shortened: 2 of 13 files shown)


StringHelper.cpp
```diff
@@ -18,21 +18,19 @@ std::string StringHelper::ConvertToUpper(const std::string& str)
 }
 
 // Returns a new vector with matching strings removed. Case insensitive.
-std::vector<std::string> StringHelper::RemoveStrings(const std::vector<std::string>& stringsToSearch, const std::vector<std::string>& stringsToRemove)
+std::vector<std::string> StringHelper::RemoveStrings(std::vector<std::string> stringsToSearch, const std::vector<std::string>& stringsForRemoval)
```
DanRStevens: These changes look good. I had to think about this one a bit. I noticed there is now an implicit vector copy done in the parameter list while calling the function, rather than an explicit one within the function. That nicely cuts down on code. The loop counting changed, looks like you were fixing a bug where the old code started at the end of the vector and used `++i`, while the new code counts down according to the recently discussed idiom.

I believe this code avoids iterator invalidation problems after calling [`erase`](https://en.cppreference.com/w/cpp/container/vector/erase). After `erase`, any iterators at or beyond the point of erasure are invalidated.

I wonder if this code should make use of [`remove_if`](https://en.cppreference.com/w/cpp/algorithm/remove). It would probably result in less copying in the case where multiple values are removed. That might be a subject for another branch though.

Brett208: Yeah, the ++ in the loop was the bug allowing for only sing file removal from archives. I had only been testing removal of single files in the past. When reviewing the code I noticed it looked odd to be a ++ and tested better.

Generally agree this passage is difficult to read and could be improved on.

XFile.cpp
```diff
@@ -83,7 +83,8 @@ std::vector<std::string> XFile::GetFilesFromDirectory(const std::string& directo
 {
 	std::vector<std::string> filenames = GetFilesFromDirectory(directory);
 
-	for (std::size_t i = filenames.size() - 1; i >= 0; i--) 
+	// Loop starts at index size - 1 and ends after index 0 executes
+	for (std::size_t i = filenames.size(); i-- > 0; )
```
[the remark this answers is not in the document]
Brett208: Oh yeah, I forgot about this. Thanks for pointing it out. I'll fix before committing.

Brett208:
I see we are building twice using AppVeyor. It looks like they are both MSVC builds. The git fetch and clone commands seem a bit different on the 2 builds. What is the difference between them?

Thanks,
Brett

Brett208:
Merge Standardizing For Loops

DanRStevens:
Hmm, one is labelled as "branch", and the other as "pr" (pull request). The branch run seems to test only the exact branch/commit that was pushed. The pull request test seems to test master with the pull request merged in. Interesting. Makes sense though, at least assuming it's able to merge automatically, which you do kind of want when it comes time to merge a pull request in.

Brett208:
Okay, that makes sense. You want to know both if the pull request compiles by itself and if the master will compile if the pull request is merged.

Thanks,
Brett

PR #101, opened 2018-06-30
Reduce charStar usage

Brett208:
- Change function arguments from char* to string where appropriate
 - Remove unnecessary string.c_str calls

Some of the string.c_str calls were leftover from bad refactoring and some removals were allowed by changing out the function arguments.

This should implement string everywhere in the codebase where it would be preferred over char*. That took a long time to reach.

### DIFF
--- a/Archives/ClmFile.cpp
+++ b/Archives/ClmFile.cpp
@@ -292,7 +292,7 @@ namespace Archives
 		for (std::size_t i = 0; i < internalNames.size(); ++i)
 		{
 			// Copy the filename into the entry
-			std::memcpy(indexEntries[i].fileName.data(), internalNames[i].data(), sizeof(IndexEntry::fileName));
+			std::strncpy(indexEntries[i].fileName.data(), internalNames[i].data(), sizeof(IndexEntry::fileName));
 
 			// Set the offset of the file
 			indexEntries[i].dataOffset = offset;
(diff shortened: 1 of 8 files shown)


Brett208:
Nice catch in the FileStreamWriter constructor. For some reason I was thinking that conversion was necessary.

Good call on cleaning up at ClmFile.cpp:294. I've been wanting to remove that c style strncpy call for a while now but hadn't gotten around to it.

-Brett

Archives/ClmFile.cpp
```diff
@@ -291,7 +292,7 @@ namespace Archives
 		for (std::size_t i = 0; i < internalNames.size(); ++i)
 		{
 			// Copy the filename into the entry
-			strncpy((char*)&indexEntries[i].fileName, internalNames[i].c_str(), 8);
+			std::memcpy(indexEntries[i].fileName.data(), internalNames[i].data(), sizeof(IndexEntry::fileName));
```
DanRStevens: I think I might still use [`stdncpy`](https://en.cppreference.com/w/cpp/string/byte/strncpy) here. It will not attempt to read past the end of the input string. Additionally, it null pads the buffer up to the stated size. That means no uninitialized data laying around in the index table.

DanRStevens:
I looked briefly into using [`std::basic_string::copy`](https://en.cppreference.com/w/cpp/string/basic_string/copy), but it doesn't do the null padding, which is quite handy here.

Archives/ClmFile.cpp
DanRStevens: For the record, I do rather like the use of `sizeof` here. That was a nice touch.

PR #103, opened 2018-06-30
Add CircleCI config.yml

DanRStevens:
This adds a Linux Continuous Integration environment.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,9 @@
+version: 2
+jobs:
+  build:
+    docker:
+      - image: gcc:8
+    steps:
+      - checkout
+      - run: make CXX=g++
+


Brett208:
This seems like a good idea. Anything I need to think about when submitting changes that might require changes to config.yml?

DanRStevens:
Not that I can think of. Though it can be hard to get these config files right. Worse, you have to push the commit for the build system to see it and pick up the changes. For sure make the changes on a branch. I was using force push to edit and fix so I wouldn't have a string of 10 useless broken commits on my branch.

The output is really informational only, so can be ignored. It does get annoying though if it constantly shows up broken due to a bad config.

Additionally, this is using an older version of g++. It spits out some spurious warnings for `enum` values stored in bit fields, claiming they are too small, even though the bit fields are wide enough to hold all `enum` values. The warnings were deemed a bug in the compiler, and fixed in a future version.

I might still do a bit of work on the CI setup. Perhaps I'll change the base image to get a newer compiler, or add a build using clang. Likely I'll try setting up TravisCI as well. If it gets to be too much, we can remove whichever ones we don't like.

Should be easy to turn off using the web interface for CircleCI. If you just delete the `.circleci/config.yml` I think it might still try to run, but complain about the missing file and fail with an error. Feel free to play around with it. Maybe try making a few adjustments and commits on this branch and in the web interface if you want. You can always reset the branch and force push to get it back to the current state when you're done toying with the system.

DanRStevens:
I just realized you're probably waiting for me to merge this in, while I'm waiting for you to merge this in.

I'm going to go ahead and merge this. If you want to test changing settings, or deleting the settings file, you can always do so in a new branch. You can also try deleting/re-creating the project on the CircleCI website if you want.

Brett208:
I think for now I'll just trust you to get the settings right on CircleCI unless I start producing failing code. 

Thanks for setting this up.

-Brett

PR #104, opened 2018-07-01
Allow Calling Slice on SeekableStreamReaders

Brett208:
I had to wrap the returned stream slice in a unique_ptr.

I'm a bit split about the addition here. What do others think?

MemoryStreamReader's Slice function is untested, so worth taking a closer look.

Thanks,
Brett

### DIFF
--- a/Archives/ClmFile.cpp
+++ b/Archives/ClmFile.cpp
@@ -1,4 +1,5 @@
 #include "ClmFile.h"
+#include "../Streams/FileSliceReader.h"
 #include "../XFile.h"
 #include <stdexcept>
 #include <algorithm>
@@ -80,11 +81,11 @@ namespace Archives
 
 			waveFileWriter.Write(header);
 
-			std::unique_ptr<SeekableStreamReader> reader = clmFileReader.Slice(
+			FileSliceReader reader = clmFileReader.Slice(
 				indexEntries[fileIndex].dataOffset, 
 				indexEntries[fileIndex].dataLength);
 
-			ArchiveUnpacker::WriteFromStream(waveFileWriter, *reader, reader->Length());
+			ArchiveUnpacker::WriteFromStream(waveFileWriter, reader, reader.Length());
 		}
 		catch (const std::exception& e)
 		{
--- a/Archives/ClmFile.cpp
+++ b/Archives/ClmFile.cpp
@@ -80,10 +80,11 @@ namespace Archives
 
 			waveFileWriter.Write(header);
 
-			// Seek to the beginning of the file data (in the .clm file)
-			clmFileReader.Seek(indexEntries[fileIndex].dataOffset);
+			std::unique_ptr<SeekableStreamReader> reader = clmFileReader.Slice(
+				indexEntries[fileIndex].dataOffset, 
+				indexEntries[fileIndex].dataLength);
 
-			ArchiveUnpacker::WriteFromStream(waveFileWriter, clmFileReader, indexEntries[fileIndex].dataLength);
+			ArchiveUnpacker::WriteFromStream(waveFileWriter, *reader, reader->Length());
 		}
 		catch (const std::exception& e)
 		{
--- a/Archives/VolFile.cpp
+++ b/Archives/VolFile.cpp
@@ -1,4 +1,5 @@
 #include "VolFile.h"
+#include "../Streams/FileSliceReader.h"
 #include "../XFile.h"
 #include <stdexcept>
 #include <algorithm>
@@ -86,7 +87,7 @@ namespace Archives
 	{
 		SectionHeader sectionHeader = GetSectionHeader(fileIndex);
 
-		return archiveFileReader.Slice(archiveFileReader.Position(), static_cast<uint64_t>(sectionHeader.length));
+		return std::make_unique<FileSliceReader>(archiveFileReader.Slice(archiveFileReader.Position(), static_cast<uint64_t>(sectionHeader.length)));
 	}
 
 	VolFile::SectionHeader VolFile::GetSectionHeader(int index)
@@ -132,9 +133,9 @@ namespace Archives
 			// Calling GetSectionHeader moves the streamReader's position to just past the SectionHeader
 			SectionHeader sectionHeader = GetSectionHeader(fileIndex);
 
-			std::unique_ptr<SeekableStreamReader> slice = archiveFileReader.Slice(sectionHeader.length);
+			FileSliceReader slice = archiveFileReader.Slice(sectionHeader.length);
 
-			ArchiveUnpacker::WriteFromStream(pathOut, *slice, slice->Length());
+			ArchiveUnpacker::WriteFromStream(pathOut, slice, slice.Length());
 		}
 		catch (const std::exception& e)
 		{
--- a/Archives/VolFile.cpp
+++ b/Archives/VolFile.cpp
@@ -1,5 +1,4 @@
 #include "VolFile.h"
-#include "../Streams/FileSliceReader.h"
 #include "../XFile.h"
 #include <stdexcept>
 #include <algorithm>
@@ -87,8 +86,7 @@ namespace Archives
 	{
 		SectionHeader sectionHeader = GetSectionHeader(fileIndex);
 
-		return std::make_unique<FileSliceReader>(m_ArchiveFileName, 
-			archiveFileReader.Position(), static_cast<uint64_t>(sectionHeader.length));
+		return archiveFileReader.Slice(archiveFileReader.Position(), static_cast<uint64_t>(sectionHeader.length));
 	}
 
 	VolFile::SectionHeader VolFile::GetSectionHeader(int index)
@@ -134,7 +132,9 @@ namespace Archives
 			// Calling GetSectionHeader moves the streamReader's position to just past the SectionHeader
 			SectionHeader sectionHeader = GetSectionHeader(fileIndex);
 
-			ArchiveUnpacker::WriteFromStream(pathOut, archiveFileReader, sectionHeader.length);
+			std::unique_ptr<SeekableStreamReader> slice = archiveFileReader.Slice(sectionHeader.length);
+
+			ArchiveUnpacker::WriteFromStream(pathOut, *slice, slice->Length());
 		}
 		catch (const std::exception& e)
 		{
--- a/Streams/FileSliceReader.cpp
+++ b/Streams/FileSliceReader.cpp
@@ -58,3 +58,24 @@ void FileSliceReader::SeekRelative(int64_t offset)
 
 	fileStreamReader.SeekRelative(offset);
 }
+
+std::unique_ptr<SeekableStreamReader> FileSliceReader::Slice(uint64_t sliceLength) 
+{
+	auto slice = Slice(startingOffset + Position(), sliceLength);
+
+	// Wait until slice is successfully created before seeking forward.
+	SeekRelative(sliceLength);
+
+	return slice;
+}
+
+std::unique_ptr<SeekableStreamReader> FileSliceReader::Slice(uint64_t sliceStartPosition, uint64_t sliceLength) 
+{
+	if (sliceStartPosition + sliceLength > Length() ||
+		sliceStartPosition + sliceLength < sliceStartPosition) // Check if length wraps past max size of uint64_t
+	{
+		throw std::runtime_error("Unable to create a slice of an existing file stream slice. Requested slice is outside bounds of underlying stream slice.");
+	}
+
+	return std::make_unique<FileSliceReader>(GetFilename(), sliceStartPosition, sliceLength);
+}
--- a/Streams/FileSliceReader.cpp
+++ b/Streams/FileSliceReader.cpp
@@ -69,9 +69,9 @@ std::unique_ptr<SeekableStreamReader> FileSliceReader::Slice(uint64_t sliceLengt
 	return slice;
 }
 
-std::unique_ptr<SeekableStreamReader> FileSliceReader::Slice(uint64_t sliceStartPosition, uint64_t sliceLength) 
+std::unique_ptr<SeekableStreamReader> FileSliceReader::Slice(uint64_t sliceStartPosition, uint64_t sliceLength) const
 {
-	if (sliceStartPosition + sliceLength > Length() ||
+	if (sliceStartPosition + sliceLength > this->sliceLength ||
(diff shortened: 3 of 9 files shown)


Streams/FileSliceReader.cpp
```diff
@@ -58,3 +58,24 @@ void FileSliceReader::SeekRelative(int64_t offset)
 
 	fileStreamReader.SeekRelative(offset);
 }
+
+std::unique_ptr<SeekableStreamReader> FileSliceReader::Slice(uint64_t sliceLength) 
+{
+	auto slice = Slice(startingOffset + Position(), sliceLength);
+
+	// Wait until slice is successfully created before seeking forward.
```
DanRStevens: Very good observation. This is important to exception handling safety.

Streams/MemoryStreamReader.cpp
```diff
@@ -48,3 +48,28 @@ void MemoryStreamReader::SeekRelative(int64_t offset)
...
+		throw std::runtime_error("Slice starting position and Slice length for creating a new memory stream slice must be smaller values.");
+	}
+
+	if (sliceStartPosition + sliceLength > Length() ||
+		sliceStartPosition + sliceLength < sliceStartPosition) // Check if length wraps past max size of uint64_t
+	{ 
+		throw std::runtime_error("Unable to create a slice of memory stream. Requested slice is outside bounds of underlying stream.");
+	}
+
+	return std::make_unique<MemoryStreamReader>((void*)&streamBuffer[sliceStartPosition], static_cast<std::size_t>(sliceLength));
```
DanRStevens: There's an old C-style cast here which could be changed to a `static_cast`.

Brett208: I forgot to change this in the most recent push, but I'll clean it up before a merge. Thanks for pointing out.

Brett208:
I tried implementing this again by removing the Slice function from the SeekableStreamReader interface.

Ran into a new problem though.

Error C2280 'FileSliceReader::FileSliceReader(const FileSliceReader &)': attempting to reference a deleted function OP2Utility c:\ . . . \op2utility\streams\filestreamreader.cpp 45

Likely caused because the FileSliceReader does not have a copy constructor defined due to the contained underlying std::ifstream.

I don't think we want to let FileSliceReaders or FileStreamReaders be created by copy construction. I think this means we must return it by some sort of reference, whether that is a raw pointer or some sort of smart pointer?

Maybe I'm missing something though.

---

The MemoryStreamReader's version of Slice hasn't been updated yet. Also, the comments on the FileStreamReader slice functions need to be applied to all the Slice functions. I didn't see any reason to update these until the error is fixed.

I made one version of Slice a const function. I had to use the private length variable in the associated streams to accomplish this. It makes the Slice function a bit more brittle than calling the public Length function directly. Not sure if the tradeoff is worth it as I'm not sure when you would be passing around a const stream. Maybe best to remove this.

-Brett

Archives/ClmFile.cpp
```diff
@@ -80,11 +81,11 @@ namespace Archives
 
 			waveFileWriter.Write(header);
 
-			std::unique_ptr<SeekableStreamReader> reader = clmFileReader.Slice(
+			FileSliceReader reader = clmFileReader.Slice(
```
DanRStevens: We should use `auto` here to protect against return type changes in the FileStreamReader code, particularly since the API is not yet stable.

Archives/VolFile.cpp
```diff
@@ -132,9 +133,9 @@ namespace Archives
 			// Calling GetSectionHeader moves the streamReader's position to just past the SectionHeader
 			SectionHeader sectionHeader = GetSectionHeader(fileIndex);
 
-			std::unique_ptr<SeekableStreamReader> slice = archiveFileReader.Slice(sectionHeader.length);
+			FileSliceReader slice = archiveFileReader.Slice(sectionHeader.length);
```
DanRStevens: Ditto.

Streams/FileSliceReader.h
```diff
@@ -17,8 +17,8 @@ class FileSliceReader : public SeekableStreamReader
 	void Seek(uint64_t position) override;
 	void SeekRelative(int64_t offset) override;
 
-	std::unique_ptr<SeekableStreamReader> Slice(uint64_t sliceLength) override;
-	std::unique_ptr<SeekableStreamReader> Slice(uint64_t sliceStartPosition, uint64_t sliceLength) override;
+	std::unique_ptr<SeekableStreamReader> Slice(uint64_t sliceLength);
+	std::unique_ptr<SeekableStreamReader> Slice(uint64_t sliceStartPosition, uint64_t sliceLength) const;
```
DanRStevens: The return type here now seems inconsistent, and not very specific.

Streams/FileStreamReader.cpp
```diff
@@ -40,16 +40,16 @@ void FileStreamReader::SeekRelative(int64_t offset) {
...
 }
 
-std::unique_ptr<SeekableStreamReader> FileStreamReader::Slice(uint64_t sliceLength) 
+FileSliceReader FileStreamReader::Slice(uint64_t sliceLength) 
 {
-	auto slice = Slice(Position(), sliceLength);
+	return Slice(Position(), sliceLength);
 
 	// Wait until slice is successfully created before seeking forward.
 	SeekRelative(sliceLength);
```
DanRStevens: This is currently dead code.

Streams/MemoryStreamReader.h
```diff
@@ -14,8 +14,8 @@ class MemoryStreamReader : public SeekableStreamReader {
 	void Seek(uint64_t position) override;
 	void SeekRelative(int64_t offset) override;
 
-	std::unique_ptr<SeekableStreamReader> Slice(uint64_t sliceLength) override;
-	std::unique_ptr<SeekableStreamReader> Slice(uint64_t sliceStartPosition, uint64_t sliceLength) override;
+	std::unique_ptr<SeekableStreamReader> Slice(uint64_t sliceLength);
+	std::unique_ptr<SeekableStreamReader> Slice(uint64_t sliceStartPosition, uint64_t sliceLength) const;
```
DanRStevens: This return type now seems inconsistent, and not very specific.

DanRStevens:
Actually, it's perfectly reasonable to commit broken code to a branch and request review, particularly for a hard problem like this. It's a convenient way to discuss a problem with code to demonstrate the problem.

It just happens that I was going through the emails in order, and that showed me the additions before the comment about how the additions were broken. I suppose if you'd commented first, and then pushed second, I probably would have seen them in order. Or maybe I just need to be more aware of the overall situation.

----

I'm leaning towards option 2. You're correct in that `ifstream` itself can't be copy constructed. Hence you'd need a `FileSliceReader` copy constructor that constructs a new `ifstream`, using the filename, rather than the old `ifstream`. This doesn't circumvent the lack of copy constructor on the `ifstream` per se, as it isn't being copied, rather a new object is being constructed with the same parameters as the old one.

As for deleting open files, Linux allows this, where the directory entry is removed, but the file contents are preserved on disk until the last file handle is closed. Hence a second open to an already opened file could potentially fail. I'm not worried about this, because I don't see us relying on this behaviour long term. Plus, the FileSliceReader already relies on this behavior as is.

Additionally, we just have to get the interface right at this point. The details of how it's implemented are subject to change. We could address this later on as an internal library change.

Finally, completing option 2 doesn't prevent us from later adding option 1 to the library as well. I don't think we need to, but we could. And yeah, the virtual interface would be somewhat limited in some respects, in that we'd have to choose between memory safety or covariance.

Brett208:
I went ahead and implemented the copy constructors. It is my first time writing one, but they are working.

I actually don't know how to remove the c-style void* cast in the MemoryStreamReader slice function. static_cast<void*> doesn't seem to work.

-Brett

Streams/MemoryStreamReader.cpp
```diff
@@ -71,5 +71,5 @@ std::unique_ptr<SeekableStreamReader> MemoryStreamReader::Slice(uint64_t sliceSt
 		throw std::runtime_error("Unable to create a slice of memory stream. Requested slice is outside bounds of underlying stream.");
 	}
 
-	return std::make_unique<MemoryStreamReader>((void*)&streamBuffer[sliceStartPosition], static_cast<std::size_t>(sliceLength));
+	return MemoryStreamReader((void*)(&streamBuffer[sliceStartPosition]), static_cast<std::size_t>(sliceLength));
```
DanRStevens: Ahh, `static_cast` can't cast away `const`. The `streamBuffer` points to a `const` buffer.

I believe the correct solution here, is to add the necessary `const` to the constructors, as this data is expected to be `const`. You may need to add a few as the `const`-ness spreads.

Streams/FileSliceReader.cpp
```diff
@@ -2,15 +2,30 @@
...
+
+FileSliceReader::FileSliceReader(const FileSliceReader& fileSliceReader) :  
+	fileStreamReader(fileSliceReader.GetFilename()), 
+	startingOffset(fileSliceReader.startingOffset), 
+	sliceLength(fileSliceReader.sliceLength)
+{
+	Initialize();
+}
+
+void FileSliceReader::Initialize()
```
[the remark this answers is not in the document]
Brett208: How about InitializeFileStream. Could be used on both the FileSliceReader and FileStreamReader's version of the function.

-Brett

Streams/MemoryStreamReader.cpp
Brett208: I just looked into const_cast as I didn't know it existed. We need to figure out how to change a const void const * buffer to a const char const * buffer, which I haven't figured out yet. 

I think switching the constructor to accept a const char const * buffer instead would fix the issue. Is there any reason to make the MemoryStreamReader's constructor a void* if the underlying buffer is going to be a char*?

-Brett

Streams/FileSliceReader.cpp
```diff
@@ -59,23 +74,23 @@ void FileSliceReader::SeekRelative(int64_t offset)
 	fileStreamReader.SeekRelative(offset);
 }
 
-std::unique_ptr<SeekableStreamReader> FileSliceReader::Slice(uint64_t sliceLength) 
+FileSliceReader FileSliceReader::Slice(uint64_t sliceLength) 
 {
-	auto slice = Slice(startingOffset + Position(), sliceLength);
+	FileSliceReader slice = Slice(startingOffset + Position(), sliceLength);
```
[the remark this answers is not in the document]
Brett208: I'm trying to decide when to use auto.

http://www.acodersjourney.com/2016/02/c-11-auto/

Streams/FileSliceReader.cpp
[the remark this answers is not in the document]
DanRStevens: The "Initialize" part sounds kind of vague. Also maybe a bit inaccurate now that I think about it, since the fields are already initialized at that point. They're just being error checked. Plus you have the somewhat unrelated operation of seeking.

The code of course makes perfect sense as written. I just feel the name doesn't help to understand what the method really does.

To be fair, the name isn't terrible. This is certainly something we could change post merge if a better name comes to mind.

Streams/MemoryStreamReader.cpp
DanRStevens: There's very little reason to use `const_cast`. I think it's included more because the operation was allowed by the old C-style cast. I suppose it might be necessary in rare circumstances, where `const` was not used properly in library code you need to interface with.

In terms of the conversion, assuming the const-ness doesn't change, you can just use `static_cast`:
```
static_cast<const char*>(someConstVoidPtr)
```

The constructor should accept a `const void*`, rather than a `void*`. That's the real problem. As long as the const-ness matches, you can `static_cast` it.

I think using `const void*` in the interface makes sense, since it allows easy cast free conversion from a pointer to any type. The memory could come from `new`, or from `malloc`, or allocated with a `vector`, over overlayed on top of a `struct`. The `MemoryStreamReader` shouldn't know and shouldn't care. It's just memory that needs to be used as an input buffer so it can be sliced and diced into something new.

Streams/FileSliceReader.cpp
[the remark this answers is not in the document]
DanRStevens: Lol, nice meme.

I found the comments at the end of the article particularly interesting. I think I agree with some of those comments more so than the negative points in the article. In particular, not needing to know or care what a type is in many cases, and the type is much less important than a good descriptive name that tells you what the value represents. Knowing what the value represents provides a good clue as to how a value should be used.

Compare the following:
```
auto str = SomeFunction();  // Name indicates type, but it otherwise vague
std::string a = SomeFunction();  // Type known, but name is vague
auto filename = SomeFunction();  // Name implies usage (which may suggest possible types)
```
I'd say the last is the most useful, and would be useful if the underlying type was a `std::string`, or if it was a `std::filesystem::path`. Further, chances are, surrounding code wouldn't care which of the two it was, leaving open the possibility of changing the function signature with little impact on the rest of the code.

Worth considering, is languages that don't have strict typing. In particular, scripting languages like Ruby or Python. Python is known for duck typing (if it quacks like a duck, it's a duck). You rely on good descriptive names, because there is no fixed type. C++ templates basically work with duck typing. Provided a type supports all the needed operations performed on it, the template can be instantiated for that type. What matters is how it's used, more so than the exact type.

Streams/MemoryStreamReader.cpp
Brett208: Okay, thanks. It actually never occurred to me that it would be legal to place a const inside a static_cast like 'static_cast<const char* const>'. Duh.

I was also writing const char const * buffer, which was incorrect. const is left associative with the exception of the first type. Fortunately MSVC starting squawking about it with a warning.

I just ran out of time, so I'll have to commit this later. Thanks for the help and sorry for being so dense.

Thank you for the compliment on the quality of the code being checked in.

Streams/FileSliceReader.cpp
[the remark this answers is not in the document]
Brett208: Good point. I think maybe I meant it more as InitializeClass when first writing. I'll just leave it alone for now though.

-Brett

Streams/MemoryStreamReader.cpp
DanRStevens: Fun fact, the following three lines are all valid code:
```
const char * var;
char const * var;
char * const var;
```

But what do they mean?

Turns out the first two are equivalent.

Brett208: Before yesterday I would have said the second option meant the third option and the third option wasn't valid.

Brett208:
I think this branch is completed and ready for merging.

-Brett

PR #106, opened 2018-07-01
Add TravisCI

DanRStevens:
Adds a TravisCI config file.

The initial commit did builds for both MacOS and Linux, but with outdated packages. The results didn't seem too usable. I've since updated the config file to install newer versions of Linux compilers. I'm leaving the MacOS builds out of scope for now. They were having issues with `#include <experimental/filesystem>`, likely due to outdated packages.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,15 @@
+
+language: cpp
+sudo: false
+os:
+  - linux
+  - osx
+compiler:
+  - gcc
+  - clang
+script:
+  # Use -k to continue after errors, ensuring full build log with all errors
+  - make -k
+notifications:
+  email: false
+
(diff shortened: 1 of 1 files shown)


Brett208:
It looks like TravisCI is mirroring appveyor for GCC and Clang (compiling the branch by itself and then compiling it by merging it into master).

I'm not sure what circleci is doing though?

-Brett

DanRStevens:
Looks like CircleCI has support to do the same, but isn't configured to do so. You can view the output of the section "Checkout code" of the build log. It has a number of environment variable guarded git checkout commands. Looks like it can support running a specific commit, or to merge an arbitrary branch into another arbitrary branch. Maybe I should play around with CircleCI some more, it seems promising. Looked to be easier to run with more up to date images and compilers, without having to install each run, which makes for faster test times. TravisCI is slow, with outdated images and compilers, though with a free OSX option.

PR #110, opened 2018-07-07
Allow StreamWriter to consume StreamReaders

Brett208:
**PLEASE READ BEFORE REVIEWING COMMITS** NOTE: This code will compile on MSVC, but will produce unexpected results. Currently the template function StreamWriter::Write will consume calls to StreamWriter::Write(SeekableStreamReader seekableStreamReader). There must be a way to prevent the template from taking priority here.

 - Move code for for consuming streams out of ArchivePacker & ArchiveUnpacker into StreamWriter.
 - Add a property to StreamWriter that allows getting/setting the pass size used when writing from a stream.

Assuming this code looks like a good idea (once the hidden function is resolved that is).

I could probably work a temp fix where within the template function StreamWriter::Write, we use dynamic_cast to check if the item could be cast as a SeekableStreamReader, and then call a private function to consume the SeekableStreamReader. I would see this as a temp fix until we can figure a better solution if nothing comes to mind.

Thanks,
Brett

### DIFF
--- a/Streams/StreamWriter.cpp
+++ b/Streams/StreamWriter.cpp
@@ -1,0 +1,41 @@
+#include "StreamWriter.h"
+#include "SeekableStreamReader.h"
+#include <vector>
+#include <stdexcept>
+
+void StreamWriter::Write(SeekableStreamReader& seekableStreamReader) {
+	Write(seekableStreamReader, seekableStreamReader.Length() - seekableStreamReader.Position());
+}
+
+void StreamWriter::Write(StreamReader& streamReader, uint64_t writeLength)
+{
+	std::size_t bytesToRead = 0;
+	uint64_t offset = 0;
+	std::vector<char> buffer;
+	buffer.resize(streamPassLength);
+
+	do
+	{
+		bytesToRead = buffer.size();
+
+		// Check if less than a streamPassLength of data remains for writing
+		if (offset + bytesToRead > writeLength) {
+			bytesToRead = static_cast<std::size_t>(writeLength - offset);
+		}
+
+		streamReader.Read(buffer.data(), bytesToRead);
+
+		offset += bytesToRead;
+
+		Write(buffer.data(), bytesToRead);
+	} while (bytesToRead); // End loop when BytesToRead/Write is equal to 0
+}
+
+void StreamWriter::SetStreamPassLength(std::size_t streamPassLength) 
+{
+	if (streamPassLength == 0) {
+		throw std::runtime_error("Stream pass length must be set to a value greater than 0.");
+	}
+
+	this->streamPassLength = streamPassLength;
+}
--- a/Streams/StreamWriter.h
+++ b/Streams/StreamWriter.h
@@ -1,22 +1,42 @@
 #pragma once
 
 #include <cstddef>
+#include <cstdint>
+
+class StreamReader;
+class SeekableStreamReader;
 
 class StreamWriter {
 public:
+	StreamWriter() : streamPassLength(0x00020000) { };
 	virtual ~StreamWriter() = default;
 
 	inline void Write(const void* buffer, std::size_t size) {
 		WriteImplementation(buffer, size);
 	}
 
+	// Writes a length of data from the provided stream starting at the provided stream's current position
+	void Write(StreamReader& streamReader, uint64_t length);
+
+	// Writes the remaining length of provided stream from the provided stream's current position
+	void Write(SeekableStreamReader& seekableStreamReader);
+
 	// Inline templated convenience methods, to easily write arbitrary data types
 	template<typename T> inline void Write(const T& object) {
 		WriteImplementation(&object, sizeof(object));
 	}
 
+	// StreamPassLength is the buffer size used when copying data from a stream reader into the writer.
+	// Changing the pass size may affect performance of copying in streams for writing.
+	void SetStreamPassLength(std::size_t streamPassLength);
+
+	inline std::size_t GetStreamPassLength() const { return streamPassLength; }
+
 protected:
 	// WriteImplementation is named differently from Write to prevent name hiding of the 
 	// Write template helpers in derived classes.
 	virtual void WriteImplementation(const void* buffer, std::size_t size) = 0;
+
+private:
+	std::size_t streamPassLength;
 };
--- a/Archives/ArchivePacker.cpp
+++ b/Archives/ArchivePacker.cpp
@@ -3,12 +3,9 @@
 #include "../StringHelper.h"
 #include <cstddef>
 #include <stdexcept>
-#include <array>
 
 namespace Archives
 {
-	const uint32_t ARCHIVE_WRITE_SIZE = 0x00020000;
-
 	ArchivePacker::ArchivePacker() { }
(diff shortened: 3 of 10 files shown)


Streams/StreamWriter.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+		streamReader.Read(buffer.data(), bytesToRead);
+
+		offset += bytesToRead;
+
+		Write(buffer.data(), bytesToRead);
+	} while (bytesToRead); // End loop when BytesToRead/Write is equal to 0
+}
+
+void StreamWriter::SetStreamPassLength(std::size_t streamPassLength)
```
DanRStevens: Err, what?

Streams/StreamWriter.h
```diff
@@ -1,22 +1,42 @@
...
 
 #include <cstddef>
+#include <cstdint>
+
+class StreamReader;
+class SeekableStreamReader;
 
 class StreamWriter {
 public:
+	StreamWriter() : streamPassLength(0x00020000) { };
```
DanRStevens: Ehh, not liking this

Streams/StreamWriter.h
```diff
@@ -1,22 +1,42 @@
...
+
+	inline std::size_t GetStreamPassLength() const { return streamPassLength; }
+
 protected:
 	// WriteImplementation is named differently from Write to prevent name hiding of the 
 	// Write template helpers in derived classes.
 	virtual void WriteImplementation(const void* buffer, std::size_t size) = 0;
+
+private:
+	std::size_t streamPassLength;
```
DanRStevens: This definitely kills the "interface" nature that was intended here.

At least with the template methods, they were implemented entirely in terms of the virtual methods, without touching any member variables. They might as well have been friend functions, except the usage syntax wouldn't have been quite as nice.

Streams/StreamWriter.cpp
Brett208: StreamPassLength is the size read from the stream on each pass for writing. The exact same concept as `const std::size_t ARCHIVE_WRITE_SIZE = 0x00020000;`. I thought the following comment explained it reasonably well. Guess not.

```// StreamPassLength is the buffer size used when copying data from a stream reader into the writer. 
// Changing the pass size may affect performance of copying in streams for writing. ```

**Comment by Brett208** [comment]
Oh, if you want to demo the function name hiding, download this branch and compile. Set a breakpoint in VolFile.cpp at line 138: `archiveWriter.Write(archiveFileReader.Slice(sectionHeader.length));`. 

Place art.vol in the root directory designated for debugging by the project file. Then run the code in debug mode with the command `EXTRACT art.vol well0002.bmp'. You should be able to single step into the function and see that you enter into the template function Write and attempt to write a FileSliceReader to disk as a simple object instead of calling `void Write(SeekableStreamReader& seekableStreamReader);` and writing the underlying slice to disk. 

Well, it does on MSVC anyways, no idea what Clang or others compilations would do.

-Brett

**Comment by DanRStevens** [comment]
Ahh, quite detailed instructions. It occurs to me now I'm not all that familiar with debugging tools on Linux. I suppose this is the sort of thing unit testing comes in handy for.

I wonder if it's because the object being passed is derived from `SeekableStreamReader`, rather than being one exactly itself, and so the template becomes a better match. If you `static_cast<SeekableStreamReader&>` the first parameter, would that make it call the right function?

----

Correct in that I planned to implement the `ReadPartial` method. That allows processing of streams without worrying about the exact format or length, as it just fills as much of a buffer as possible, rather than throw an exception. As such, it would make a very good base for copying one stream to another, without trying to format or size the data. Hence why it was discussed in 'Allow for automatic writing from an existing stream'.

Admittedly though, I have been a bit slow about it. What you started on here might make for a reasonable step in the right direction, and doesn't need to wait on the work I'd promised to do. Though I would like more minimal changes to the interface.

**Comment by Brett208** [comment]
Right after making the last post I was thinking this is a fairly difficult thing to setup in a debugging because you have to run OP2Archive with OP2Utility loaded into the debugger to hit the breakpoint. While Visual Studio has its own quirks and problems, it is definitely a very strong debugging tool if you have symbols or the source code compiling. 

I implemented your suggestion with the `static_cast`. MSVC was then able to choose the correct function and wrote a properly formatted BMP to disk. Good idea.

When I have some time, I'll check out your branch that limits the things consumed by the Write template function. I'm wondering if with that code added if MSVC would be smart enough to reject a child of SeekableStreamReader for basic copy and then pass it to the proper function.

We can leave this branch open until implementation of ReadPartial is completed for reference. I'm not tied to this implementation over another. I really just want to remove the code duplication occuring in ArchiveUnpacker and ArchivePacker by consolidating the ability to write from a stream to keep the code base cleaner.

-Brett

**Comment by DanRStevens** [comment]
Ok, good to know. The `static_cast` idea was meant more as a debugging aid. It shows the wrong function was picked because it was deemed a better match. I don't expect we'd want the `static_cast` in the final version of the code.

With that said, I believe the template `enable_if` guard will solve that particular problem. In particular, any class with `virtual` methods will not be `is_trivially_copyable`, hence the template will not expand for any `StreamReader` instances.

Next step should be to add the `enable_if` guard, and remove the `static_cast` and see what happens.

----

I believe with the overload issue dealt with, you could make suitable changes to finish this branch without waiting. Though yes, I do see more changes coming to this code after `ReadPartial` is complete, so I understand if you want to wait.

**Comment by Brett208** [comment]
I just checked, and yes using the new Write template with the enable_if_t call will properly call the correct function without the static_cast. Nice.

@DanRStevens, what strategy are you thinking for allowing the user to edit the buffer size when transferring data from a read stream to a write stream? A gathered you do not like a public Get, Set field.

I had originally thought about a function overload:

```void Write(SeekableStreamReader& seekableStreamReader, size_t passLength = DEFAULT_VALUE);
void Write(StreamReader& streamReader, uint64_t length, size_t passLength = DEFAULT_VALUE)```

This didn't work for me because the template write function would steal the function call if the DEFAULT_VALUE for passLength was used. Now that we have the include guard though, we could try using this function signature instead.

However, I still worry about the non-seekable StreamReader having too close of a signature to the SeekableStreamReader if the default passLength is used. It would be a size_t vs a uint64_t. If that makes sense.

-Brett

**Comment by DanRStevens** [comment]
I was thinking the default parameter would be a better way to specify the copy buffer size. That eliminates the need for the private variable, and the get/set methods.

Going back to the `ReadPartial` discussion, I think we only need a single method to support copying streams, and it should be the full stream. A partial stream can be copied by slicing before doing the copy. That should simplify the code, and the interface to it. It also reduces the surface area for function call stealing. The method I think we need is:
```
void Write(StreamReader& streamReader, std::size_t copyBufferSize = DefaultCopyBufferSize)
```

Your proposed methods are fine for now. The `ReadPartial` stuff hasn't been merged into master yet. If it's not ambiguous, we can go with that.

The potential for function call stealing is still a concern that we'd need to look at closely. I believe `uint64_t` and `std::size_t` are distinct types, as in, I think they're not simply a `typedef` of one to the other, so we should be alright there.

----

Looks like you have some merge conflicts. Have fun resolving those :wink:

DanRStevens:
Implemented in PR#127.

Issue #84, opened 2018-06-23
Allow for automatic writing from an existing stream


Brett208:
A common use case is reading from one stream and then writing that data immediately to another stream. We could standardize the process by including support in the stream library.

One solution would be to attach to following functions to StreamWriter. The first function could be omitted, which would allow adding only StreamReader as an include instead of SeekableStreamReader.

```
// Writes from the current position of the passed stream until the end of the stream.
// Takes an optional bufferSize to specify how large of chunks to read each pass.
void WriteFromStream(SeekableStreamReader& seekableStreamReader, size_t bufferSize =  0x00020000)
{
    WriteFromStream(seekableStreamReader, seekableStreamReader.Length() - seekableStreamReader.Position(), bufferSize);
}

// Writes from the current position of the passed stream through the specified writeLength.
// Takes an optional bufferSize to specify how large of chunks to read each pass.
void WriteFromStream(StreamReader& streamReader, uint64_t writeLength, size_t bufferSize = 0x00020000);
```

Another option might be to create a separate header that includes the functions standalone. It would require adding the StreamWriter as an input to the functions.

These 2 functions would both be used in ClmFile and VolFile, so I'm guessing their application would apply to other applications as well.

I just chose the default bufferSize because it is used in other parts of Archives and am not sure if a different default might be better.

DanRStevens:
I think we should support this in the Stream library. I definitely think such functionality will be generally re-usable.

The method, as given, does seem like a generic standalone method, so a separate standalone .cpp file makes sense. We could modify it and attach it to a specific stream class as well, with StreamWriter being the most obvious. It feels slightly weird to attach it to StreamWriter, but at the same time, I suspect that might be most convenient.

I'm wondering if "Consume" might be a good name for the method. That would sound natural in the context of a StreamWriter class. An alternative might be "Pipe", though more for the context of a StreamReader class.

----

If we get around to writing test code, we could do some actual speed tests and perhaps choose a buffer size that hits a sweet spot for performance versus memory use. For now though, I'm happy to just stick with whatever value was being used previously.

Brett208:
If attached to StreamWriter, it would read as streamWriter.ConsumeStream(streamReader, 100);

I like how ConsumeStream sounds better than WriteFromStream. I think the only downside is that this gives the connotation that the entire stream is consumed, where in reality it could only be a portion of the stream.

I haven't used a library that uses Pipe before, so this one doesn't seem as appealing to me. 

[...]

PR #112, opened 2018-07-11
Add partial read write

DanRStevens:
This should implement the `ReadPartial` method discussed in Issue #84. This could be used as a base to implement stream copying, without needing to know sizes up front.

I believe this has the right idea down, though this code is untested. We should really look into how we're going to write unit tests. I suppose one way to test this manually, without proper unit tests, is to write the stream copying code on top of it.

I didn't not write a corresponding `WritePartial` method. Do we even need one? I'm not sure there is as clear of a case for it.


Streams/FileSliceReader.cpp
```diff
@@ -42,6 +42,13 @@ void FileSliceReader::ReadImplementation(void* buffer, std::size_t size)
 	fileStreamReader.Read(buffer, size);
 }
 
+std::size_t FileSliceReader::ReadPartial(void* buffer, std::size_t size) {
+	auto bytesLeft = sliceLength - (fileStreamReader.Position() - startingOffset);
```
Brett208: Any reason not to make this line read:

```auto bytesLeft = sliceLength - Position();```

I think this change would read a bit easier.

DanRStevens:
Oh geeze duh on the change suggestion. I'll fix that.

I also prefer public methods at the top. Though I'm tempted to experiment a bit with this one. It's certainly a valid point that readability should come before micro optimizations. Though the micro optimization was not really the point. If it was, I'd have done that right from the beginning. Rather there's something about the organization that seems somehow right. The protected method was originally meant to be public. It's not public because of a name hiding workaround. However, it is still very much at the heart of the public API, as all the public methods are just short wrappers to add syntax sugar to calling the protected method. It's the protected method you must understand to use this API.

I guess I still haven't quite made up my mind on which way I want it. That's partly why I want to try it this way. It's easy enough to change later if we decide against it.

The documentation for the functions should be improved a bit. In particular, they could use more description about the exception behaviour. I'll try to add more there.

Interesting idea about the `noexcept`. I like it, except I'm not certain I want that strong of a guarantee. Is it true that all derived classes will never throw exceptions under any circumstance for that method? Or is it merely a promise to return a partial buffer rather than throw an exception if it hits the end of file before filling the buffer. I suppose I need to consider if there are any other valid reasons for such a method to throw.

I'll see what I can do to address the warnings.

DanRStevens:
I think I like the `noexcept` idea. If `ReadPartial` raised exceptions for other reasons, I think that would limit its usefulness.

Issue #68, opened 2018-06-18
Stream Read/Write helpers for complex types


DanRStevens:
It is desirable to be able to save complex data using a compact notation such as:
```
stream.Write(someComplexValue);
```
However, there is no safe and general way to do this for any data type. In particular, an object with a virtual function table pointer should not be saved or loaded this way. Such saved/loaded objects don't work with relocatable code, and could be a big security concern otherwise. In general, it's of questionable use for data that may contain pointers, or other resource handles. That may be through explicit use of pointers, or implicit use of pointers such as by virtual functions and virtual function tables.

It is also questionable to use a generic method for data types of non-fixed size, as that creates compatibility issues between code compiled for 32-bit and for 64-bit. It may be permissible to allow non-fixed sized types, but perhaps result in a warning.

For simple fixed sized data types, such as `int8_t`, `int16_t`, `int32_t`, `int64_t`, and the corresponding unsigned types, it should be fine. Similarly, combining elements into a struct or array should also be fine.

The standard references things like [Aggregate](https://en.cppreference.com/w/cpp/language/aggregate_initialization) types, [Standard Layout](https://en.cppreference.com/w/cpp/named_req/StandardLayoutType) types, and [POD](https://en.cppreference.com/w/cpp/named_req/PODType) types. I think we want something similar, though the exact definition may vary from anything the standard specifies.

We might be able to piggyback off some of the type traits defined by the standard. For example: [`is_aggregate`](https://en.cppreference.com/w/cpp/types/is_aggregate), [`is_standard_layout`](https://en.cppreference.com/w/cpp/types/is_standard_layout), [`is_pod`](https://en.cppreference.com/w/cpp/types/is_pod), [`is_trivial`](https://en.cppreference.com/w/cpp/types/is_trivial), [`is_scalar`](https://en.cppreference.com/w/cpp/types/is_scalar), [`is_pointer`](https://en.cppreference.com/w/cpp/types/is_pointer). These can be combined with [**`enable_if`**](http://en.cppreference.com/w/cpp/types/enable_if).

This is getting to the limits of my knowledge of templates, but I believe I've seen code that variously used either `using` or `enable_if` to limit template instantiation to safe types. That should prevent unexpected behaviour from undesired or unintended template expansion.

----

[...]

PR #113, opened 2018-07-11
Add `Read`/`Write` template guard

DanRStevens:
This should prevent inappropriate expansions of the `Read`/`Write` template methods.

This is related to Issue #68. Still left to do is to provide helpers for common complex types, such as `std::vector`.

### DIFF
--- a/Streams/StreamReader.h
+++ b/Streams/StreamReader.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <cstddef>
+#include <type_traits>
 
 class StreamReader {
 public:
@@ -11,7 +12,8 @@ public:
 	}
 
 	// Inline templated convenience methods, to easily read arbitrary data types
-	template<typename T> inline void Read(T& object) {
+	template<typename T>
+	inline std::enable_if_t<std::is_trivially_copyable<T>::value> Read(T& object) {
 		ReadImplementation(&object, sizeof(object));
 	}
 
--- a/Streams/StreamWriter.h
+++ b/Streams/StreamWriter.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <cstddef>
+#include <type_traits>
 
 class StreamWriter {
 public:
@@ -11,7 +12,8 @@ public:
 	}
 
 	// Inline templated convenience methods, to easily write arbitrary data types
-	template<typename T> inline void Write(const T& object) {
+	template<typename T>
+	inline std::enable_if_t<std::is_trivially_copyable<T>::value> Write(const T& object) {
 		WriteImplementation(&object, sizeof(object));
 	}
 


DanRStevens:
Good to hear it's catching bad uses of the template with `vector` and `SeekableStreamReader`.

As for the other data types, you are correct, a `char*`, `vector.data()`, and `const void*` are all `is_trivially_copyable` types, with size equal to that of a pointer. Effectively, you are saving the pointer, not the pointed to data. Hence this is expected behaviour, given the code.

You're right though, about this being potentially confusing to new programmers. The above is why I wanted to explicitly disallow pointers. Although pointers are trivially copyable, they are generally useless in a file, and it's probably an error to write a pointer to a file. Hence, the earlier discussion about wanting something POD-like, but without pointers. That should now be trivially-copyable-like, but without pointers.

For now, I'm going to consider this close enough. At the very least, it's an improvement. Perhaps we can make some modifications in the future if we learn how `is_trivially_copyable` is implemented, and find a way to do similar, but exclude pointers and references.

----

Glad to see you noticed the `void` change. The reason why is interesting. It's how the `enable_if` template works, using SFINAE (Substitution Failure Is Not An Error).

If the parameter to `enable_if` is `true`, the expression `std::enable_if_t<std::is_trivially_copyable<T>::value>` evaluates to `void`. If the parameter to `enable_if` is `false`, the expression `std::enable_if_t<std::is_trivially_copyable<T>::value>` evaluates to `` (the empty string). As the empty string produces invalid code (a function with no return type), it's considered a substitution failure, and so removes the template from consideration for overload resolution. In particular, this sort of syntax error is not a failure, and so the compiler does not abort compiling.

If you need a type other than `void`, you can specific it after the test expressions. For example, here the expression would evaluate to `int` if the test expression is true, and the empty string otherwise:
```
std::enable_if_t<std::is_trivially_copyable<T>::value, int>
```

DanRStevens:
There is now a merge conflict here, due to the merging of the branch that normalized CRLF to master.

Interestingly, the majority of all my files appear to have LF endings.

I'd like to merge this without upsetting the line ending normalization, though I'm a little uncertain what is going on. This may be a tomorrow problem.

Brett208:
Well, I'm guessing the 11 files that I fixed in the include branch merge should be giving you trouble. If you notice more than those 11 files, then something else is going on. 

I'm going to try to submit moving the src code soon. Maybe the best call is just to reapply this branch and the ReadPartial after that branch is merged (or base them off that branch).

Sorry if it was my fault in the first place on all the line ending shenanigans.

-Brett

DanRStevens:
After investigating the line ending changes, it appears they are not normalized across the project. As such I'm not going to worry about it too much.

I got a bit lost trying to figure out what was going on with the line ending changes, and how to deal with the conflicts while merging this in. Somewhere along the way I rebased the changes on top of the new head of master. I'm just going to go with that, and stop worry about this. I forced pushed to this branch as an easy way to get it back into a known working state.

PR #115, opened 2018-07-12
Add template methods to Read/Write vector objects

DanRStevens:


DanRStevens:
Thanks for the testing and code update

PR #118, opened 2018-07-14
Add CMake Project Generation

Brett208:
I am generating the new Visual Studio project via CMake into the directory VisualStudio. I added the directory VisualStudio to .gitignore for this purpose, although others can use whatever directory they want.

There are 2 Visual Studio specific parts right now:
 - Setting /MT and /MTd to ensure the library compiles in a static link
 - Adding filters to the Visual Studio project since Visual Studio decided to support virtual folders instead of the real thing.

I'm unsure if these lines need to be prevented from running on Linux. I suspect they may do nothing though for a makefile generation.

Coupling this with a CMake file for OP2Archive, I have an OP2Archive solution generating on Windows that brings in OP2Utility and compiles into a successful executable, all starting from CMake. 

For Visual Studio users, it adds 4 default build configurations. I would highly recommend only using Debug and RelWithDebugInfo. Using Release or RelMinSize means no PDB file will be generated for debugging, which is likely not what is desired. Losing these files would hinder local debugging of release builds before they are actually released and attempting to debug a release build that has been released for a long time.

For now, we can leave the makefile and Visual Studio project files in the code. They should probably be removed once everything is smooth and comfortable though.

-Brett

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -67,6 +67,10 @@ add_library(OP2Utility STATIC "${SOURCE_CODE}" "${EXTERNAL_HEADERS}")
 
 target_include_directories(OP2Utility PUBLIC ${CMAKE_CURRENT_LIST_DIR}/include)
 
+IF (NOT MSVC) #stdc++fs creates a LNK1104 on Visual Studio 2017
+target_link_libraries(${PROJECT_NAME} stdc++fs)
+ENDIF()
+
 # Set Visual Studio to using static code generation in all default Cmake generated configurations
 set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} /MT")
 set(CMAKE_CXX_FLAGS_MINSIZEREL "${CMAKE_CXX_FLAGS_RELEASE} /MT")


DanRStevens:
Ok, I gave it a try on Linux, and was able to produce an executable. I needed to add a few settings to the CMakeLists.txt file to get the required compiler flags.

Useful reading:
[How to Activate C++11 in CMake](https://stackoverflow.com/questions/10851247/how-to-activate-c-11-in-cmake)  (We actually need C++14)
[Build project with "experimental/filesystem" using CMake](https://stackoverflow.com/questions/44476810/build-project-with-experimental-filesystem-using-cmake)

Additionally, we might want to use one of these:
[CXX_STANDARD_REQUIRED](https://cmake.org/cmake/help/v3.1/prop_tgt/CXX_STANDARD_REQUIRED.html)
[target_compile_features()](https://cmake.org/cmake/help/v3.1/command/target_compile_features.html)

Some additional info which I skimmed:
[Enabling C++11 And Later in CMake](https://crascit.com/2015/03/28/enabling-cxx11-in-cmake/)

I was able to compile with the modified files.

Issue #109, opened 2018-07-06
Implement ClmFile::OpenSeekableStreamReader


Brett208:
Currently, the function OpenSeekableStreamReader is only implemented on VolFile. The challenge in ClmFile is that the required header file for the music track is contained in a separate location than the body and must be read first. 

Also, some work is currently done on the header when it is extracted to create a legitimate file. So, it is not a simple read from header first, then read from body problem. (This work on the header might be avoidable if we change how the header is structured and loaded into the ClmFile, but I haven't spent any thought on it).

```std::unique_ptr<SeekableStreamReader> ClmFile::OpenSeekableStreamReader(int fileIndex)```

I would propose as a solution: 
 1. Write the entire music track into a memory buffer by constructing the header and body together.  
 2. Load the new buffer into a MemoryStreamReader
 3. Pass the MemoryStreamReader out as a unique_ptr for consumption

The code to construct the stream would parallel (maybe reuse most of) the code required when extracting a single music file. 

-Brett

DanRStevens:
I'm not sure we want `OpenSeekableStreamReader` to produce a stream identical to a .wav file. It might make more sense to produce a stream equivalent to the contents of the data section of a .wav file. That is, a stream of PCM data. A WaveReader class might produce the same from a .wav file for going the other direction. If you want to build a .wav file from CLM output, you slap the appropriate headers onto it, and then dump the PCM stream contents.

I suppose this highlights another area for expansion of the stream library. Concatenating multiple streams together. That would be one way of building a .wav file from a header and a PCM data stream. Though, I don't think that is strictly necessary. Extract code can just manually write a header before copying the PCM stream to the output.

I think we can extract .wav data from a CLM without doing a full memory load of the PCM data, and it shouldn't be too difficult to accomplish. Especially not once we merge in code to copy data from one stream to another.

Oh, and we should probably rename `OpenSeekableStreamReader`. It's a bit wordy, and it's just naming a data type. Instead, we should describe what it does, or what the returned data represents. Perhaps `OpenPcmStream` (recommended), or `OpenWaveStream` (not recommended).

Brett208:
Good point that OpenSeekableStreamReader is a poor name. OpenSeekableStreamReader is an inherited function from ArchiveUnpacker. We would have to destroy the virtual function to name it something music file specific as suggested. How about OpenPackedFileStream?

[...]

Issue #102, opened 2018-06-30
Allow for extracting a file from an archive by referencing its internal filename only


Brett208:
Currently a user must use the internal filename to determine the file's index and then pass the index to extract the file. This should be accomplished by one call to the library.

DanRStevens:
Hmm, I think I'd be willing to accept one-liner as opposed to one-call.

As it stands, it feels like we have too many virtual functions, some being simple glue for the others. I'm wondering if we should reduce the number of virtual methods, and instead replace some with small inline methods? That might not be very interface like, but it seems to make more sense.

This is all just thinking out loud. I may change my mind on this.

PR #121, opened 2018-07-20
Improve archive unpacker extraction functions

Brett208:
Closes issue #109 
Closes issue #102 

Will need to followup with OP2Archive to reduce code using new ArchiveUnpacker::ExtractFile function overload.

### DIFF
--- a/src/Archives/ArchiveUnpacker.h
+++ b/src/Archives/ArchiveUnpacker.h
@@ -23,7 +23,7 @@ namespace Archives
 
 		virtual std::string GetInternalFilename(int index) = 0;
 		// Returns -1 if internalFilename is not present in archive.
-		virtual int GetInternalFileIndex(const std::string& internalFilename) = 0;
+		int GetInternalFileIndex(const std::string& internalFilename);
 		virtual uint32_t GetInternalFileSize(int index) = 0;
 		virtual void ExtractFile(int fileIndex, const std::string& pathOut) = 0;
 		void ExtractFile(const std::string& internalFilename, const std::string& pathOut);
--- a/src/Archives/ArchiveUnpacker.h
+++ b/src/Archives/ArchiveUnpacker.h
@@ -26,6 +26,7 @@ namespace Archives
 		virtual int GetInternalFileIndex(const std::string& internalFilename) = 0;
 		virtual uint32_t GetInternalFileSize(int index) = 0;
 		virtual void ExtractFile(int fileIndex, const std::string& pathOut) = 0;
+		void ExtractFile(const std::string& internalFilename, const std::string& pathOut);
 		virtual void ExtractAllFiles(const std::string& destDirectory);
 		virtual std::unique_ptr<SeekableStreamReader> OpenSeekableStreamReader(const std::string& internalFilename) = 0;
 		virtual std::unique_ptr<SeekableStreamReader> OpenSeekableStreamReader(int fileIndex) = 0;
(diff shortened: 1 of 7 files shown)


src/Archives/ClmFile.cpp
```diff
@@ -112,7 +112,11 @@ namespace Archives
 	{
 		CheckPackedFileIndexBounds(fileIndex);
 
-		throw std::logic_error("OpenStream not yet implemented for Clm files.");
+		FileSliceReader reader = clmFileReader.Slice(
+			indexEntries[fileIndex].dataOffset,
+			indexEntries[fileIndex].dataLength);
+
+		return std::make_unique<FileSliceReader>(reader);
```
DanRStevens: I'm wondering if there are performance consequences of this, and if the result of `Slice` should be stored immediately into such a variable, or even constructed as part of the return statement. I'm guessing there are a few constructor and copy constructor calls.

src/Archives/ArchiveUnpacker.h
```diff
@@ -23,7 +23,7 @@ namespace Archives
 
 		virtual std::string GetInternalFilename(int index) = 0;
 		// Returns -1 if internalFilename is not present in archive.
-		virtual int GetInternalFileIndex(const std::string& internalFilename) = 0;
+		int GetInternalFileIndex(const std::string& internalFilename);
```
DanRStevens: Why was `virtual` removed?

src/Archives/ArchiveUnpacker.h
```diff
@@ -20,15 +20,16 @@ namespace Archives
 		uint64_t GetVolumeFileSize() { return m_ArchiveFileSize; };
 		int GetNumberOfPackedFiles() { return m_NumberOfPackedFiles; };
 		bool ContainsFile(const std::string& filename);
+		// Returns -1 if internalFilename is not present in archive.
+		int GetInternalFileIndex(const std::string& internalFilename);
+		void ExtractFile(const std::string& internalFilename, const std::string& pathOut);
```
DanRStevens: This is one of those base classes where I had a mix of virtual methods with non-virtual methods, and a handful of member variables. I kind of regretted some aspects of the design later. As a note for going forward, my design of this base class might not be such a great example to base future code on.

In terms of the current changes, you seem to have done a pretty decent job with the breakdown of what should be virtual and what should be non-virtual. As much as the original implementation still worries me, I think I actually agree with what you've done here.

src/Archives/ClmFile.cpp
Brett208: I didn't create the slice inside the make_unique call because it was making the line of code long and more difficult to understand. So readability drove the separation. If there are performance concerns, we can combine it, as it wouldn't be too hard to read.

src/Archives/ArchiveUnpacker.h
Brett208: I think virtual classes are a fine solution in this case. It was more the multiple inheritance of 2 virtual classes into ArchiveFile I was/am unsure about.

Brett208:
Thank you for the quick review of the code and the compliments on the work.

-Brett

DanRStevens:
I'm going to go ahead and merge this.

The two points above don't really have a recommended change. The first one might require some investigation to see if there is any performance impact due to multiple copies of the object being constructed. The performance impact is minimal, if any. The second point was more about design considerations. I am somewhat concerned about mixing non-virtual methods into the base class, which is otherwise mostly virtual methods. Though it's done in such a way that the non-virtual methods delegate all significant work to the virtual methods. I think I can live with that, and it matches the use of template methods in the stream design.

As such, I think this PR is complete.

PR #122, opened 2018-07-20
Set CMake standard required setting

DanRStevens:
Without the standard required setting, CMake can try to fall back to previous versions of C++. We explicitly use C++14 features though, so the code would fail to compile under earlier standards.

The reason for this setting seems to be the default standard on Linux is an older one, with new features and updates needing to be explicitly enabled. Hence code does not compile using the latest updates unless it is explicitly requested. This does not always mean code needs the latest features to compile.

Also modified the spacing on `set` to match other instances.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,8 @@ cmake_minimum_required(VERSION 3.1 FATAL_ERROR)
 
 project("OP2Utility")
 
-set (CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
 # https://cmake.org/cmake/help/v3.12/command/file.html
 # CMake documentation recommends not collecting all source files using a GLOB


Brett208:
This seems like a good change to me. VS2017 assumes use of C++14 so I hadn't thought about it.

@DanRStevens what do you think about adding these lines to the OP2Archive CMakeLists.txt script as well to indicate that project is using C++11/14 features as well? My intention is to default to default using C++11/14 features in OP2Archive and OP2MapImager. For OP2Utility, I'd also like to default to using C++11/14 with the addition of C++17 filesystem.

I haven't actually checked to see if OP2Archive actually contains any C++14 calls, but I would like to make it available by default for future development if that makes sense?

DanRStevens:
I think the other projects should be updated as well. I agree about assuming C++14 features being available. By virtue of those projects depending on this library, you can expect them to have a C++14 capable compiler available. Further, we've used a few C++14 features in the header files which are imported by the other projects, so they should now require C++14 to compile.

I'll make some updates and test with the other projects.

PR #124, opened 2018-07-21
Update makefile

DanRStevens:
Some small changes to keep the old makefile up to date.

I guess we'll probably remove this along with the Visual Studio project files at some point. Though it would be nice to have working up to date files at the point of removal, in case we ever need to go back and compare.


Brett208:
Sounds good to me, although I don't have enough knowledge to comment on the changes.

PR #127, opened 2018-07-22
Implement stream copy

### LINKED ISSUE #84: Allow for automatic writing from an existing stream

DanRStevens:
Allow a StreamWriter to Write the entire contents from a StreamReader.

I think this is sufficient to close #84.

My first concern is the design. As implemented, it adds a dependency on StreamReader to the StreamWriter class. I generally like to avoid this, but it seems reasonable here.

It would be nice to start adding unit tests to the library. These kind of changes get a bit difficult and tedious to check manually. The current code compiles, but I haven't run it on anything.


Brett208:
Thanks for writing this. I'll test it by hooking it to ClmFile and VolFile. I'm not sure if I'll have time to complete testing today though.

Are you planning to implement a version that consumes an entire stream and/or handles reads past the end of the stream as an exception? This one seems more in line with the ReadPartial solution.

It looks like a pretty decent chunk of code to leave in the header file as an inline function. I've heard of entire C++ libraries being contained in the header file for so I don't have a good feeling on when it is too much to implement header only. Could you go over what made you decide to make this an inline function?

My general plan was to incorporate some way to package OP2Archive and OP2MapImager before circling back to unit testing. I haven't been working as quickly as in the past though.

Thanks,
-Brett

DanRStevens:
I think this is the only method we need. It consumes the entire stream. If we want to consume a partial stream, we can slice and then just consume the slice. No length parameter is required, and it works with the base `StreamReader` type.

End of stream is handled by using `ReadPartial`. It consumes buffer sized chunks of data until there is no data left to fill the buffer. No exceptions are used to detect end of stream. Rather, a read of 0 bytes determines end of stream.

----

I put the method in the header file simply because it was a template. I would prefer it not be in the header file, as I don't like the additional includes, and I don't like significant function implementations in the header. However, template methods usually do end up in the header file. This is largely due to C++ restrictions, where you get link errors if the template definition and the types to instantiate it for can't both be seen at the same time. For reference, see [Templates Definition versus Declaration](https://isocpp.org/wiki/faq/templates#templates-defn-vs-decl). I don't believe the method needs or should be declared `inline`.

I did just now try moving the method to a .cpp file, and it seems to compile and link fine. I even tried calling it with an alternate template buffer size parameter, and it compiled and linked fine. I was expecting to get link errors, particularly with a non-default buffer size. I'm wondering if this is a compiler specific extension, or an update to the C++ standard that allows it to work. I may do some more testing in that regard. I'm still a little afraid moving it to a .cpp file will bite us in the butt in some unexpected way.

----

During my testing, I noticed it's not quite so convenient to use slices, since the parameter is a non-const reference, and slices are by default rvalues until they are assigned to a local variable. That means this patch may need a bit more work if it wants to support:
```cpp
streamWriter.Write(streamReader.Slice(length));
```
Currently, I needed to use:
```cpp
auto slice = streamReader.Slice(length);
streamWriter.Write(slice);
```

----

I think unit testing is an important part of the packaging process. Certainly it helps verify changes as they are prepared for the next version. I'm starting to feel we need to prioritize testing at this point.

----

I may do a bit more work on this today. In particular, I may change some of the old code to use the new method. That may speed up your ability to test later on.

DanRStevens:
The above commits are the changes needed to use the new code.

I'm going to include one more commit below to move the template method to an implementation file. I'm a little concerned this might not be a good idea, but I figure we should at least test it. If it turns out to be a bad idea, we can revert to the previous commit.

Brett208:
Thank you for updating the archive code to use the new stream read/write. It did make testing easier. I like seeing the ArchiveUnpacker and ArchivePacker code lose their individual interpretations of consuming a stream. 

This branch produces the following errors on an x86 debug build for VS2017:

Error	LNK2019	unresolved external symbol "public: void __thiscall StreamWriter::Write<131072>(class StreamReader &)" (??$Write@$0CAAAA@@StreamWriter@@QAEXAAVStreamReader@@@Z) referenced in function "public: virtual void __thiscall Archives::ClmFile::ExtractFile(int,class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > const &)" (?ExtractFile@ClmFile@Archives@@UAEXHABV?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@@Z) OP2Archive C:\ . . . \OP2Archive\OP2Utility.lib(ClmFile.obj) 1	

Warning LNK4221 This object file does not define any previously undefined public symbols, so it will not be used by any link operation that consumes this library OP2Utility C:\ . . . \OP2Archive\OP2Utility\StreamWriter.obj 1
	
Error	LNK2001 unresolved external symbol "public: void __thiscall StreamWriter::Write<131072>(class StreamReader &)" (??$Write@$0CAAAA@@StreamWriter@@QAEXAAVStreamReader@@@Z) OP2Archive C:\ . . . \OP2Archive\OP2Utility.lib(VolFile.obj) 1	

Error	LNK1120	1 unresolved externals OP2Archive C:\ . . . \OP2Archive\Debug\OP2Archive.exe	1	

When switching to the version containing the template implementation in the header file, the code compiles without error or warning. So, I guess we are stuck with leaving it in the header file.

If we do include the .cpp file, I would recommend switching it to 4 space tabs to match the rest of the codebase and we need to add it to the VS2017 project file and filters file for compilation.

I think supporting `streamWriter.Write(streamReader.Slice(length));` is worth trying to get to eventually. I would also prefer seeing a way to copy a portion of a stream without slicing it if someone wanted to use the library as such.

Unless we want to put some time trying to get StreamWriter.cpp working with MSVC, I would be happy seeing this merged in current state.

DanRStevens:
Ok, I reverted the change that moved the code from the header to its own file. I had a feeling that would cause issues somewhere. Curiously though, the AppVeyor builds didn't seem to have a problem with it.

Before the revert, I converted from spaces to tabs, just so we'd have a commit in that state in case we want to go back and investigate. Good catch on spotting that.

We may be able to support `streamWriter.Write(streamReader.Slice(length));` by using rvalue references. I might want to do a bit more reading in that area. I'll consider this for a future update.

PR #128, opened 2018-07-28
Disable MSVC warning 4996 for strncpy in ClmFile

### LINKED ISSUE #125: Handle MSVC warning when calling strncpy in ClmFile.cpp
Within ClmFile.cpp, MSVC posts a warning, requesting strncpy be switched to strncpy_s since it is a 'safer' alternative. I'd like to silence this warning.

strncpy_s differs in execution from strncpy in that it does not continue writing zeroes to pad up to count. I think we discussed wanting this behavior, writing zeros to pad to the count. I'm not sure why this matters though. In this case we initialized IndexEntries ourselves so there shouldn't be any nefarious data to overwrite inside of IndexEntries[i].filename. 

Oddly though, even after I add the header string.h on MSVC, I have to call strncpy_s without prepending std:: to it. MSVC refuses to recognize std::strncpy_s.

If there is a legitimate reason to use strncpy over strncpy_s here, I can push a preprocessor directive to ignore the warning.

Explaining strncpy_s from https://en.cppreference.com/w/c/string/byte/strncpy
> Same as (1), except that the function does not continue writing zeroes into the destination array to pad up to count, it stops after writing the terminating null character (if there was no null in the source, it writes one at dest[count] and then stops). Also, the following errors are detected at runtime and call the currently installed constraint handler function: 
> * src or dest is a null pointer 
> * destsz or count is zero or greater than RSIZE_MAX 
> * count is greater or equal destsz, but destsz is less or equal strnlen_s(src, count), in other words, truncation would occur 
> * overlap would occur between the source and the destination strings

I'd also be open to a third alternative to strncpy or strncpy_s, since they both feel a little bit out of place to me here.

Brett208:
Closes #125

### DIFF
--- a/src/Archives/ClmFile.cpp
+++ b/src/Archives/ClmFile.cpp
@@ -273,13 +273,11 @@ namespace Archives
 		uint32_t offset = headerSize + internalNames.size() * sizeof(IndexEntry);
 		for (std::size_t i = 0; i < internalNames.size(); ++i)
 		{
-#pragma warning( push )
-#pragma warning( disable : 4996 ) // Disable MSVC warning 'strncpy': This function or variable may be unsafe. Consider using strncpy_s instead.
-			// As of 28Jul18, strncpy_s was not available on Linux.
-			
+			// Disable MSVC warning 'strncpy': This function or variable may be unsafe. Consider using strncpy_s instead.
+			// As of 28Jul18, strncpy_s was not available on Linux.		
 			// Copy the filename into the entry
+#pragma warning( suppress : 4996 )
 			std::strncpy(indexEntries[i].filename.data(), internalNames[i].data(), sizeof(IndexEntry::filename));
-#pragma warning ( pop )
 
 			// Set the offset of the file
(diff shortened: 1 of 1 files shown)


src/Archives/ClmFile.cpp
```diff
@@ -273,8 +273,13 @@ namespace Archives
 		uint32_t offset = headerSize + internalNames.size() * sizeof(IndexEntry);
 		for (std::size_t i = 0; i < internalNames.size(); ++i)
 		{
+#pragma warning( push )
+#pragma warning( disable : 4996 ) // Disable MSVC warning 'strncpy': This function or variable may be unsafe. Consider using strncpy_s instead.
```
DanRStevens: Looks correct.

This is slightly wordier than I expected it to be. I had erroneously thought the push and disable could be combined. Turns out there is an even better option though. Checking MSDN for [warning](https://msdn.microsoft.com/en-us/library/2c8f766e.aspx), I noticed there is a `suppress` option, which will push the current state, disable a warning for the next line, and then pop the state.

PR #130, opened 2018-08-07
Standardize use of Tileset

Brett208:
There were still a few uses of TileSet in the codebase. All changed to Tileset.


Brett208:
Some difficulty with Git recognizing the difference in filenames because it was only a case change involved.

Brett208:
Good call on the squash and merge

PR #131, opened 2018-08-07
Read/Write strings with size prefixed

Brett208:
- Test using MapReader/Writer

Completed tests with both an x86 and x64 compilation of a modified OP2MapImager that allowed saving a new map. Map used in testing was Ashes.map. Was able to render the map properly, create a new map that was the same size as the original, and load the newly created map using Outpost 2.

### DIFF
--- a/src/Streams/StreamReader.h
+++ b/src/Streams/StreamReader.h
@@ -3,6 +3,7 @@
 #include <cstddef>
 #include <type_traits>
 #include <vector>
+#include <string>
 
 class StreamReader {
 protected:
@@ -38,5 +39,19 @@ public:
 	template<typename T, typename A>
 	inline void Read(std::vector<T, A>& vector) {
 		ReadImplementation(vector.data(), vector.size() * sizeof(T));
+	}
+
+	// std::string prefixed by the string's size. The type of integer representing the size must be provided. 
+	template<typename SizeType>
+	void Read(std::string& string) {
+		SizeType stringSize;
+		Read(stringSize);
+		if (stringSize < 0) {
+			throw std::runtime_error("String's size may not be a negative number");
+		}
+
+		string.clear();
+		string.resize(stringSize);
+		Read(&string[0], stringSize);
 	}
 };
--- a/src/Streams/StreamWriter.h
+++ b/src/Streams/StreamWriter.h
@@ -6,6 +6,7 @@
 #include <type_traits>
 #include <vector>
 #include <array>
+#include <string>
 
 class StreamWriter {
 protected:
@@ -34,6 +35,19 @@ public:
 	template<typename T, typename A>
 	inline void Write(const std::vector<T, A>& vector) {
 		WriteImplementation(vector.data(), vector.size() * sizeof(T));
+	}
+
+	// std::string prefixed by the string's size. The type of integer representing the size must be provided.
+	// Does not null terminate string unless null terminator is specifcally included in passed std::string. 
+	template<typename SizeType>
+	void Write(const std::string& string) {
+		auto stringSize = string.size();
+		if (stringSize > std::numeric_limits<SizeType>::max()) {
+			throw std::runtime_error("String's size is too large to write in provided size field");
+		}
+		auto typedSize = static_cast<SizeType>(stringSize);
+		Write(typedSize);
+		Write(string.data(), string.size());
 	}
 
 	// Copy a StreamReader to a StreamWriter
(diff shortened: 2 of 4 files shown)


Brett208:
I missed some header includes that MSVC was able to cope with missing but were needed on Linux, and really should have been included on MSVC as well. 

Not sure what to do about Travis CI not compiling for the current pull request.

src/Streams/StreamReader.h
```diff
@@ -39,4 +41,18 @@ class StreamReader {
 	inline void Read(std::vector<T, A>& vector) {
 		ReadImplementation(vector.data(), vector.size() * sizeof(T));
 	}
+
+	// std::string prefixed by the string's size. The type of integer representing the size must be provided. 
+	template<typename SizeType>
+	void Read(std::string& string) {
+		SizeType stringSize;
+		Read(stringSize);
+		if (stringSize < 0) {
```
DanRStevens: I wonder if there is a way to eliminate this check if `SizeType` is `unsigned`.

What if `SizeType` is greater than `std::size_t`, such as a `uint64_t` on a 32-bit build.

src/Streams/StreamWriter.h
```diff
@@ -36,6 +39,19 @@ class StreamWriter {
...
+	// Does not null terminate string unless null terminator is specifcally included in passed std::string. 
+	template<typename SizeType>
+	void Write(const std::string& string) {
+		auto stringSize = string.size();
+		if (stringSize > std::numeric_limits<SizeType>::max()) {
+			throw std::runtime_error("String's size is too large to write in provided size field");
+		}
+		auto typedSize = static_cast<SizeType>(stringSize);
+		Write(typedSize);
+		Write(string.data(), string.size());
```
DanRStevens: There is some inconsistency between the use of `string.data()` and `&string[0]` in this change set.

Additionally, it feels slightly weird to `Write(typedSize)`, and then write `string.size()` characters. I almost feel like the same variable should be used in both places. Though the datatype may feel more natural with the code as written.

Brett208: I prefer string.data(). However in StreamReader, I receive the error below when writing the line: 

```
Read(string.data(), stringSize);
```

Error C2664 'void StreamReader::Read(void *,size_t)': cannot convert argument 1 from 'const char *' to 'void *' OP2Utility c:\ . . . \op2mapimager\op2utility\src\streams\streamreader.h 56

So I was forced to write it there as &string[0]

Looking at this a little closer I'm not sure there is any reason to keep the line with the static_cast. It doesn't do anything. instead it could be written as:
```
Write(string.size());
Write(string.data(), string.size());
```
The variable typedSize and the static_cast can just be removed. After we check the numeric_limit, there isn't any reason to cast anything???

Thanks,
Brett

Brett208:
Good to know about TravisCI. I'll try that next time it fails. I had noticed that it wasn't pulling dependencies, but didn't realize that I could 'redo' the test.

Overall, I think I prefer overloading Read to more function names. It makes calling Read cleaner in the rest of the code. However, it requires close reading of the header file to understand the different options. Someone newer to the library might prefer more description in the function names. We could use a couple of user tests to see what someone new to the library does.

I'm generally happy with the code in this form besides the comment I made about removing the static_cast. It sounds like you would like to improve the safety checks, which I think is smart but don't really have any thoughts on how to accomplish. If you want to do so on this branch, I don't mind if you want to 'take the reigns'. If you want to mull over / research before making the changes, I would prefer just merging this branch and then opening a new one when the changes are ready.

Any branch open when the x64 branch is merged will likely have merge conflicts.

-Brett

src/Streams/StreamWriter.h
DanRStevens: Actually, there is reason for the cast, since the `Write` is dependent on the data size. If you got rid of the variable you would still need the cast to write the correct number of bytes. The following may work well:
```cpp
Write(static_cast<SizeType>(stringSize));
Write(string.data(), stringSize);
```

Depending on the compiler and compiler options, using `string.size()` everywhere may result in multiple calls to the `.size()` method. I suspect most compilers won't be smart enough to realize the value is constant and optimize the extra calls away and just use a cached temporary. Optimizing across calls is more difficult than within a method.

----

As for [`.data()`](https://en.cppreference.com/w/cpp/string/basic_string/data), keep in mind the signature was changed over different C++ versions. Before C++17, it was `const`, so inappropriate to pass to read methods (which write into the passed object). For ['operator[]`](https://en.cppreference.com/w/cpp/string/basic_string/operator_at) the constness matches the constness of the object itself. This naturally works where it's supposed to, and gives compilers errors where needed.

DanRStevens:
I'd be happy to merge as is and tackle the static checks later. I went and opened an issue for them.

Brett208:
Sounds good.

PR #136, opened 2018-08-10
Implement length prefix vector read

### LINKED ISSUE #133: Implement length prefixed string/vector StreamReader/Writer template methods
A template method for length prefixed vector write was implemented in PR #129. We should write the corresponding read method. We should also write corresponding methods for length prefixed strings.

As part of this, we should update the crudely written comments and exception message for the existing implementation.

DanRStevens:
Curiously, the implementation is exactly the same as the templated
length prefixed string read. Perhaps there is a way to make the template
parameter more generic and provide both with one source method.

One complication with combining length prefixed vector and string
methods is they both contain an implicit assumption of contiguous
memory. This assumption comes from the `Read` on the last line of both
methods. Although this assumption is guaranteed by the standard for both
`vector` (since C++03), and `string` (since C++11), it is not true for
other more generic STL containers. Hence combining code to one method
would need to restrict the method to those two data types to prevent bad
code generation for other STL containers.

For now though, this works, and closes #133.

### DIFF
--- a/src/Streams/StreamReader.h
+++ b/src/Streams/StreamReader.h
@@ -42,6 +42,19 @@ public:
 		ReadImplementation(vector.data(), vector.size() * sizeof(T));
 	}
 
+	// Size prefixed vector data types
+	template<typename SizeType, typename T, typename A>
+	void Read(std::vector<T, A>& vector) {
+		SizeType vectorSize;
+		Read(vectorSize);
+		if (vectorSize < 0) {
+			throw std::runtime_error("Vector's size may not be a negative number");
+		}
+		vector.clear();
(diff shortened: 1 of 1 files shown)


Brett208:
I wouldn't mind supporting read/write of other standard library containers. Map and list come to mind. In C#, I would serialize a dictionary (C++ map) into binary and xml fairly often.

DanRStevens:
Yeah, I was thinking of ways to extend support for other containers. Perhaps those ideas can go into a future branch. In particular, the `Read` line can be replaced by a template method, which might just call `Read` for contiguous containers, or use an iterated read for other containers. Would need to think about it and test though.

PR #137, opened 2018-08-11
Template string methods

### LINKED ISSUE #68: Stream Read/Write helpers for complex types

DanRStevens:
Allow reading/writing of strings without length prefix.

Closes #68

### DIFF
--- a/src/Archives/VolFile.cpp
+++ b/src/Archives/VolFile.cpp
@@ -381,7 +381,7 @@ namespace Archives
 
 		std::string charBuffer;
 		charBuffer.resize(actualStringTableLength);
-		archiveFileReader.Read(&charBuffer[0], actualStringTableLength);
+		archiveFileReader.Read(charBuffer);
 
 		m_StringTable.push_back("");
 		for (std::size_t i = 0; i < charBuffer.size(); ++i)
--- a/src/Streams/StreamReader.h
+++ b/src/Streams/StreamReader.h
@@ -43,6 +43,7 @@ public:
 	}
 
 	// Size prefixed vector data types
+	// Ex: Read<uint32_t>(vector); // Read 32-bit vector size, allocate space, then read vector data
 	template<typename SizeType, typename T, typename A>
 	void Read(std::vector<T, A>& vector) {
 		SizeType vectorSize;
--- a/src/Streams/StreamReader.h
+++ b/src/Streams/StreamReader.h
@@ -55,9 +55,16 @@ public:
 		Read(vector);
 	}
 
-	// std::string prefixed by the string's size. The type of integer representing the size must be provided. 
-	template<typename SizeType>
-	void Read(std::string& string) {
+	// String data types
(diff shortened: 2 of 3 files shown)


DanRStevens:
Note: This relates to Issue #68. I think we should close that issue after merging this. I'm adding an edit to that effect.

Brett208:
I found part of the archive code that can be used to test this. I'll check it out hopefully later today.

DanRStevens:
Ahh, good catch on the typo and the period.

The new comments are a good idea.

PR #138, opened 2018-08-12
Fix MapData::GetCellIndex Argument mismatch between prototype and definition

Brett208:


DanRStevens:
Hmm, odd this wasn't caught earlier.

Looking at the code, I think all the `x` and `y` values should be `std::size_t`. They are all used for memory indexing. I was also thinking the tile data could be represented as a struct with bit fields.

I'll make a few updates.

DanRStevens:
Ok, in terms of `std::size_t`, this is what I think the code should look like.

----

It seems the bit field struct was already being used. I just wasn't reading the code closely enough.

I suppose the use of `TileData` makes some of the Get methods of questionable value, since you could just access the `TileData` field directly. The main benefit of the class methods is taking care of the indexing. We could perhaps just have a `TileData(x, y)` method, which returns a `TileData` struct, and eliminate the other public methods.

Instead of:
```cpp
auto cellType = mapData.GetCellType(x, y);
```

This could be:
```cpp
auto cellType = mapData.TileData(x, y).cellType;
```

Issue #132, opened 2018-08-07
Implement binary search for VOL files


DanRStevens:
We should implement the filename lookup using a binary search. That would minimize differences between the utility and the game. Plus there's the small added bonus of the lookups being faster.

The binary search only applies to VOL files.

The CLM files don't appear to have any implicit assumption or requirement for the filename table to be sorted. The game only uses the names in the CLM file in one place, during initialization, where it uses a linear search to build an index table that maps a known table of expected names to actual CLM file indexes. The game then references all music using indexes from the pre-set known table, rather than the actual CLM contents.

```
004E2700 >004E26E4  ASCII "EDEN11"
004E2704  004E26DC  ASCII "EDEN21"
004E2708  004E26D4  ASCII "EDEN22"
004E270C  004E26CC  ASCII "EDEN31"
004E2710  004E26C4  ASCII "EDEN32"
004E2714  004E26BC  ASCII "EDEN33"
004E2718  004E26B4  ASCII "EP41"
004E271C  004E26AC  ASCII "EP42"
004E2720  004E26A4  ASCII "EP43"
004E2724  004E269C  ASCII "EP51"
004E2728  004E2694  ASCII "EP52"
004E272C  004E268C  ASCII "EP61"
004E2730  004E2684  ASCII "EP62"
004E2734  004E267C  ASCII "EP63"
004E2738  004E2670  ASCII "PLYMTH11"
004E273C  004E2664  ASCII "PLYMTH12"
004E2740  004E2658  ASCII "PLYMTH21"
004E2744  004E264C  ASCII "PLYMTH22"
004E2748  004E2640  ASCII "PLYMTH31"
004E274C  004E2634  ASCII "PLYMTH32"
004E2750  004E2628  ASCII "PLYMTH33"
004E2754  004E261C  ASCII "STATIC01"
004E2758  004E2610  ASCII "STATIC02"
004E275C  004E2604  ASCII "STATIC03"
004E2760  004E25F8  ASCII "STATIC04"
004E2764  004E25EC  ASCII "STATIC05"
```

DanRStevens:
Here's an initial rough implementation of the binary search:
```cpp
	std::size_t VolFile::GetIndex(const std::string& name)
	{
		std::size_t start = 0;
		std::size_t end = GetCount(); // We need GetCount() - 1, but...

		// This guards against wrap around when GetCount() == 0
		if (end-- > 0) {
			// Binary search
			do {
[...]

PR #140, opened 2018-08-13
Add override to archive methods

DanRStevens:
This contains a few updates for compiler checks. I noticed this while preparing to tackle Issue #132.

In particular, the `std::size_t GetIndex(const std::string& name)` method was not virtual. For the VOL file and CLM file to have different implementations, this will need to be virtual.

The `VolFile` class had declared a corresponding `virtual` method, but never implemented it. This did not cause an error because the base class didn't declare the method virtual, and calling code used a base class pointer, hence the compiler never saw any calls to the unimplemented method.

As a new implementation does not yet exist, the line to override this method is commented out. Hence VOL file is still using the base class linear search algorithm at this point.


Brett208:
Looks like good cleaning of the archive code. 

Tested by compiling in both x86 and x64 on MSVC. Able to create and extract contents of a vol file in both x86 and x64 without any issues.

PR #141, opened 2018-08-14
Rename ClipRect struct to Rect

DanRStevens:
It's the variable instance that represents a `clipRect`. The struct is just a generic Rect.


Brett208:
Seems like a good improvement.

Issue #134, opened 2018-08-07
Review HuffLZ methods `GetNumExtraBits` and `GetRepeatOffset`


DanRStevens:
The `GetNumExtraBits` and `GetRepeatOffset` methods contain cascaded `if`/`else` blocks. This is slightly undesirable as it embeds data into the code in a hard to read way. Perhaps it can be rewritten as a data table lookup.

Some attention should be paid to performance, as this is an inner part of the decompression algorithm. It is however a branch of the inner part, which might not execute too frequently, so performance might not be a critical concern in practice.

Closing this issue without changes is also a valid option. The code as written is correct, and does not suffer from known performance issues. The issue is merely one of code style.

Brett208:
I'm spending some time refactoring HuffLZ. It may not lead to a solution to this problem, but wanted to flag my work to reduce duplicated efforts in HuffLZ.

DanRStevens:
I was thinking about this today. I was tempted to close it as not a priority, but it managed to grab my curiosity.

If we go with a data structure approach, we need a way to map a point to a range, or rather, values associated with a range. Important details here are the intervals are non-overlapping, and adjacent, and there is a default case that covers all possible values. That is, no point belongs to more than one range, and no point falls between or around ranges. That is, every point belongs to exactly one range.

There are various algorithms that map points to ranges, such as [Interval Trees](https://en.wikipedia.org/wiki/Interval_tree). Specifically, it mentions in the naive case, where intervals do not overlap, a binary search tree will do.

As our data is static, we could implement a binary search of a sorted table. However, I think a linear search would actually be better here. In particular, we want lookup to be fast. Normally, a binary search would have better worst case lookup time. Here however, we are dealing with a decompression algorithm, where input data is not evenly distributed. The codes themselves are variable length to account for this, in that shorter codes are assigned to more likely patterns. As such, we want lookup of shorter codes to be fastest, as they will be most common. Hence, a linear search is most suited here, since it will have a lower amortized lookup cost over the set of compression codes.

The current cascaded `if` statements are basically an unrolled loop doing a linear search of a data table.

In terms of the data needed, lookup is done by the range end value, though the start value is also used in calculations. There is also the `numExtraBits`, a `shiftFactor`, and an `additiveFactor`. One method is a simple lookup for `numExtraBits`. The other method is a lookup for the `range.start`, `shiftFactor`, and `additiveFactor`, followed by a calculation using those values.

[...]

PR #147, opened 2018-08-19
Refactor huff lz

Brett208:
I was looking at issue #134. My knowledge of what the code is accomplishing is insufficient to correct the issue. However, I did fix a couple of typos and the construction process in the HuffLZ which I thought were worth committing. 
 
I forgot to update the master branch with merged changes which created a merge conflict with the new name for BitStream. I was happy that GitHub chose to ignore the changes in the merge commit so they are not confusing the actual changes to the code base.


Brett208:
AdaptHuffTree rename is complete. Still need to fix class variable pointers in HuffLZ. Should have time tomorrow for this.

Brett208:
@DanRStevens,

What do you think of the following three lines:

```
HuffLZ::HuffLZ(BitStreamReader& bitStream) :
	m_BitStreamReader(bitStream),
	m_ConstructedBitStreamReader(), // Don't need to delete stream in destructor 
```

I think this is the proper way to handle creating the HuffLZ class around an existing bitstream, but wanted to make sure it looked good to you.

I have compiled the code, but have not used it to actually read a HuffLZ compressed volume yet. I will do that before merging.

Thanks,
Brett

src/Archives/HuffLZ.cpp
```diff
@@ -4,10 +4,10 @@
 namespace Archives
 {
 	// Constructs the object around an existing bit stream
-	HuffLZ::HuffLZ(BitStreamReader *bitStream) :
+	HuffLZ::HuffLZ(BitStreamReader& bitStream) :
 		m_BitStreamReader(bitStream),
-		m_ConstructedBitStreamReader(0), // Don't need to delete stream in destructor
-		m_AdaptiveHuffmanTree(new AdaptiveHuffmanTree(314)), 
+		m_ConstructedBitStreamReader(), // Don't need to delete stream in destructor
```
DanRStevens: I find the empty parenthesis a bit odd here. I'm wondering if we should be more explicit about this. Passing a `nullptr` would be appropriate for the initialization value.

src/Archives/HuffLZ.h
```diff
@@ -31,9 +30,9 @@ namespace Archives
 		void WriteCharToBuffer(char c);
 
 		// Member variables
-		BitStreamReader *m_BitStreamReader;
-		BitStreamReader *m_ConstructedBitStreamReader;
-		AdaptiveHuffmanTree *m_AdaptiveHuffmanTree;
+		BitStreamReader m_BitStreamReader;
+		BitStreamReader m_ConstructedBitStreamReader;
```
DanRStevens: Ahh, I see you've changed the type of this field as well. Hmm, this field actually makes much less sense now. We should probably eliminate it. Previously the combination of `m_BitStreamReader` and `m_ConstructedBitStreamReader` together functioned somewhat like a `shared_ptr`. It would either reference an existing object which it did not own (and hence would not free), or it would allocated the object itself and thus be responsible for the lifetime and cleanup of that object.

By using containment, we are implying ownership. There is no question about it. If we are passed an object, it must be copied before it can be used. That is, for the constructor that takes the `BitStreamReader` object, the object must be copied rather than simply storing a reference to it.

In hindsight, I probably wrote the code as it was thinking there was a bigger performance gain by avoiding the copy. I had probably erroneously assumed the entire data buffer would be copied. In actuality though, it's a shallow copy, in that only the 4 fields defined in `BitStreamReader` are copied, such as the buffer pointer. It doesn't need to copy the actual data buffer itself. The copy operation is actually fairly cheap, and not frequently occurring. Performance implications should be fairly minimal. The copy operation is slightly more expensive this way, though there is less indirection when accessing the object, so some slight efficiency gains there. This might even be a net gain in terms of performance, and a big gain in terms of code readability.

My advice is to drop the `m_ConstructedBitStreamReader` field now that the referenced objects are contained in the host class.

As a side note, it would be reasonable to eliminate the alternate constructor that takes a pointer and size, and instead only take a constructed `BitStreamReader`. Such an object is cheap to construct and copy. The parameter could be `const`, since a (non-const) copy is made. That could allow a temporarily constructed object to be passed to the constructor. Eliminating the alternate constructor that takes individual pointer and size values cuts down on code, and cuts down on the number of ways the library could be used to accomplish the same task. That might actually be a usability gain.

DanRStevens: Side side note: The Stream code uses the parameter order `(void* buffer, std::size_t size)`, but the HuffLZ object reverses it to `(std::size_t size, void* buffer)`. We should probably try to be consistent about the ordering.

DanRStevens:
Ok, I removed the field in question.

I did not remove any of the constructors, nor address the order of constructor parameters. After taking a quick look, I think those are separate issues that should be addressed in their own PR. In particular, any changes there would relate to the compressed streams design, which we haven't fully fleshed out yet. Let's close this off meanwhile, and I'll add a separate issue with my notes on the other areas.

Brett208:
It looks like something we did has damaged the Huffman LZ decompression algorithm. The program executes without an exception, but the results appear partially mangled. It looks worse then just the file endings that we discussed earlier. I'll post the results in the forum.

-Brett

Brett208:
I did some more troubleshooting. Both x64 and x86 produce the same garbled results. Unpacking a version of sheets.vol that is not compressed produces proper results. 

I switched back to the master branch of OP2Utility, but still had the same garbled results. So it looks like we broke this somewhere before this branch.

I reverted OP2Utility back to 7/28/2018 "Fix warnings produced when writing strings to map files and unknown tileGroup variable" After modifying OP2Archive to work with this version of OP2Utility, I was able to extract a proper version of sheets.vol. It looks like the error was added sometime between 7/28/2018 and before the current branch.

It may bear fruit to check for changes in both how VolFile calls the decompress code and any changes to the decompress code between then and now.

Brett

DanRStevens:
Nice work discovering that bug, and determining it was pre-existing. Sounds a bit like you were doing a manual `git biset` to narrow in on a range of commits where the error was introduced. This seems like additional justification to have more unit tests.

Given the error is pre-existing, do we go ahead and merge the current change set, or hold until the error is fixed. It seems like the proper thing to do in a large widely used repository, would be to fix the error first, before committing new code. However, for a small library, it may make sense to just keep moving forward, and fix things along the way. At the very least, there would be fewer merge conflicts that way.

I'll look into adding some additional unit tests.

DanRStevens:
After looking at this with fresh eyes, I realized the additional changes I had proposed in Issue #150 are actually quite simple, so I went ahead and did them.

Brett208:
All the changes look good. Thanks for taking the time to make them.

After merging the master into this branch, I tested by decompressing sheets.vol again. Everything checked out fine.

-Brett

PR #148, opened 2018-08-19
Add tests

DanRStevens:
Add `makefile` support for compiling and running unit tests (`make check`).

Add example Google Test test cases for MemoryStreamReader. Currently there is some duplication, to provide examples of writing tests with and without fixtures. One block of code can be removed in a future edit.


Brett208:
@DanRStevens it looks like you are planning on using Google Test without including its sourcecode or some sort of .lib/.obj/.dll file. I just checked and Visual Studio offers built in support for adding a Google Test project to a solution. I have to download the module first though as I don't have it installed. I'll download it and try to add the tests you wrote and compile on MSVC without issues.

You will likely need to update the MemoryStreamReader examples to use the new Stream namespace and filenames.

-Brett

DanRStevens:
For this initial commit, I just assumed a system install of Google Test, for both the headers, and a compiled static library file. The compiled static library assumption is likely to change, as I suspect that would fail if we tried to run tests on the CI systems.

I was thinking of updating the makefile to do a local compile of the globally installed Apt source package for Google Test. I should probably hold off on that until you've had a chance to try things on Windows. If Windows practice is to include a local copy of GoogleTest, then I can set the Linux makefile to use the same copy. If Windows practice is to have a separate install, then I can continue using the Apt package. I'll let the Windows setup and CI build requirements inform how I proceed here.

Good point about the recent namespace change. I'll update a bit later. That should be independent of the Windows additions for Google Test.

Brett208:
I finished installing the VS Google Test module. After fiddling some with the settings, I managed to add a Test project to the OP2Utility Solution that compiled. It is giving me the following warnings:

Warning MSB8028 The intermediate directory (x64\Debug\) contains files shared from another project (OP2Utility.vcxproj). This can lead to incorrect clean and rebuild behavior. OP2UtilityTest C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\Common7\IDE\VC\VCTargets\Microsoft.CppBuild.targets	391

Warning C4996 'std::tr1': warning STL4002: The non-Standard std::tr1 namespace and TR1-only machinery are deprecated and will be REMOVED. You can define _SILENCE_TR1_NAMESPACE_DEPRECATION_WARNING to acknowledge that you have received this warning. OP2UtilityTest c:\ . . . \op2archive\op2utility\packages\microsoft.googletest.v140.windesktop.msvcstl.static.rt-static.1.8.0\build\native\include\gtest\internal\gtest-port.h 996

Warning C4996 'std::tr1': warning STL4002: The non-Standard std::tr1 namespace and TR1-only machinery are deprecated and will be REMOVED. You can define _SILENCE_TR1_NAMESPACE_DEPRECATION_WARNING to acknowledge that you have received this warning.	OP2UtilityTest	c:\ . . . \op2archive\op2utility\packages\microsoft.googletest.v140.windesktop.msvcstl.static.rt-static.1.8.0\build\native\include\gtest\internal\gtest-port.h 997

Warning C4996 'std::tr1': warning STL4002: The non-Standard std::tr1 namespace and TR1-only machinery are deprecated and will be REMOVED. You can define _SILENCE_TR1_NAMESPACE_DEPRECATION_WARNING to acknowledge that you have received this warning.	OP2UtilityTest	c:\ . . . \op2archive\op2utility\packages\microsoft.googletest.v140.windesktop.msvcstl.static.rt-static.1.8.0\build\native\include\gtest\internal\gtest-port.h 998

And a dozen other warnings of similar nature.

I named the project OP2UtilityTest. It resides in the same directory as OP2Utility's project files. Because of this, they are sharing the same build folders, which I believe is what is producing the Warning MSB8028. I could place OP2UtilityTest in a sub-directory or I could manually set 2 different named directories for the build to occur in. 

I'm not yet sure what is producing the rest of the warnings.

If we go with using the Visual Studio Google Test module, VS adds a packages subdirectory that contains about 35 MB of information which appears to be the Google Test dependencies. The root directory contains a new file called packages.config. Its contents are:

```
<?xml version="1.0" encoding="utf-8"?>
<packages>
  <package id="Microsoft.googletest.v140.windesktop.msvcstl.static.rt-static" version="1.8.0" targetFramework="native" />
</packages>
```

I think packages.config would be included in the repository but not the packages subdirectory. I assume when someone first loaded the Visual Studio solution on their computer, it would go and fetch this package for them using Microsoft's NUGET system. 

I'm leaning towards maybe manually adding Google Test and skipping the Microsoft Package Manager. It seems a bit fiddly. The next step is probably researching what others with VS users prefer.

-Brett

DanRStevens:
Ok, I merged in the changes from master, and updated the test cases for the name changes.

I've heard of Nuget, but have never played with it. I've heard it's one of the recommended ways to fetch dependencies, and is preferred over Git submodules, or embedding the dependent projects directly in the host repository. The data it fetches should be listed in the `.gitignore` file. Unless there is some kind of lock file. A version lock file is typically checked in to version control.

As for the shared build folder issue, what if the project files were moved into the `src` and `test` folders. Technically those are project folders, so it'd make sense to put the project files in there.

DanRStevens:
Ok, I'll go ahead and merge then. We can always make changes later once testing is setup on Windows.

I have little experience with multiple projects in the same workspace. I believe Visual Studio had a default where projects would be stored in a sub-folder of solution folders. I remember hating it and always unchecking that option, but then all my projects were just single projects. Maybe we should look to other open source projects for guidance.

PR #153, opened 2018-08-23
Refactor typedef nodeindex

### LINKED ISSUE #149: Unsigned Short typedef in AdaptiveHuffManTree
I noticed the typedef defined in the header file for AdaptiveHuffmanTree for unsigned short (USHORT) is not used in the cpp file. Instead of using USHORT, the cpp file uses unsigned short directly. This makes Visual Studio report that it cannot find the function definition (although it compiles fine with MSVC)

I was thinking about removing the typedef from the codebase. I'm assuming that since the typedef is defined in the public portion of the class declaration, it will be visible to any code that includes AdaptiveHuffmanTree, which I think is overall not desirable since if forces the typedef on other's code?

If we want to use it, I guess we should probably unify the class' cpp file to also use USHORT.

Thanks,
Brett

DanRStevens:
This closes #149, by consistently using the type alias throughout the code. I opted to use the newer C++ style `using` alias, instead of the older `typedef` alias.

Contains a number of other refactorings to clean up the code. In particular, replacing manual memory management with `std::vector`.

Edit: My quick and dirty unit test shows these changes do not impact decompression output.

### DIFF
--- a/src/Archives/AdaptiveHuffmanTree.cpp
+++ b/src/Archives/AdaptiveHuffmanTree.cpp
@@ -172,10 +172,6 @@ namespace Archives
 	// **NOTE**: I may change the bit ordering! (Make that I WILL change the bit ordering)
 	int AdaptiveHuffmanTree::GetEncodedBitString(int code, int &bitString)
 	{
-		int curNodeIndex;
-		int bitCount;
-		bool bBit;
-
 		// Make sure the code is in range
 		if (code >= m_TerminalNodeCount)
 		{
@@ -183,15 +179,16 @@ namespace Archives
 		}
 
 		// Get the node containing the given code
-		curNodeIndex = m_Parent[code];
+		NodeIndex curNodeIndex = m_Parent[code];
 
 		// Record the path to the root
 		bitString = 0;
+		int bitCount = 0;
 		while (curNodeIndex != m_RootNodeIndex)
 		{
-			bBit = curNodeIndex & 0x01;	// Get the direction from parent to current node
+			bool bBit = curNodeIndex & 0x01;	// Get the direction from parent to current node
 			bitCount++;
-			bitString = (bitString << 1) + bBit;// Pack the bit into the returned string
+			bitString = (bitString << 1) | bBit;// Pack the bit into the returned string
 		}
 
 		return bitCount;					// Return number of bits in path from root to node
--- a/src/Archives/AdaptiveHuffmanTree.cpp
+++ b/src/Archives/AdaptiveHuffmanTree.cpp
@@ -4,33 +4,33 @@
 
 namespace Archives
 {
-	// Creates an (adaptive) Huffman tree with numTerminalNodes at the bottom.
-	AdaptiveHuffmanTree::AdaptiveHuffmanTree(NodeType numTerminalNodes) :
+	// Creates an (adaptive) Huffman tree with terminalNodeCount at the bottom.
+	AdaptiveHuffmanTree::AdaptiveHuffmanTree(NodeType terminalNodeCount) :
 		// Initialize tree properties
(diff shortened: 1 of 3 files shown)


src/Archives/AdaptiveHuffmanTree.cpp
```diff
@@ -1,36 +1,32 @@
...
 
 namespace Archives
 {
 	// Creates an (adaptive) Huffman tree with numTerminalNodes at the bottom.
-	AdaptiveHuffmanTree::AdaptiveHuffmanTree(unsigned short numTerminalNodes)
-	{
-		unsigned short i;
-		unsigned short left;
-
+	AdaptiveHuffmanTree::AdaptiveHuffmanTree(NodeType numTerminalNodes) :
```
Brett208: What do you think about updating this variable name to terminalNodeCount?

All the class level variables could be initialized using constructor initialization list. I'm not sure if there is a preference one way or the other here. From my limited point of view, I think it is better to use the the initialization list?

src/Archives/AdaptiveHuffmanTree.cpp
```diff
@@ -146,14 +122,22 @@ namespace Archives
 
 
 
+	// Raise exception if nodeIndex is out of range
+	void AdaptiveHuffmanTree::VerifyValidNodeIndex(NodeIndex nodeIndex)
+	{
+		// Check that the nodeIndex is in range
+		if (nodeIndex >= m_NumNodes)
+		{
+			throw std::runtime_error("Index out of range");
```
Brett208: I like moving the repeated check and error throw into a separate function. If this error is raised by OP2Archive, the message would be fairly vague. Maybe we could include some more detail to help the user out. Something like "During data compression or decompression, an Adaptive Huffman Tree encountered an out of range Node Index"?

src/Archives/AdaptiveHuffmanTree.h
```diff
@@ -29,20 +32,21 @@ namespace Archives
 		// code and places the path in bitString
 
 	private:
-		void SwapNodes(USHORT node1, USHORT node2);
+		void VerifyValidNodeIndex(NodeIndex nodeIndex);
+		void SwapNodes(NodeIndex nodeIndex1, NodeIndex nodeIndex2);
 
+		// Tree properties
+		NodeType m_NumTerminalNodes;
+		NodeType m_NumNodes;
```
Brett208: I would prefer the name m_NodeCount here as well. Although not a big deal either way.

src/Archives/AdaptiveHuffmanTree.cpp
DanRStevens: I prefer not using initialization lists, except where required. They are order dependent in non-intuitive ways. For instance, you may notice how that commit re-ordered the declaration of member variables to match initialization order.

At any rate, all member variables are currently initialized in the constructor initializer list. Or at least, as far as they can be. The vectors have space allocated in the initializer list. Initializing the values of the vectors is too complicated to be done in an initializer, so they must be done in the body of the constructor.

src/Archives/AdaptiveHuffmanTree.h
DanRStevens: Alright, I'll update the names. On a related matter, I think I would prefer dropping the "m_" prefix as well.

Brett208:
Changes look good. I had finally become used to the m_ prefix although I think it is probably better removed.

I compiled the changes for both x86 and x64 with no warnings or errors. It appeared to extract the compressed sheets.vol as expected.

-Brett

DanRStevens:
Alright, I was tempted to add tests before doing this merge, partly as a way to verify the refactoring both before and after. I think I'll settle for a few simple tests with this merge though. If I'm not careful, I might get bogged down writing tests that lead to more refactoring.

PR #156, opened 2018-08-26
Add file reader tests

Brett208:
First attempt at adding some basic FileReader checks.

The Bit Twiddling change is for @DanRStevens to review! I think this is a legitimate check to prevent seeking before start of stream bounds?

-Brett


test/Streams/FileReader.test.cpp
```diff
@@ -0,0 +1,44 @@
+#include "Streams/FileReader.h"
+#include <gtest/gtest.h>
+#include <array>
+
+
+TEST(FileReaderTest, ZeroSizeStreamHasSafeOperations) {
+	Stream::FileReader stream("Streams/EmptyStream.txt");
```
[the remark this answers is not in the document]
Brett208: Will update shortly

Brett208:
See first draft of interface testing. This uses GoogleTest Typed Tests as defined here https://github.com/google/googletest/blob/master/googletest/docs/advanced.md. The sample here https://github.com/google/googletest/blob/master/googletest/samples/sample6_unittest.cc was very helpful.

I'm not good at template code, so hopefully everything is mostly correct??

It appears to test the code actually using the interface (Creating a MemoryReader and then testing it by calling it as a SeekableReader).

-Brett

test/Streams/SeekableReader.test.cpp
```diff
@@ -0,0 +1,65 @@
...
+#include <array>
+
+template <class T>
+Stream::SeekableReader* CreateSeekableReader();
+
+template <>
+Stream::SeekableReader* CreateSeekableReader<Stream::MemoryReader>() {
+	std::array<char, 5> buffer{ 't', 'e', 's', 't', '!' };
+	return new Stream::MemoryReader(buffer.data(), buffer.size() * sizeof(char));
+}
```
DanRStevens: This should be added to MemoryReader.test.cpp.

test/Streams/SeekableReader.test.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+template <>
+Stream::SeekableReader* CreateSeekableReader<Stream::FileReader>() {
+	return new Stream::FileReader("Streams/SimpleStream.txt");
+}
```
DanRStevens: This should be added to FileReader.test.cpp.

test/Streams/SeekableReader.test.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+template <>
+Stream::SeekableReader* CreateSeekableReader<Stream::FileSliceReader>() {
+	Stream::FileReader fileReader("Streams/SimpleStream.txt");
+	return &fileReader.Slice(5);
+}
```
DanRStevens: This should be added to FileSliceReader.test.cpp.

test/Streams/SeekableReader.test.cpp
```diff
@@ -0,0 +1,65 @@
...
+template <class T>
+class SimpleSeekableReader : public testing::Test {
+protected:
+	// The ctor calls the factory function to create a reader implemented by T
+	SimpleSeekableReader() : seekableReader(CreateSeekableReader<T>()) {}
+
+	virtual ~SimpleSeekableReader() { delete seekableReader; }
+
+	Stream::SeekableReader* const seekableReader;
+	const unsigned int size = 5;
```
DanRStevens: This is probably too detailed for a generic test fixture. It may not be feasible for some valid streams to be constructed of length 5.

One possible problem example is a stream which produces an infinite sequence of `0`s. It won't have a finite length. Though that might also make it not a seekable stream, in that it doesn't have a well defined `Length()` result.

A modified example is a stream that produces pages of `0`s, for some fixed size.

Another possible example, is a duplicator stream, which doubles everything read from it. It would necessarily have an even length.

Other odd cases, might be testing network sockets. You could do it, but you'd have to create both ends of the socket to feed it.

test/Streams/SeekableReader.test.cpp
```diff
@@ -0,0 +1,65 @@
...
+};
+
+using testing::Types;
+
+typedef Types<Stream::MemoryReader, Stream::FileReader> SeekableStreamImplementations;
```
DanRStevens: This requires seeing the definitions of all stream types at the same time. That's maybe not the best design, especially since library users could define their own streams, but might still want to make use of this test suite to verify their own implementations.

I think this is where `REGISTER_TYPED_TEST_CASE_P` comes in handy. We can collect all the tests together into a test suit, and let them instantiate it for individual types in their own test files using `INSTANTIATE_TYPED_TEST_CASE_P`. It's slightly more work on our end to manually register all the tests into a test suite with the register macro, but I think the benefits are strongly worth it.

test/Streams/SeekableReader.test.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+TYPED_TEST_CASE(SimpleSeekableReader, SeekableStreamImplementations);
```
DanRStevens: This line would be replaced by individual calls to `INSTANTIATE_TYPED_TEST_CASE_P` in each of the stream .test.cpp files. Each instantiation would pass a single type.

test/Streams/SeekableReader.test.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+typedef Types<Stream::MemoryReader, Stream::FileReader> SeekableStreamImplementations;
+
+TYPED_TEST_CASE(SimpleSeekableReader, SeekableStreamImplementations); 
+
+TYPED_TEST(SimpleSeekableReader, SeekRelativeOutOfBoundsBeginningPreservesPosition) {
+	auto position = this->seekableReader->Position();
+	EXPECT_THROW(this->seekableReader->SeekRelative(-1), std::runtime_error);
+	EXPECT_EQ(position, this->seekableReader->Position());
+}
```
DanRStevens: This test is only appropriate for classes that offer strong exception safety guarantee for the `SeekRelative` method.

test/Streams/SeekableReader.test.cpp
```diff
@@ -0,0 +1,65 @@
...
+// Then use TYPED_TEST(TestCaseName, TestName) to define a typed test,
+// similar to TEST_F.
+TYPED_TEST(SimpleSeekableReader, StreamPositionUpdatesOnReadVer2) {
+	std::array<char, 6> destinationBuffer;
+
+	EXPECT_EQ(this->seekableReader->Position(), 0);
+
+	this->seekableReader->Read(&destinationBuffer[0], 1);
+	EXPECT_EQ(this->seekableReader->Position(), 1);
+}
```
DanRStevens: This test should be fine, assuming we're dealing with a non-empty stream.

We might want to update the buffer size here. Even just using a single `char destinationBuffer` would suffice.

test/Streams/SeekableReader.test.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+TYPED_TEST(SimpleSeekableReader, StreamSizeMatchesInitialization) {
+	EXPECT_EQ(this->seekableReader->Length(), this->size);
+}
```
DanRStevens: This test seems a bit brittle, in that it assumes a test of a given const size can be created. Generalizing it for a variable size might not be testing anything of value, since the size field would likely be initialized using `Length()`.

src/Streams/FileReader.cpp
```diff
@@ -56,7 +57,17 @@ namespace Stream
 		file.seekg(position);
 	}
 
-	void FileReader::SeekRelative(int64_t offset) {
+	void FileReader::SeekRelative(int64_t offset) 
+	{
+		if (offset < 0) {
+			const int mask = offset >> ( sizeof(int64_t) * CHAR_BIT - 1 );
+			const uint64_t offsetAbsValue = (offset ^ mask) - mask;
```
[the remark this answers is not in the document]
DanRStevens: Ahh, `-x` is the same as flip all the bits and add 1, which is the same as XOR with -1 and then add -1. The first part is sign bit replication to the full value. A negative number replicates the sign bit (1) to all bits, which results in -1. A position number replicates the sign bit (0) to all bits, which results in 0. And XOR with 0 does not, and a subtraction of 0 does nothing. Hence it performs `-x` for negative `x`, and is a no-op for positive values. All branch free. Clever.

Though certainly not very clear. I'd be hesitant to use that without wrapping it in a function. It's at a much lower abstraction than the surrounding code.

Mind you, there is already a [std::abs](https://en.cppreference.com/w/cpp/numeric/math/abs) function in the standard library, with overloads for various integer types.

Still, that was fun to work through.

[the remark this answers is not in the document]
Brett208: Good point, well now I feel dumb. Sorry about that.

Besides the odd implementation, do you think this check is a good idea? I think currently the MSVC implementation of ifstream's position will just wrap to a really large positive number if you seek before zero. I think seeking before the start of a file would constitute an error and should throw.

I might also be willing to just ditch allowing a signed relative offset here in favor of an unsigned value. I thought it would be nice to allow others to seek backwards, but it seems to cause a lot of trouble. Removing the ability to use a signed value would remove the check altogether.

-Brett

test/Streams/SeekableReader.test.cpp
Brett208: This takes some overhead by creating SeekableReader.test.h. I just tried and it seems to work.

test/Streams/SeekableReader.test.cpp
Brett208: It is hard to write these tests when we have different ideas on how the interface will be extended and the ideas are not documented anywhere. I thought a network socket would not be a SeekableReader but just a Stream::Reader.

I can separate this back out into three individual tests.

test/Streams/SeekableReader.test.cpp
Brett208: I would like to point out this would still be extendable if the user was to fork the library and just add their implementation to the list of types. I don't have any issues with trying the REGISTER_TYPED_TEST_CASE_P implementation although I haven't learned enough to try it yet.

test/Streams/SeekableReader.test.cpp
Brett208: Good call on simplifying the code. It didn't occur to me to reduce the size of the buffer.

This or another test needs to test that the expected character is actually read as well.

test/Streams/SeekableReader.test.cpp
Brett208: I guess it can be split into individual tests between the existing seekableStream readers again. Maybe the interface needs to be more specific on what Length means so tests like this are not written against it???

Brett208:
The defined constructor for the MemoryReader is broken since the buffer is deleted after passing into MemoryReader. FileSliceReader's test will fail if added to the test suite due to seeking before 0 being unhandled. Ran out of time, so sorry if this explanation is a bit rough.

-Brett

src/Streams/FileReader.cpp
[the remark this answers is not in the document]
DanRStevens: We should bounds check things.

I don't think using an `unsigned` value really avoids the problem. Both signed and unsigned values have corresponding equivalent bit representations, which will produce the same result after the addition. Or at least that's true where signed values have well defined wrap around, which the C++ standard doesn't guarantee, but it nevertheless common.

That is, a `SeekRelative(-1)` should produce the same offset as a `SeekRelative(std::numeric_limits<uint64_t>::max - 1)`. At least that's true in the absence of error checks to prevent it. I suppose with the assumption of positive values, that would be considered an invalid offset, and rejected as too large.

I kind of wanted to stick with `unsigned` values for exactly this reason, and also because they have well defined wrap around. Plus, to seek to the end of a really long stream, you might actually need to full range of positive values. In practice we're unlikely to encounter a stream that long though. And really, a `SeekRelative` is probably expected to allow for negative values. I think it's more correct from an API standpoint to allow for negative offset seeks.

Unless of course you want a separate method to allow seeking backward? Hmm, that might actually make sense. It's probably unlikely you'd want a variable seek distance that could go either forward or backward. You'd probably choose ahead of time to either go forward or backward, or just do an absolute seek. Maybe we should make it part of the method name. That also solves the range issue when you need to do a seek all the way across a long stream. I think this is similar to why I wanted separate methods for `Seek` and `SeekRelative`, rather than a parameter to control the initial seek point (which inevitably ends up being a hardcoded constant most of the time).

test/Streams/SeekableReader.test.cpp
DanRStevens: You have a point, sockets generally aren't seekable. And yes, the other examples are a little bit esoteric to the point of perhaps not being useful. I suppose this is one of the reasons for testing though, in that writing good tests really forces you to tease out these little details.

test/Streams/SeekableReader.test.cpp
DanRStevens: Ahh, because such an edit requires forking the library. It shouldn't be necessary to fork a library to extended the classes and interfaces it provides. This is much like how designing a new Iterator doesn't and shouldn't require forking the standard C++ library.

For a more concrete motivating example, suppose someday the Stream section is separated out to it's own library. Now suppose the VOL decompression code was rewritten to conform to the Stream interface. Where would decompression code go for VOL files? Probably not in the Stream library. It would probably remain with the rest of the VOL processing code, which would import the Stream code. So how then are these decompression streams tested? Hopefully not by editing the test code within the Stream library, but by importing the tests from the Stream library into the test project for the VOL processing code.

This is what `REGISTER_TYPED_TEST_CASE_P` allows us to do.

test/Streams/SeekableReader.test.cpp
DanRStevens: I guess I'm not sure there is a meaningful test here. If we test something like this, it might be more part of a data driven test, such as conformance tests to ensure the decompression matches what Outpost 2 expects. We should try to keep such testing separate from other more generic tests which define what it means to be a stream.

test/Streams/SeekableReader.test.cpp
DanRStevens: Small thought here, but we could potentially use a list of types for this test. It could test reading in a bunch of standard primitive types, or maybe even a struct or two. The `1`s could be replaced with `sizeof(destinationBuffer)`. That could guard against stupid mistakes, such as always incrementing the position by 1 after a successful read.

Not a high priority suggestion though. Mostly if you want to make use of the new found knowledge for writing type parameterized tests.

DanRStevens:
Ok, I've taken a look at this more thoroughly now. I played around and got things to compile, though tests are still failing.

It seems I missed a bit of detail in my earlier reading. I'm going to update the test issue with some of my findings and recommendations. I'll then try to clean up what I've done here into easy to follow steps.

DanRStevens:
Ok, I think this is the general structure we are looking for. I haven't addressed the specifics of the tests themselves. The code is compiling. The tests are running.

Side note: The CI servers aren't currently running the tests. Hence they may still show as passing, even though the tests are failing. We'll get to that.

As for the tests, I'm currently seeing 10 failures on Linux. They are all the same, and are kind of a silly, in that the test data files "EmptyFile.txt" and "SimpleStream.txt" are not found. The reason likely being that I'm running the test executable from a different directory, and the tests use relative paths.

Perhaps we should generate the test files in the system temp folder, rather than store them in the repository. Mind you, that adds the complication of finding a platform specific temp folder. It's tempting to try and use the executable path as a base, but that doesn't work since the executable file could be placed anywhere relative to the source it was compiled from. Similarly, using the `__FILE__` predefined macro doesn't work because the current working directory is unrelated to the source root folder. In fact, the source might not even be available to the compiled test suite after it's been compiled. Though in practice, it should be a reasonable assumption that the test source folder will be available somewhere for the test executable to examine.

test/Streams/SeekableReader.test.cpp
Brett208: All sounds good. I just felt like the GoogleTest documentation was a little draconian in saying that only the second method was suitable for extending outside of the library.

test/Streams/SeekableReader.test.cpp
Brett208: I think this is a valid test for MemoryReader, FileReader, and FileSliceReader.

Brett208:
I updated the Visual Studio project file by removing the deleted .cpp file and everything worked properly on Visual Studio. Nice work getting the tests switched over to the 'more complex' format.

I think that I would prefer leaving the test files in the test source code as it keeps things a bit more simple. I could look at updating the Visual Studio environment starting location to be the same directory as the Linux build, which would solve the relative directory issues. I don't know how difficult it would be to update the makefile test portion to use the test directory as it's default directory?

-Brett

Brett208:
Okay, I just updated the Visual Studio environment to start at OP2Utility instead of the test directory. Unfortunately, this is considered a user setting and is not stored in the repository, so others who download the library would have to manually set this property to get the tests to pass. It is an easy settings change though.

I'd like to figure out a better permanent project setting at some point, but this should get it compiling on both Linux and Windows for the time being.

test/Streams/SeekableReader.test.cpp
DanRStevens: I think it's about technical limitations. With the first method, the list of types must be specified before the tests. Hence there is no way to import the tests from the library and then specify the types later. When playing around with the original typed test code, I tried moving the type instantiations to after the tests and it was considered an error. There's like a technical reason for that buried in the macros.

With the type-parameterized tests, I think they have problems wrapping all necessary parts into a single macro. Hence the separate declaration of the test case above, and then manually registering all tests in the test case below. It makes it less pleasant to define tests that way. You lose out on the auto registration of tests. I suppose that's why they offer both methods, so you don't pay the cost of the extra setup when you don't need it.

----

Maybe there is a way with the first method to define the types first, and then import the library. Though that's an awkward backwards way of doing things, which seems non-standard and surprising. Personally, I like the second method from a library user's standpoint. The hard stuff has been taken care of in the library, so it's natural and easy for library users to use and extend.

DanRStevens:
Ok, I think what we can do then is to update the Linux makefile so `make check` runs the test from the "test" subfolder. Then you should be able to remove the Visual Studio user setting, and tests should work with default settings anywhere.

DanRStevens:
Ok, we've solved the big issues on this branch. We now have a good template structure for new tests. The tests should be passing on both Linux and Windows (awaiting verification) with default project settings. We've also set an expected test run folder, so we have a clear way to access test data files.

Should we move the test data files to a subfolder? Right now the folder is an arbitrary mingling of test and data files, which makes it harder to see what and where the tests are.

----

I'd still like to address some issues of test quality, though that's a deeper issue and may take some learning. We could defer that refactoring to future branches. There's plenty of value in this branch that I'd be happy to merge now.

As an example of future refactoring, we should consider moving `ZeroSizeStreamHasSafeOperations` to a header file and making it type parameterized. It would need to go in a separate test case, as it's written specifically for streams of size 0. Alternatively, we could split it into components that are valid elsewhere. In particular, each section relies on different assumptions:
```cpp
TEST_F(EmptyMemoryStreamReader, ZeroSizeStreamHasSafeOperations) {
	// Length and position (expects a seekable stream of length 0)
	EXPECT_EQ(0, stream.Length());
	EXPECT_EQ(0, stream.Position());
	// Seek to current position (should not cause error) (expects a seekable stream)
	ASSERT_NO_THROW(stream.Seek(0));
	ASSERT_NO_THROW(stream.SeekRelative(0));
	// Read 0 bytes is safe (valid for any stream, of any length)
	ASSERT_NO_THROW(stream.Read(nullptr, 0));
	EXPECT_EQ(0, stream.ReadPartial(nullptr, 0));
}
```

src/Streams/FileReader.cpp
```diff
@@ -60,8 +60,7 @@ namespace Stream
 	void FileReader::SeekRelative(int64_t offset) 
 	{
 		if (offset < 0) {
-			const int mask = offset >> ( sizeof(int64_t) * CHAR_BIT - 1 );
-			const uint64_t offsetAbsValue = (offset ^ mask) - mask;
+			const uint64_t offsetAbsValue = -offset;
```
Brett208: You removed the bit twiddling...

Brett208:
I pushed all the test data into a directory called data. I'm going to merge the branch now so we can start new branches to continue other endeavors. Everything is compiling on Windows. No problem putting the data in a modified directory from this, but I don't want to keep fiddling with this branch.

I think you have a very specific idea of what you want for the stream::reader test refinement, so I would probably leave that effort to you. 

I'd like to start creating a couple of simple tests for other classes in the library. This would populate the test source code files so they exist in sort of skeleton form as we currently only have the stream::reader covered at all.

-Brett

src/Streams/FileReader.cpp
DanRStevens: :scream: 

Hey..., what does this "Resolve conversation" button do? I've never noticed it before....

DanRStevens:
Alright, I agree. Time to merge this and move on. We can make improvements in future branches.

I noticed the Windows Pull Request build has failed. That would seem to indicate there will be a problem after merging? Looks related to the file rename from moving test data into a subfolder.

Hmm, indeed. The merge commit also failed.

Issue #155, opened 2018-08-25
CI should build and run tests


DanRStevens:
The CI jobs should be updated to build and run the tests. The Linux jobs should be easy, with a small tweak to the make command used to build the project. It's already setup so a `make check` will build and run tests, which depend on the project itself, and so will compile and link the project library automatically.

For AppVeyor, we may want to build a config file to be added to version control. Initially the AppVeyor job was setup using the web interface, with no files added to version control. That made for an easy setup process, but it less transparent than a config file. Additionally, a config file would make it easier for a project fork to setup their own CI solutions.

Finally, getting the tests to run on AppVeyor means I'd have an easy way to investigate test failures or warnings that only affect a Windows environment.

DanRStevens:
I took a quick shot at this and then quickly realized Google Test wasn't installed by default in the Linux environments. Further, because of the refusal to ship binaries with the Google Test library, it makes it kind of a pain to install the dependency.

CircleCI may be easiest, by using a custom Dockerfile, based on the currently used `gcc:8` image, but with extra setup to install and compile Google Test. The extra layer needed for the Docker image should be fairly minimal, so it should add much to image download and launch time.

For TravisCI, this seems like more of a pain. TravisCI has no ability to customize the Docker image. If we use the faster booting Docker images, there is no root access. They have a special module to allow installing `apt` packages without root access, so that part is fine, but there is no ability to globally install Google Test libraries afterwards. Hence we would need to compile and store the binaries elsewhere. That could work, in that we could compile them to the project's build folder, and access the libraries from there, it's just a bit awkward, and might require some updates to the build scripts or environment to account for the moved library location.

Alternatively, we could go for the full VM option, rather than the Docker image. That would allow for root access, which enables compiling and installing global libraries for Google Test. Unfortunately, using a full VM adds 30 seconds of startup time to each build, and TravisCI builds are already slow.

Brett208:
I looked up pricing on TravisCI, CircleCI, and AppVeyor. Unfortunately, the absolute cheapest package is $30 / month on AppVeyor. TravisCI and CircleCI are both a fair amount more expensive than this. While I'm not opposed at spending some money if it helps the build process, $360+ / year is a fair amount more than I wanted to see. Too bad there isn't some room for something better than free but less than $100 / year. And I guess this would only get us Windows Builds with AppVeyor and not even Linux.

I think if we stick with the free packages, we should just bite the time bullet and expect long lead build times.

-Brett

[...]

PR #159, opened 2018-08-28
Add AppVeyor configuration file

DanRStevens:
This gets both the core OP2Utility library project and the Google Test based test project to build on AppVeyor. Additionally, it adds `nuget restore` to the build process, so it will download Google Test before building. Finally, it runs the compiled test project.

Currently there are 19 build warnings for the test project:

> warning C4996: 'std::tr1': warning STL4002: The non-Standard std::tr1 namespace and TR1-only machinery are deprecated and will be REMOVED. You can define _SILENCE_TR1_NAMESPACE_DEPRECATION_WARNING to acknowledge that you have received this warning.

The [Microsoft TestAdapterForGoogleTest Issue 119](https://github.com/Microsoft/TestAdapterForGoogleTest/issues/119) has some details on this. It seems this has been fixed in the Google Test master branch, but not released as a Nuget package. The solution for Nuget mentioned there was to define `_SILENCE_TR1_NAMESPACE_DEPRECATION_WARNING` in the project configuration under preprocessor settings. For an example see [Microsoft TestAdapterForGoogleTest PR 115](https://github.com/Microsoft/TestAdapterForGoogleTest/pull/115/files).

Additionally there is a build failure for x64:

> (Link target) -> 
  LINK : fatal error LNK1246: '/SAFESEH' not compatible with 'x64' target machine; link without '/SAFESEH' [C:\projects\op2utility\test\OP2UtilityTest.vcxproj]

This also looks like a project configuration item.

@Brett208: Could you please update the Visual Studio project settings. Hopefully that's all it takes to address these warnings and errors.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -6,3 +6,4 @@ platform:
 before_build: nuget restore
 build:
   project: OP2Utility.sln
+test_script: '%APPVEYOR_BUILD_FOLDER%\%CONFIGURATION%\OP2UtilityTest.exe'
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,5 +1,7 @@
 image: Visual Studio 2017
-configuration: Release
+configuration:
+  - Debug 
+  - Release
 platform:
   - x86
   - x64
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,0 +1,8 @@
+image: Visual Studio 2017
+configuration: Release
+platform:
+  - x86
+  - x64
+before_build: nuget restore
+build:
+  project: OP2Utility.sln
--- a/test/OP2UtilityTest.vcxproj
+++ b/test/OP2UtilityTest.vcxproj
@@ -33,12 +33,12 @@
   <ImportGroup Label="PropertySheets" />
   <PropertyGroup Label="UserMacros" />
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">
-    <OutDir>$(SolutionDir)$(Platform)\$(Configuration)\</OutDir>
-    <IntDir>$(Platform)\$(Configuration)\</IntDir>
+    <IntDir>x86\$(Configuration)\</IntDir>
+    <OutDir>$(SolutionDir)x86\$(Configuration)\</OutDir>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">
-    <IntDir>$(Platform)\$(Configuration)\</IntDir>
-    <OutDir>$(SolutionDir)$(Platform)\$(Configuration)\</OutDir>
+    <IntDir>x86\$(Configuration)\</IntDir>
+    <OutDir>$(SolutionDir)x86\$(Configuration)\</OutDir>
   </PropertyGroup>
   <ItemGroup>
     <ClCompile Include="Streams\MemoryStreamReader.test.cpp" />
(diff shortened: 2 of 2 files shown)


Brett208:
@DanRStevens, nice research discovering those issues as usual.

The safe exception handling was a mistake on my part. I wasn't aware that this property only applies to x86 builds and set it erronously on the x64 release build.

I added the TR1 preprossesor setting which prevented the warnings from showing locally on my computer.

I also updated appveyor to build both a release and debug version of the code. I think it is important to make sure both configurations are working.

Unfortunately, something is wrong with appveyor still and I'm not sure what the problem is.

DanRStevens:
Alright, thank you for the updates. That's wonderful.

I see what the error is. I tried to run the compiled test executable, but it seems I got the path wrong for x64 builds. Should hopefully be an easy fix once I get a moment.

DanRStevens:
I just noticed there is a difference in build path patterns. There is no platform component to the path for x86 builds, but there is for x64.

Confiuration x86 (pass):

> Debug\OP2UtilityTest.exe
> Release\OP2UtilityTest.exe

Configuration x64 (fail):

> x64\Debug\OP2UtilityTest.exe
> x64\Release\OP2UtilityTest.exe

The pattern used was: `'%APPVEYOR_BUILD_FOLDER%\%CONFIGURATION%\OP2UtilityTest.exe'`
I tried updating to: `'%APPVEYOR_BUILD_FOLDER%\%PLATFORM%\%CONFIGURATION%\OP2UtilityTest.exe'`

Naturally though, that just reversed the failures, where now the x86 builds fail, and the x64 builds pass.

Perhaps we can update project settings to be more consistent with the build folder names. This is primarily for the test project.

DanRStevens:
I believe the VS project setting will be under Properties -> General -> Output Directory. The default is supposedly `$(SolutionDir)$(Configuration)\`. Maybe we can use `$(SolutionDir)$(Platform)\$(Configuration)\`, though I have a feeling that will produce "Win32" rather than "x86", as per what is listed in the project file.

In case this helps:
[Visual Studio variables](https://msdn.microsoft.com/en-us/library/c02as0cs.aspx)
[App Veyor variables](https://www.appveyor.com/docs/environment-variables/)

Failing the existence of a convenient environment variable, we could always have a hardcoded path component for the x86 platform settings in the Visual Studio project settings.

Brett208:
I actually like the idea of having a win32 directory and a x64 directory that each contain a debug and release directory. I made the needed changes in OP2UtilityTest, but it didn't fix the problems.

Maybe I need to apply the change in OP2Utility as well? I ran out of time to keep testing though. Could try this tomorrow.

-Brett

DanRStevens:
Thank you for the update. That at least lets me see the expected structure of the project file configuration settings.

The problem looks to be the AppVeyor `%CONFIGURATION%` variable expands to "x86", while the Visual Studio `$(Platform)` variable expands to "Win32". Hence the project is compiling to one path, but the test script expects the executable in a different path:
"Win32\Debug\OP2UtilityTest.exe"  <- What Visual Studio produces
"x86\Debug\OP2UtilityTest.exe"  <- What the AppVeyor test script expects

I updated the Visual Studio project file to use a hardcoded "x86" string in place of the `$(Platform)` variable for the Win32 platform configurations.

DanRStevens:
Ok, if you're fine with the hardcoded project configuration setting, I think this closes off the Windows portion of Issue #155.

Brett208:
Looks good to me. Thanks for running with this and getting it working.

PR #163, opened 2018-08-31
Fix AppVeyor builds

DanRStevens:
For some reason, the current working directory used when runnings tests appears to have changed. This PR adds an explicit command to set the current working directory before running tests.

The original problem may be related to the branch that added the `.appveyor.yml` configuration file, or perhaps due to recent project file changes, I'm not really sure. I didn't see any documentation that addressed the default working directory, nor how MSBuild might have affect the current working directory. It would be nice to know what changed to cause this behaviour. Meanwhile though, explicitly setting the current working directory seems to fix it.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -9,4 +9,6 @@ before_build: nuget restore
 build:
   project: OP2Utility.sln
 test_script:
+  - cd %APPVEYOR_BUILD_FOLDER%\test
   - '%APPVEYOR_BUILD_FOLDER%\%PLATFORM%\%CONFIGURATION%\OP2UtilityTest.exe'
+  - cd %APPVEYOR_BUILD_FOLDER%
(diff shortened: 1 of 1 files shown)


DanRStevens:
Since this fixes broken builds, I'm going to go ahead and merge this one before review. Not fixing this means all other opened branches based on the tip of master will also show failing tests.

Brett208:
Thanks for fixing this. I noticed that appVeyor failed and I should have not merged into source with it failing. I'm also not sure what changed to affect this negatively. 

I can see why people pay to increase the speed that automated testing happens. It takes a long time for all 5 of our tests to complete and it is frustrating to wait for the tests to run.

-Brett

DanRStevens:
Indeed, slow tests are a pain. I have a few ideas to speed things up though, or to have certain failing tests fail faster.

It's also interesting to finally see an example where the branch build passed, but the pull request build failed.

PR #164, opened 2018-08-31
Disable AppVeyor Debug build

DanRStevens:
Double the configurations means double the test runs. Due to limited
parallelism on the free open source plan, this is doubling the test
time.

Currently we have no code that behaves obviously different in Debug
versus Release mode. In particular there are no `#define` or
`#ifdef`/`#ifndef` blocks that depend on Debug or Release settings.

Contrast that with 32-bit versus 64-bit setting, which affect the
`sizeof` operators litered throughout the code base.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,6 +1,5 @@
 image: Visual Studio 2017
 configuration:
-  - Debug 
   - Release
 platform:
   - x86


Brett208:
I didn't realize you could open up AppVeyor and watch the build process move forward real time. That is really interesting. I am seeing 4-4.5 minutes for the PR and the Merge each. I don't think it will do these in parallel, so that is 8-9 minutes to complete both of them. You are correct that this is too long. I can write my PR message and it still not be done building.

Since I routinely program in debug mode, we could skip them as you suggested as I will probably catch these fairly quickly. The Release discrepancies are the ones I'm missing regularly. I typically only test release occasionally. 

I'm okay if you want to pull the debug builds. 

-Brett

DanRStevens:
I had a few additional thoughts.

I was thinking what the differences are between Debug and Release builds. From what I remember, the main difference is debugging information versus optimizations. With optimizations on, certain tranformations complicate the mapping between the generated code and the original source. That can make debugging sessions more difficult to make sense of. Hence it may be easier to debug using the Debug build, and meanwhile the Release build often has the generation of debugging information turned off.

What if we turned on debugging information for the Release build?

I don't know how heavily you rely on the VS debugger to do your work. I rarely if ever used it. That's part of why I just stuck with Release builds. The other reason being I didn't like having a disctinction between Debug and Release. My long forgotten reasoning being it seemed like a bad idea to spend all my time developing and testing in Debug mode, but then shipping an untested Release version.

The other different between Debug and Release, is macro definitions that can be used in `#define` and `#ifdef`/`#ifndef`, but we're not currently making use of that feature. There may be a few other small things too, such as the runtime libraries used, and whether they have debugging information included or optimizations turned on. I went looking for other differences, but didn't find much of significance.

If we turned on debugging information for the Release build, we might not need the Debug build. It might take slighly longer to produce the Release build, but we'd save a lot of time by not having to do a Debug build.

----

With that said, I would like to be producing builds for all configurations. Disabling builds for a particular configuration seems like a temporary fix. Perhaps there are other ways we can optimize the build process to speed things up.

----

For reference, the Linux builds test minimal configurtions. It's only the compiler that is varied there, g++ or Clang. There is no 32-bit versus 64-bit, nor any kind of Debug versus Release. I noticed the Linux build has the "-g" flag to enable debug information, though this information is never used. Meanwhile they are lacking any optimization flags.

I suppose the Linux build settings correspond to the Debug build, with no Release build. Though in terms of distribution and licensing restrictions on Windows, I think it would need to be the other direction there. You can redistribute the Release runtimes, but you have no license to redistribute the Debug runtimes.

PR #165, opened 2018-08-31
Enable caching of Nuget packages on AppVeyor

DanRStevens:
Enabling caching of the Nuget "packages" folder should result in less network traffic, and hopefully faster builds.


Brett208:
This seems like a really good idea. Thank you for looking into it.

-Brett

PR #166, opened 2018-08-31
Add Stream::Reader tests

DanRStevens:


Brett208:
I added the new test header to the Visual Studio project. After this, everything compiled and without warning or error and tests ran fine. It looks like a clean start to me. 

We should probably eliminate one version of ZeroSizeStreamHasSafeOperations Inside the MemoryReader test code at this point. I wasn't sure which version you favored, so I left it alone for now.

I'm surprised how complicated writing even basic tests look with all the macros and templates involved. :/

-Brett

DanRStevens:
Ahh, thank you for the update. I forgot about updating the Visual Studio project file.

We should definitely eliminate some of the duplicate tests. I'll address that in a separate issue. Related, is being able to run the same tests on multiple instances of the same type of stream. Some of those tests, such a zero byte reads are valid for any stream instance, of any type, whether they are empty, or have data. Currently that isn't handled well by the overly verbose fixture and factory function setup code. That's something I may need to think about a bit more.

You're right though, in that the tests have a disappointing amount of boilerplate setup code.

PR #168, opened 2018-09-01
Basic map reader tests

Brett208:
I started writing basic tests for MapReader. Several topics worth discussing immediately emerged, so I wanted to stop where I was at and discuss.

 - We need to determine a strategy for testing exception handling. Currently, hitting any runtime_exception will pass when using ASSERT_THROW. However, almost all of our exceptions are runtime_exceptions, so there is no guarantee you are hitting the correct one. 

    - We could check the exception message to ensure it matches. This would require doubling the text of the exception message up in both places unless we somehow give the test fixture access to the message from the class throwing somehow. Maybe a protected variable containing the message, but I don't like this idea.

 - MapReader will fail when attempting to read an empty file because it will eventually try to read something improperly and throw an exception. This will fail due to a non-runtime_exception from the std::library (I think) and not our code. I was thinking it would be better to test if the size of the stream is greater than zero or not and fail then instead of letting the reader continue reading garbage to eventually fail. However, that would imply we would want to test the remaining size of the stream before every read. This seems to get too cumbersome by bloating the source code and possible performance implications from constantly reading the and checking the length. Maybe the performance hit is small enough not to worry about? Open for suggestions here.

 - I opened a separate issue to discuss if we want to force an exception if the file extension if unexpected. We should probably discuss this there.

 - I would like to test passing an incorrect Stream::MemoryReader and Stream::FileReader into MapReader. We will have basic MemoryReader and FileReader test fixtures created in the Stream code for testing the stream code. Would it make sense to try and use these fixtures in the MapReader code as well? If so, I need to do some reading to figure out how to incorporate these test fixtures sense they are instantiated with constructors called from the templates to allow proper testing of the interface itself (If this doesn't make sense it is because I'm not sure what I'm talking about yet.)

Thanks,
Brett

### DIFF
--- a/src/Maps/MapType.h
+++ b/src/Maps/MapType.h
@@ -2,5 +2,5 @@
 
 enum class MapType {
 	Map,
-	Save
+	SavedGame
 };
--- a/src/Maps/MapType.h
+++ b/src/Maps/MapType.h
@@ -1,0 +1,6 @@
+#pragma once
+
+enum class MapType {
+	Map,
+	Save
+};
--- a/src/Maps/MapType.h
+++ b/src/Maps/MapType.h
@@ -1,6 +1,0 @@
-#pragma once
-
-enum class MapType {
-	Map,
-	SavedGame
-};
--- a/test/Maps/MapReader.test.cpp
+++ b/test/Maps/MapReader.test.cpp
@@ -5,14 +5,14 @@
 
 TEST(MapReader, MissingFile) {
 	EXPECT_THROW(MapReader::Read("MissingFile.map", MapType::Map), std::runtime_error);
-	EXPECT_THROW(MapReader::Read("MissingFile.op2", MapType::Save), std::runtime_error);
+	EXPECT_THROW(MapReader::Read("MissingFile.op2", MapType::SavedGame), std::runtime_error);
 
 	// Check if filename is an empty string
 	EXPECT_THROW(MapReader::Read(std::string(""), MapType::Map), std::runtime_error);
-	EXPECT_THROW(MapReader::Read(std::string(""), MapType::Save), std::runtime_error);
+	EXPECT_THROW(MapReader::Read(std::string(""), MapType::SavedGame), std::runtime_error);
 }
 
 TEST(MapReader, EmptyFile) {
 	EXPECT_THROW(MapReader::Read("Maps/data/EmptyMap.map", MapType::Map), std::runtime_error);
-	EXPECT_THROW(MapReader::Read("Maps/data/EmptySave.OP2", MapType::Save), std::runtime_error);
+	EXPECT_THROW(MapReader::Read("Maps/data/EmptySave.OP2", MapType::SavedGame), std::runtime_error);
 }
--- a/test/Maps/MapReader.test.cpp
+++ b/test/Maps/MapReader.test.cpp
@@ -4,8 +4,8 @@
 #include <string>
 
 TEST(MapReader, EmptyFilename) {
-	EXPECT_THROW(MapReader::Read(std::string(), MapType::Map), std::runtime_error);
-	EXPECT_THROW(MapReader::Read(std::string(), MapType::Save), std::runtime_error);
+	EXPECT_THROW(MapReader::Read(std::string(""), MapType::Map), std::runtime_error);
+	EXPECT_THROW(MapReader::Read(std::string(""), MapType::Save), std::runtime_error);
 }
 
 TEST(MapReader, MissingFile) {
--- a/test/Maps/MapReader.test.cpp
+++ b/test/Maps/MapReader.test.cpp
@@ -17,8 +17,3 @@ TEST(MapReader, EmptyFile) {
 	EXPECT_THROW(MapReader::Read("Maps/data/EmptyMap.map", MapType::Map), std::runtime_error);
 	EXPECT_THROW(MapReader::Read("Maps/data/EmptySave.OP2", MapType::Save), std::runtime_error);
 }
-
-//TEST(MapReader, WrongExtension) {
-//	EXPECT_THROW(MapReader::Read("Maps/data/WrongExtension.txt", MapType::Map), std::runtime_error);
-//	EXPECT_THROW(MapReader::Read("Maps/data/WrongExtension.txt", MapType::Save), std::runtime_error);
-//}
--- a/test/Maps/MapReader.test.cpp
+++ b/test/Maps/MapReader.test.cpp
@@ -1,23 +1,24 @@
 #include "Maps/MapReader.h"
+#include "Maps/MapType.h"
 #include "gtest/gtest.h"
 #include <string>
 
 TEST(MapReader, EmptyFilename) {
-	EXPECT_THROW(MapReader::Read(std::string(), false), std::runtime_error);
-	EXPECT_THROW(MapReader::Read(std::string(), true), std::runtime_error);
+	EXPECT_THROW(MapReader::Read(std::string(), MapType::Map), std::runtime_error);
+	EXPECT_THROW(MapReader::Read(std::string(), MapType::Save), std::runtime_error);
 }
 
 TEST(MapReader, MissingFile) {
-	EXPECT_THROW(MapReader::Read("MissingFile.map", false), std::runtime_error);
-	EXPECT_THROW(MapReader::Read("MissingFile.op2", true), std::runtime_error);
+	EXPECT_THROW(MapReader::Read("MissingFile.map", MapType::Map), std::runtime_error);
+	EXPECT_THROW(MapReader::Read("MissingFile.op2", MapType::Save), std::runtime_error);
 }
 
 //TEST(MapReader, EmptyFile) {
-//	EXPECT_THROW(MapReader::Read("Maps/data/EmptyMap.map", false), std::runtime_error);
-//	EXPECT_THROW(MapReader::Read("Maps/data/EmptySave.OP2", true), std::runtime_error);
+//	EXPECT_THROW(MapReader::Read("Maps/data/EmptyMap.map", MapType::Map), std::runtime_error);
+//	EXPECT_THROW(MapReader::Read("Maps/data/EmptySave.OP2", MapType::Save), std::runtime_error);
 //}
 
 //TEST(MapReader, WrongExtension) {
-//	EXPECT_THROW(MapReader::Read("Maps/data/WrongExtension.txt", false), std::runtime_error);
-//	EXPECT_THROW(MapReader::Read("Maps/data/WrongExtension.txt", true), std::runtime_error);
+//	EXPECT_THROW(MapReader::Read("Maps/data/WrongExtension.txt", MapType::Map), std::runtime_error);
+//	EXPECT_THROW(MapReader::Read("Maps/data/WrongExtension.txt", MapType::Save), std::runtime_error);
 //}
--- a/test/Maps/MapReader.test.cpp
+++ b/test/Maps/MapReader.test.cpp
@@ -1,18 +1,17 @@
 #include "Maps/MapReader.h"
-#include "Maps/MapType.h"
 #include <gtest/gtest.h>
 #include <string>
 
 TEST(MapReader, MissingFile) {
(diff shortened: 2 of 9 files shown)


src/Maps/MapType.h
```diff
@@ -0,0 +1,6 @@
+#pragma once
+
+enum class MapType {
+	Map,
+	Save
+};
```
DanRStevens: I believe when an `enum` is this closely tied to a class, it is appropriate to declare it in the class header file, and probably within the namespace of the class itself, as a public member. Everything you need related to the class is then in one place.

For a really long enum, it may make sense to put it in a separate file. For an enum used by several classes with no clear owner, it definitely makes sense to put it in a separate file. This case I think is optional.

test/Maps/MapReader.test.cpp
```diff
@@ -0,0 +1,24 @@
+#include "Maps/MapReader.h"
+#include "Maps/MapType.h"
+#include "gtest/gtest.h"
```
DanRStevens: We should probably use angle brackets for `<gtest/gtest.h>`. It's not part of our project, nor is it part of our source tree.

test/Maps/MapReader.test.cpp
```diff
@@ -0,0 +1,24 @@
...
+#include <string>
+
+TEST(MapReader, EmptyFilename) {
+	EXPECT_THROW(MapReader::Read(std::string(), MapType::Map), std::runtime_error);
+	EXPECT_THROW(MapReader::Read(std::string(), MapType::Save), std::runtime_error);
+}
```
DanRStevens: This test would probably be more clear with a literal empty string `""`.

test/Maps/MapReader.test.cpp
```diff
@@ -0,0 +1,24 @@
...
+
+TEST(MapReader, MissingFile) {
+	EXPECT_THROW(MapReader::Read("MissingFile.map", MapType::Map), std::runtime_error);
+	EXPECT_THROW(MapReader::Read("MissingFile.op2", MapType::Save), std::runtime_error);
+}
```
DanRStevens: I had to think about this. I wasn't sure if this is really testing the `MapReader`, or if it's actually just testing the underlying `Stream::FileReader` class. I think it does make sense though, since at this level we don't know it's using an underlying `Stream::FileReader` object, and technically that is subject to change. From an API standard, we do want well defined behaviour if a bad filename is passed, and that includes throwing an exception.

Do we need to test both an empty file name, as per the test above, and a missing file name? Both are technically the same, a non-existent file. Or maybe we should test them all together in the same test. Hmm, the choice is not clear to me.

test/Maps/MapReader.test.cpp
```diff
@@ -0,0 +1,24 @@
...
+
+//TEST(MapReader, EmptyFile) {
+//	EXPECT_THROW(MapReader::Read("Maps/data/EmptyMap.map", MapType::Map), std::runtime_error);
+//	EXPECT_THROW(MapReader::Read("Maps/data/EmptySave.OP2", MapType::Save), std::runtime_error);
+//}
```
DanRStevens: Yes, I'm perfectly fine with an empty file test. It should throw.

test/Maps/MapReader.test.cpp
```diff
@@ -0,0 +1,24 @@
...
+
+//TEST(MapReader, WrongExtension) {
+//	EXPECT_THROW(MapReader::Read("Maps/data/WrongExtension.txt", MapType::Map), std::runtime_error);
+//	EXPECT_THROW(MapReader::Read("Maps/data/WrongExtension.txt", MapType::Save), std::runtime_error);
+//}
```
DanRStevens: No, I don't think we should be checking for a bad extension, nor throwing in that case.

DanRStevens:
Good point about any `runtime_exception` satisfying the `ASSERT_THROW`.

I suppose form a testing brittleness standpoint, we don't want to be too specific about the exceptions we get back. In particular, we should avoid testing on the exception messages, since that is purely informational, and not behavioural. From a caller's standpoint, we don't care why something failed, only that it did fail, and there is some information we can display to the user to allow them to make sense of what just happened.

With that said, sometimes a clear and proper error message is part of what you want to test. I've seen some frameworks that scan exception messages using string matchers to find certain keywords. This may be possible using the Google Mock portion of the library.

If you feel the parser should be more specific about the types of errors that are thrown, then that does become a design issue. Especially if you want the caller to be able to take different action based on the type of error thrown. I'm not sure we can really take different action based on why a map failed to load, only that it did fail to load. Hence I'm not sure we need specific exception types to handle the various errors. Or at least, we don't need to promise specific exception types for various errors to callers of the API.

If you want to add more specific exception types in the parsing code, I'm fine with that. It may help debugging if a problem arises. However, I'm not certain such promises should become part of the API. As such, I'm not certain if it makes sense to test for specific exception types, since you are then testing private internal details, rather than the public API.

Though I do get that it helps build confidence that the tests are actually testing what they purport to test.

I think I need to think about this one.

----

Sounds like you've uncovered a bug in the map reading code. In particular, we should be using the exception raising `Stream::Read` method, which will throw if the stream does not contain enough data to complete the read. For an empty file, that should be the case. In particular, the `Stream::Read` API should already be responsible for checking stream size, and raising exceptions for an incomplete read. This design point is specifically to avoid having to liter the calling code with checks as you've mentioned.

Looking at the map reader code, the first read is:
```cpp
streamReader.Read(mapData.header);
```

Hence, we should expect to see an exception from this line when reading from an empty file.

The bug is likely in:
```cpp
void FileReader::ReadImplementation(void* buffer, std::size_t size) {
  file.read(static_cast<char*>(buffer), size);
}
```
You'll notice that we are failing to check for error flags on the `ifstream` object, and not throwing an exception upon attempting to read past the end of the stream.

This failure of the `FileReader` implementation can leave uninitialized data in the buffer we read into, which could be why we are failing in mysterious ways.

----

I think a new test fixture is needed here. I don't think there is a sensible way to re-use existing test fixtures.

You're right though, in that the current object construction method is seriously lacking. We need a better way to construct objects for the purpose of testing. I may need to think about this further.

src/Maps/MapType.h
Brett208: I'm happy either way embedding in MapData or not. I was more worried about trying to find a better name than MapType, which I thought could be improved.

test/Maps/MapReader.test.cpp
Brett208: Good call here.

test/Maps/MapReader.test.cpp
Brett208: Okay, I'll add them in.

test/Maps/MapReader.test.cpp
Brett208: I'm not sure it is necessary. 

While an empty string should also report as a missing file, something could break or not play nicely if using the filename for error reporting or something else where a non-empty string was expected. This shouldn't matter in the current MapReader code though.

I think for an easier test like this, it doesn't really bloat the testing documentation to include it even if it mostly overlaps what is being tested?

test/Maps/MapReader.test.cpp
Brett208: Sounds good, will remove

Brett208:
Good idea on making a ReadMap and ReadSavedGame function, which would probably be a little cleaner than using the enum. I think there is still room for using the enum internal to the MapReader and MapData to distinguish which type is being represented. 

Agreed that there are some sticky point when adding actual map data for testing. I just wanted to get some basic tests up for MapReader so we have a starting point in the future if/when we figure that out.

I think we are basically on the same page when it comes to testing exception handling and the different challenges. Although the difficulty will be the right balance in implementation. Since you solved the underlying Stream issue that caused the problem, we can punt this issue to another time for now.

Thank you for the quick, helpful, and thorough review of the code. I'll take a look at separating out the ReadMap and ReadSavedGame functions and then be ready to merge.

-Brett

src/Maps/MapType.h
DanRStevens: The enum name seems alright. Even if it was embedded, I think it would still be fine. I maybe feel "Save" is a bit short though, and would prefer "SavedGame". The reason being that "Save" is a verb, whereas the other two are nouns.

test/Maps/MapReader.test.cpp
DanRStevens: Ahh, not such a bad idea to consider the error reporting like that. Ok, I'm fine with including both empty strings and non-empty strings for non-existent files.

My preference is still to combine them into a single test method, since they are so similar.

test/Maps/MapReader.test.cpp
DanRStevens: Slight note, but the `std::string` wrapping/conversion is automatic. There is an implicit conversion from C-style strings. The explicit wrapping also begs the question, why is this different from the filenames being passed in the code below?

DanRStevens:
I just took a look at the code, in regards to using `ReadMap` and `ReadSavedGame`. It looks possible to completely remove the enum in terms of parsing the data. However, it may still make sense to tag the data afterwards as being from a map file or a saved game file.

Of course if saved game parsing was ever implemented, a method could determine that attribute by checking for the existence of saved game data.

I'd be fine if the method split was done on another branch.

src/Maps/MapType.h
Brett208: Yeah, save is pretty generic, good call.

test/Maps/MapReader.test.cpp
Brett208: Okay, I'll combine them.

test/Maps/MapReader.test.cpp
Brett208: Well, I can remove the quotes so the c-style string conversion isn't happening again. I thought it was clear enough without the quotes anyways.

test/Maps/MapReader.test.cpp
```diff
@@ -1,19 +1,17 @@
...
-
 TEST(MapReader, MissingFile) {
-	EXPECT_THROW(MapReader::Read("MissingFile.map", MapType::Map), std::runtime_error);
-	EXPECT_THROW(MapReader::Read("MissingFile.op2", MapType::Save), std::runtime_error);
+	EXPECT_THROW(MapReader::ReadMap("MissingFile.map"), std::runtime_error);
+	EXPECT_THROW(MapReader::ReadSavedGame("MissingFile.op2"), std::runtime_error);
+
+	// Check if filename is an empty string
+	EXPECT_THROW(MapReader::ReadMap(std::string("")), std::runtime_error);
+	EXPECT_THROW(MapReader::ReadSavedGame(std::string("")), std::runtime_error);
```
DanRStevens: To clear up any confusion concerning these lines, my intention was to remove the explicit `std::string` wrapper:
```cpp
EXPECT_THROW(MapReader::ReadMap(""), std::runtime_error);
EXPECT_THROW(MapReader::ReadSavedGame(""), std::runtime_error);
```

DanRStevens:
Nice job splitting the map reading methods. That worked out nice and clean.

test/Maps/MapReader.test.cpp
Brett208: oh, okay

Brett208:
Thanks. Everything looks ready to merge now.

Something strange happened at the end requiring a merge. I was trying to locally combine the commits `Update MapType enum option from Save to SavedGame` and `Update to SavedGame in Test`. However, apparently I had already pushed them to the public repository.

-Brett

DanRStevens:
Ahh, I was wondering what happened with that merge. Makes sense though, that's what happens when history diverges, and why people suggest not rewriting public history (branches pushed to GitHub are public, though certainly more limited than master).

If you wanted to back out of the re-write, you could potentially rebase additional changes on top of the already public history. If you really want the new history, you can force push to overwrite the old history. That's not so bad on a  branch, assuming you're the only one currently working on it. In this case I made a few contributions to the branch, so that might have seemed a bit risky, in case I had work in progress.

I hadn't thought of doing a merge before. The merge seems like an easy way to resolve divergent histories. It preserves both sets of history, both the actual of how it was originally developed, and the cleaned up version of how you think it should have been done. Well, I learned something today.

PR #169, opened 2018-09-02
Add basic map writer tests

Brett208:
This is my first time using the syntax:

```sizeof(decltype(buffer)::value_type));```

Which is a feature of C++11 as pointed out in the issue #158. I had to append decltype to make it work.

To test the memory writer overload, I created a buffer of 1000 char. This was a bit arbitrary. It just needs to big enough that it fails before running out of space in the buffer. It is failing after reading the header and checking the version tag. 

Give me a minute to refactor here.

### DIFF
--- a/makefile
+++ b/makefile
@@ -61,7 +61,7 @@ TESTSRCS := $(shell find $(TESTDIR) -name '*.cpp')
 TESTOBJS := $(patsubst $(TESTDIR)/%.cpp,$(TESTOBJDIR)/%.o,$(TESTSRCS))
 TESTFOLDERS := $(sort $(dir $(TESTSRCS)))
 TESTLDFLAGS := -L./
-TESTLIBS := -lgtest -lgtest_main -lpthread -lOP2Utility
+TESTLIBS := -lgtest -lgtest_main -lpthread -lOP2Utility -lstdc++fs
 TESTOUTPUT := $(BUILDDIR)/testBin/runTests
 
 TESTDEPFLAGS = -MT $@ -MMD -MP -MF $(TESTOBJDIR)/$*.Td
--- a/src/Maps/MapWriter.cpp
+++ b/src/Maps/MapWriter.cpp
@@ -99,7 +99,7 @@ namespace MapWriter {
 		void WriteContainerSize(Stream::Writer& streamWriter, std::size_t size)
 		{
 			if (size > UINT32_MAX) {
-				throw std::runtime_error("Container size is too large for writing into an Outpost 2 maps.");
+				throw std::runtime_error("Container size is too large for writing into an Outpost 2 map");
 			}
 
 			streamWriter.Write(static_cast<uint32_t>(size));
--- a/src/Maps/MapWriter.cpp
+++ b/src/Maps/MapWriter.cpp
@@ -9,7 +9,7 @@
 namespace MapWriter {
 	// Anonymous namespace to hold private methods
 	namespace {
-		void WriteHeader(Stream::Writer& streamWriter, const MapHeader& header);
+		void ValidateMap(const MapData& mapData);
 		void WriteTilesetSources(Stream::Writer& streamWriter, const std::vector<TilesetSource>& tilesetSources);
 		void WriteTileGroups(Stream::Writer& streamWriter, const std::vector<TileGroup>& tileGroups);
 		void WriteContainerSize(Stream::Writer& streamWriter, std::size_t size);
@@ -27,7 +27,9 @@ namespace MapWriter {
 
 	void Write(Stream::Writer& streamWriter, const MapData& mapData)
 	{
-		WriteHeader(streamWriter, mapData.header);
+		ValidateMap(mapData);
+
+		streamWriter.Write(mapData.header);
 		streamWriter.Write(mapData.tiles);
 		streamWriter.Write(mapData.clipRect);
 		WriteTilesetSources(streamWriter, mapData.tilesetSources);
@@ -46,13 +48,18 @@ namespace MapWriter {
 
 
 	namespace {
-		void WriteHeader(Stream::Writer& streamWriter, const MapHeader& header)
-		{
-			if (!header.VersionTagValid()) {
-				throw std::runtime_error("All instances of version tag in .map and .op2 files must be greater than 0x1010.");
+		void ValidateMap(const MapData& mapData) {
+			if (!mapData.header.VersionTagValid()) {
+				throw std::runtime_error("All instances of version tag in .map and .op2 files must be greater than 0x1010");
 			}
 
-			streamWriter.Write(header);
+			if (mapData.header.TileCount() != mapData.tiles.size()) {
+				throw std::runtime_error("Header reported tile width * tile height does not match actual tile count");
+			}
+
+			if (mapData.header.numTilesets != mapData.tilesetSources.size()) {
+				throw std::runtime_error("Header reported tileset count does not match actual tileset sources");
+			}
 		}
 
 		void WriteTilesetSources(Stream::Writer& streamWriter, const std::vector<TilesetSource>& tilesetSources)
--- a/src/Maps/MapWriter.cpp
+++ b/src/Maps/MapWriter.cpp
@@ -57,7 +57,7 @@ namespace MapWriter {
 				throw std::runtime_error("Header reported tile width * tile height does not match actual tile count");
 			}
 
-			if (mapData.header.numTilesets != mapData.tilesetSources.size()) {
+			if (mapData.header.tilesetCount != mapData.tilesetSources.size()) {
 				throw std::runtime_error("Header reported tileset count does not match actual tileset sources");
 			}
 		}
--- a/test/Maps/MapWriter.test.cpp
+++ b/test/Maps/MapWriter.test.cpp
@@ -8,7 +8,7 @@
 TEST(MapWriter, EmptyMapData)
 {
 	// Write to File
-	const std::string testFilename("maps/data/test.map");
+	const std::string testFilename("Maps/data/test.map");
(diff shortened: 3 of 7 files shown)

### CODE REVIEW

File: test/Maps/MapWriter.test.cpp [hunk 1]
```diff
@@ -0,0 +1,47 @@
...
+	
+	// Write to File
+	const std::string testFilename("maps/data/test.map");
+	EXPECT_THROW(MapWriter::Write(testFilename, mapData), std::runtime_error);
+	XFile::DeletePath(testFilename);
+
+	// Write to Memory
+	std::vector<char> buffer;
+	buffer.resize(1000);
+	Stream::MemoryWriter memoryWriter(buffer.data(), buffer.size() * sizeof(decltype(buffer)::value_type));
```
DanRStevens: This looks correct. We'll have to get this implemented into the library for you.

test/Maps/MapWriter.test.cpp
```diff
@@ -0,0 +1,47 @@
...
+{
+	MapData mapData;
+
+	EXPECT_THROW(MapWriter::Write("", mapData), std::runtime_error);
+}
+
+TEST(MapWriter, VersionTag) {
+	const std::string testFilename("maps/data/test.map");
+	MapData mapData;
+	const int minVersionTag = 0x1010;
```
DanRStevens: This should probably become a public member of the `MapData` class. We should also perhaps add a note that the game itself checks for this minimum version, and will fail to load a game otherwise.

test/Maps/MapWriter.test.cpp
```diff
@@ -0,0 +1,47 @@
...
+	
+	// The following 2 tests will require a legitimately constructed Map before running.
+	//mapData.header.versionTag = minVersionTag;
+	//EXPECT_NO_THROW(MapWriter::Write(testFilename, mapData));
```
DanRStevens: I think this test should be fine after some API adjustments. We shouldn't need to generate a bunch of data to save an empty map file.

test/Maps/MapWriter.test.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+	//mapData.header.versionTag = minVersionTag + 1;
+	//EXPECT_NO_THROW(MapWriter::Write(testFilename, mapData));
+
+	mapData.header.versionTag = minVersionTag - 1;
+	EXPECT_THROW(MapWriter::Write(testFilename, mapData), std::runtime_error);
```
DanRStevens: Hmm. Do we want this to fail? This certainly seems worth of a warning, though I'm not sure it constitutes failure.

Perhaps more to the point, should the user even be specifying the version number? The version number is baked into the file format, and that's controlled by the file writing code. Perhaps this value should be auto set.

src/Maps/MapWriter.cpp
```diff
@@ -27,7 +27,9 @@ namespace MapWriter {
 
 	void Write(Stream::Writer& streamWriter, const MapData& mapData)
 	{
-		WriteHeader(streamWriter, mapData.header);
+		ValidateMap(mapData);
+
+		streamWriter.Write(mapData.header);
```
DanRStevens: I like this change. It makes sense to separate the validation of `mapData` from writing of `mapData.header`.

src/Maps/MapWriter.cpp
```diff
@@ -46,13 +48,18 @@ namespace MapWriter {
...
 			}
 
-			streamWriter.Write(header);
+			if (mapData.header.TileCount() != mapData.tiles.size()) {
+				throw std::runtime_error("Header reported tile width * tile height does not match actual tile count");
+			}
+
+			if (mapData.header.numTilesets != mapData.tilesetSources.size()) {
+				throw std::runtime_error("Header reported tileset count does not match actual tileset sources");
+			}
```
DanRStevens: I'm not convinced we need the `ValidateMap` method. I think perhaps the `header` struct could be synthesized from the other fields, rather than validated against them. That could greatly simplify the need for error checking. It also means we're not pushing the problem onto the user of the library, where they have to manually fill in the `header` from all the component parts.

makefile
```diff
@@ -61,7 +61,7 @@ TESTSRCS := $(shell find $(TESTDIR) -name '*.cpp')
 TESTOBJS := $(patsubst $(TESTDIR)/%.cpp,$(TESTOBJDIR)/%.o,$(TESTSRCS))
 TESTFOLDERS := $(sort $(dir $(TESTSRCS)))
 TESTLDFLAGS := -L./
-TESTLIBS := -lgtest -lgtest_main -lpthread -lOP2Utility
+TESTLIBS := -lgtest -lgtest_main -lpthread -lOP2Utility -lstdc++fs
```
Brett208: I had not thought about this being a problem. Thank you for catching and fixing.

test/Maps/MapWriter.test.cpp
DanRStevens: I see some oddities with making this constant a static member of the `MapData`. The constant is used in `MapHeader`, which is in a separate header file included by `MapData`. There would be a circular dependency. Putting the constant in `MapHeader` would avoid the circular dependency, but doesn't seem to make as much sense.

Perhaps `MapHeader` isn't really an independent struct. Perhaps it should be declared within the `MapData` struct. The dependencies would make more sense that way. It would allow the constant to be declared in a place accessible to both the parent struct and the sub struct.

DanRStevens: Additionally, we would need to *define* the constant in a .cpp file. We can *declare* and set an initial value in a header file. That would be sufficient for some uses. But if we ever take the address of the constant, it would need to be defined in a .cpp file or we'd get errors.

Declaration (in .h):
```cpp
static const int32_t MinimumSupportedVersion = 0x1010;
```

Definition (in .cpp):
```cpp
const int32_t MapData::MinimumSupportedVersion;
```

When writing the map data, it may be useful to take the address of this constant to pass to a Stream::Write method. Hence I'm inclined to put the definition there.

----

The placement of the constant definition makes me think the reading and writing code is not very independent of the host struct. Perhaps it should not be separated into independent files. Indeed, the structure of these files is noticeably different from other classes and struct in the project.

test/Maps/MapWriter.test.cpp
Brett208: I was approaching MapWriter as having to produce a valid map for consumption with Outpost 2 to be successful. In this case, failure is okay. 

I think baking the version number or not depends on what we want the MapData class to represent. Currently, the goal of MapData is to closely mirror the structure of a Map File on disk with the addition of C++11 containers and small helper functions. It is more of a structure than a class. Eliminating the version number from MapData would diverge from this goal since it is represented on disk.

How do we want to define a valid MapData struct? If we want to say that a valid map file may not be consumable by Outpost 2, then we can eliminate the check.

We could say a valid map file does not contain any logical errors. In this case, I think we would check:
 - Does the number of tiles reported in the header match the recorded number of tiles
 - Does the number of tilesets reported in the header match the recorded number of tilsetSources
 - Can the width of the map be represented by a base 2 logarithm
 - Is the clip rectangle within the valid bounds for the map
 - Is the TerainType::tileRange within the range of valid tile definitions
 - Is the TerrainType::bulldozedTileIndex contained within the TerrainType::tileRange
 - Are the TerrainType::wall indices contained within the TerrainType::tileRange
 - Is the first index for lava tiles contained within the TerrainType::tileRange
 - Is the maximum number of tile definitions allowed Outpost 2? (Maybe this one is allowed in case of extension???)
 - etc

However, this will not necessarily allow the map to work with Outpost 2 because it isn't enforcing the hardcoded values required in TerrainType to work with things like established lava animations, bulldozing, etc. 

Probably more on this later.

-Brett

DanRStevens: I would argue `MapData` has already diverged from being a simple struct. In particular, it contains STL containers, which don't map directly to disk. You'll notice that while `MapHeader` is a packed struct, `MapData` is not. Further, `MapData` contains a private member function. I think we should make `MapData` a class. The file reading code can become a static factory method. The writing code can become regular member functions.

Of course that's too big of a refactor for this branch. We should deal with that at another time.

As for the MapHeader, perhaps that should become a contained struct, and used as a local variable of the read/write methods. The vector sizes imply some fields. Ones that aren't implied can be stored in a smaller struct, or as direct fields.

The version tag is used in about 3 different places, only one of which is in the header. Hence the version tag is best associated with MapData, rather than MapHeader. That makes using it from MapHeader awkward if MapHeader remains an independent struct in its own file.

----

As for the in-memory format of MapData, I don't think it needs to strictly conform to Outpost 2 limitations. Rather, that should be a separate check. I think some sort of verify method makes sense, but I don't think it should verify stored size fields against computed vector sizes. Rather I think it should initialize a local variable struct with the sizes of those vectors. The verify method is then left mostly with bounds checking, such as making sure the number of tiles and map dimensions fit within what can be represented by the Outpost 2 saved game file format.

I think you did a pretty good job listing potential error conditions. I would want to modify the ones that verify header fields against vector sizes though.

DanRStevens:
Ok, wow. I think we need to limit the scope of this branch. We've unearth some design issues that may require additional thought or discussion, and are best dealt with over perhaps two additional separate branches.

As it stands now, I think we could merge this branch with what we have now. There are meaningful valuable changes here. I don't see anything horribly broken that I wouldn't want on master. We have new tests that exercises code we weren't testing before. The tests are perhaps not as extensive as we'd like, but that's largely because attempting to write them has unearthed some design questions we may need to address first concerning what is proper behaviour.

I vote to merge to merge, and then tackle the design as additional work.

test/Maps/MapWriter.test.cpp
DanRStevens: I think we should keep the MapHeader struct. However, I think it should be used as a local variable during saving and loading, rather than being a member variable of MapData. Some fields are worth saving as members, some aren't.

Issue #152, opened 2018-08-23
Implement compression algorithms


DanRStevens:
The VOL file code is able to decompress the stock game data, however, it currently lacks corresponding compression code. We should implement this.

First point simply being the cool factor in finally implementing those neat little algorithms. However, there is actually a practical side to this proposal. Testing.

There is an obvious simple way to test the decompression code. We can store some compressed data within the testing code, run it through the decompressor, and check for correct output. However, currently all example compressed data comes from sheets.vol from the official Sierra releases. There are copyright issues with including such proprietary data in the repository. I would like to avoid this.

A way around this, is to write corresponding compression code. We could then do a round trip test of the decompression code, by first compressing a known string of our own choosing, and then testing that it properly decompresses to its original value. As an added bonus, we can construct strings to test specific features of the decompression algorithm, and so have small highly targeted test strings.

----

As a side task, perhaps we could comment the original code a little better. Attempting to write the corresponding compression code is likely to increase our understanding of the original decompression algorithm.

PR #174, opened 2018-09-04
Activate Huffman encoder code

DanRStevens:
This actives and fixes some previously commented out encoding code.

This allows testing part of the decompressor against the corresponding encoding code. It verifies the encoder can be properly paired with the decoder to round-trip the data.

----

This does not yet verify the particular Huffman encoding scheme necessarily matches what was used in Outpost 2, only that the encoder and decoder are properly paired with each other.

This does not verify the full compression/decompression scheme used in Outpost 2. It only deals with the adapative Huffman codes. It does not deal with the fixed Huffman codes, nor the repeated runs of data.

### DIFF
--- a/src/Archives/AdaptiveHuffmanTree.cpp
+++ b/src/Archives/AdaptiveHuffmanTree.cpp
@@ -38,6 +38,11 @@ namespace Archives
 
 
 
+	AdaptiveHuffmanTree::NodeType AdaptiveHuffmanTree::TerminalNodeCount()
+	{
+		return terminalNodeCount;
+	}
+
 	// Returns the index of the root node.
 	// All tree searches start at the root node.
 	AdaptiveHuffmanTree::NodeIndex AdaptiveHuffmanTree::GetRootNodeIndex()
--- a/src/Archives/AdaptiveHuffmanTree.cpp
+++ b/src/Archives/AdaptiveHuffmanTree.cpp
@@ -91,12 +91,7 @@ namespace Archives
 		int curNodeIndex;
 		int blockLeaderIndex;
 
-		// Make sure the code is in range
-		if (code >= terminalNodeCount)
-		{
-			throw std::runtime_error("AdaptiveHuffmanTree DataValue of " + std::to_string(code)
-				+ " is out of range " + std::to_string(terminalNodeCount));
-		}
+		VerifyValidDataValue(code);
 
 		// Get the index of the node containing this code
 		curNodeIndex = parentIndex[code + nodeCount];
@@ -140,6 +135,16 @@ namespace Archives
 		}
 	}
 
+	// Raise exception if code is out of range
+	void AdaptiveHuffmanTree::VerifyValidDataValue(DataValue code)
+	{
+		if (code >= terminalNodeCount)
+		{
+			throw std::runtime_error("AdaptiveHuffmanTree DataValue of " + std::to_string(code)
+				+ " is out of range " + std::to_string(terminalNodeCount));
+		}
+	}
+
 	// Private function to swap two nodes in the Huffman tree.
 	// This is used during tree restructing by UpdateCodeCount.
 	void AdaptiveHuffmanTree::SwapNodes(NodeIndex nodeIndex1, NodeIndex nodeIndex2)
@@ -179,10 +184,7 @@ namespace Archives
 	int AdaptiveHuffmanTree::GetEncodedBitString(int code, unsigned int& bitCount)
 	{
 		// Make sure the code is in range
-		if (code >= terminalNodeCount)
-		{
-			throw std::runtime_error("Code value is out of range");
-		}
+		VerifyValidDataValue(code);
 
 		// Record the path to the root
 		bitCount = 0;
--- a/src/Archives/AdaptiveHuffmanTree.cpp
+++ b/src/Archives/AdaptiveHuffmanTree.cpp
@@ -127,7 +127,6 @@ namespace Archives
 	// Raise exception if nodeIndex is out of range
 	void AdaptiveHuffmanTree::VerifyValidNodeIndex(NodeIndex nodeIndex)
 	{
-		// Check that the nodeIndex is in range
 		if (nodeIndex >= nodeCount)
 		{
 			throw std::runtime_error("AdaptiveHuffmanTree NodeIndex of " + std::to_string(nodeIndex)
@@ -183,7 +182,6 @@ namespace Archives
 	// Subsequent branches are stored in higher bits
 	int AdaptiveHuffmanTree::GetEncodedBitString(int code, unsigned int& bitCount)
 	{
-		// Make sure the code is in range
 		VerifyValidDataValue(code);
 
 		// Record the path to the root
--- a/src/Archives/AdaptiveHuffmanTree.cpp
+++ b/src/Archives/AdaptiveHuffmanTree.cpp
@@ -71,7 +71,7 @@ namespace Archives
 	}
 
 	// Returns the data stored in a terminal node
-	AdaptiveHuffmanTree::DataValue AdaptiveHuffmanTree::GetNodeData(NodeIndex nodeIndex)
+	AdaptiveHuffmanTree::NodeData AdaptiveHuffmanTree::GetNodeData(NodeIndex nodeIndex)
 	{
 		VerifyNodeIndexInBounds(nodeIndex);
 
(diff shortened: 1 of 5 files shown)


src/Archives/AdaptiveHuffmanTree.cpp
```diff
@@ -127,14 +124,23 @@ namespace Archives
...
-		// Check that the nodeIndex is in range
 		if (nodeIndex >= nodeCount)
 		{
 			throw std::runtime_error("AdaptiveHuffmanTree NodeIndex of " + std::to_string(nodeIndex)
 				+ " is out of range " + std::to_string(nodeCount));
 		}
 	}
 
+	// Raise exception if code is out of range
+	void AdaptiveHuffmanTree::VerifyValidDataValue(DataValue code)
```
Brett208: This function name could probably benefit from shortening. Maybe 'VerifyData'. If that seems too short, 'VerifyDataValue'.

src/Archives/AdaptiveHuffmanTree.cpp
```diff
@@ -165,36 +171,28 @@ namespace Archives
...
-
-		// Get the node containing the given code
-		NodeIndex curNodeIndex = parentIndex[code];
+		VerifyValidDataValue(code);
 
 		// Record the path to the root
-		bitString = 0;
-		int bitCount = 0;
+		bitCount = 0;
+		int bitString = 0;
```
Brett208: An int called bitString seems a bit strange to me. Not sure if there is a better name? It may just be because I'm not familiar with how Huffman trees work though.

DanRStevens: Hmm, do you feel there is confusion with the `string` data type? The `bitString` is supposed to represent a variable number of bits, which are generally processed one at a time. You might think of it as a vector of `bool`, however, you can't ask the compiler to produce an array or vector of `bool` and have it pack it for you, hence an integer base type is normally used.

In this context, the expected number of bits is about 9, though as the tree rebalances, the length can skew both shorter and longer. I expect the codes to be less than 32-bits, hence why I just used an `int`. Admittedly an `unsigned` integer type would be better. I am playing a bit fast and loose with the type here.

As for the maximum code length, it is actually as long as the number of codes. Such a length would correspond to a maximum unbalanced tree in a linear configuration. However, to unbalance the tree to a point where you have greater than 32-bit code lengths, I suspect you'd have to process about 4GB+ of specially crafted data.

I was considering building a `BitString` class, but then found myself wondering if it is sufficiently different from the `BitStream` class. Maybe I can just pass a `BitStream` class to be used by the tree code. In particular, I found myself replicating tree traversal algorithms in the test code, so perhaps that can be wrapped into tree methods.

At any rate, dealing with the type in a more precise way seems like its own branch. I can't think of a better name for what the value represents, but the type can certainly be improved.

src/Archives/AdaptiveHuffmanTree.cpp
DanRStevens: I chose the name to closely match the `VerifyValidNodeIndex` method that already exists.

I'm not entirely in love with those two names either. I thought about shortening them, but also considered if they need to actually be longer. I think it's ok to use long names here. They're private members so their scope is limited anyway.

In terms of writing "verify" methods that throw exceptions, I think it's good practice for the name to somehow indicate what they are verifying, or under what conditions they throw. The "verify" word alone doesn't describe this, and neither does a noun indicating the type it works on. Hence I don't think "Verify<NounPhrase>" is good pattern for naming. As such, I'd like to avoid using "VerifyData", or "VerifyDataValue".

In this case, we are verifying the range of values for that type falls within the expected range. Perhaps a better name is `VerifyDataValueInRange`, though that sounds almost like it expects a range argument.

DanRStevens:
I just had a flashback to junior high school, where I had an English teacher that would mercilessly mock people for saying "cannot", and insisting that it wasn't a word. ("What is a 'cannot'?") Meanwhile, I always used to grit my teeth and think "that's how everyone writes it!"

Anyway, just checked Grammarly, and you're right. The common form is "cannot", and they recommend that form unless the "not" is part of a set phrase that follows. I'll update.

----

Thanks for pointing out the Windows warning.

Perhaps we should turn on warnings as errors, that way a warning causes the CI builds to fail. In terms of Windows and the Linux Clang build, they are currently clean of warnings, so that would be a good way to catch new problems as they are introduced. It might not work so well for the g++ builds that currently have the spurious warning from the acknowledge compiler bug.

For MSVC there are two relevant settings we should enable to treat warnings as errors. There is the [`/WX` compiler flag](https://docs.microsoft.com/en-gb/cpp/build/reference/compiler-option-warning-level?view=vs-2017), and the [`/WX` linker flag](https://docs.microsoft.com/en-gb/cpp/build/reference/wx-treat-linker-warnings-as-errors?view=vs-2017).

For Clang and g++, the "-Werror" flag will promote warnings to errors.

src/Archives/AdaptiveHuffmanTree.cpp
Brett208: I think my issue is the word Valid by itself is too generic and should just be removed if left by itself. Maybe like you mentioned it would be better just to be more descriptive then. 

Maybe VerifyDataRange or VerifyDataValueRange? I think these would be more descriptive of what part of DataValue is being verified and the names are not much longer. 

After thinking about it some more, part of the problem might be because DataValue consists of 2 generic terms. It isn't clear that DataValue belongs attached to a node. Maybe NodeData would be a better name as it would indicate the data came from a node? 

```
void AdaptiveHuffmanTree::VerifyNodeDataRange(NodeData code);
void AdaptiveHuffmanTree::VerifyNodeIndexRange(NodeIndex nodeIndex);
```

I think these could be shortened to below, but maybe this becomes too terse?

```
void AdaptiveHuffmanTree::VerifyDataRange(NodeData code);
void AdaptiveHuffmanTree::VerifyIndexRange(NodeIndex nodeIndex)
```

Using the name NodeData would go well with the following function as well:
```
NodeData GetNodeData(NodeIndex nodeIndex);
```

Hope this helps?

-Brett

src/Archives/AdaptiveHuffmanTree.cpp
Brett208: Thanks for explaining. I think bitString sounds good in this context.

Brett208:
I just recompiled on Windows and the warning has disappeared. That sounds like a not fun English class to be in. :|

I added the AdaptiveHuffmanTree test file to the MSVC project. It created a merge conflict with the MapHeader test file that just got added to the master branch that I cleaned up as well.

I think that I prefer allowing the code to pass without throwing errors for warnings. I feel like the compiler marks them as warnings for a reason and the different compilers disagree on what parts of code should be a warning. Sometimes we allow a warning to enter the codebase for a while as we work on other issues before cleaning it up like the Google Test project addition. If we reject warnings altogether, we would of had to fix these warnings before merging. I'm definitely onboard with eliminating 

What would be really nice, is if the automated build tests could tell you all tests pass but with warnings. This way we would be aware a warning was introduced, but could still merge if it makes sense at the time. Guess that is beyond the ability of the automated test reporting on GitHub right now though?

If you disagree with any of the naming suggestions above, please just merge in so we can move on to other updates.

Thanks,
-Brett

src/Archives/AdaptiveHuffmanTree.cpp
DanRStevens: I think we're on the same page here. The problem isn't the length, but rather it was a bad name. Coincidentally, I was also thinking about renaming `DataValue` to `NodeData`.

Ok, I'll do some renaming then.

DanRStevens:
I'm tempted to pre-create the test project file hierarchy as a way to reduce merge conflicts in the future. I think there should be a one line shell command to generate empty files with the right names. Would still need to add them to the Visual Studio project though. I would also like to rename the `MapData` class first.

----

The CI builds only get pass/fail reporting. If they pass, we have no reason to check the logs, so we wouldn't see any warnings. By failing the build, it forces us to look at the logs. If we wanted to merge, we'd still have that ability, just the CI builds would complain loudly as we did it. That's probably a good thing, and would force us to fix warnings quickly, as we did in the case of Google Test.

Perhaps this is a topic to be decided up on in a separate issue.

DanRStevens:
Just adding a note that changes on this branch go part way towards fulfilling Issue #152.

PR #177, opened 2018-09-06
Allow reading sprites

Brett208:
This branch is to allow reading OP2_ART.prt files into memory. The read function works until reading first frameCount from from Animation. After reading the frameCount, values are incorrect. I'm unable to correct this on my own and need some help. 

The code is contained in a subdirectory call Sprites. Class is name ArtFile, ArtReader, and ArtWriter. I'm open to suggestions here on naming.

 - Write function is incomplete and untested, so I would ignore it for now until reading is good.
 - I cannot get the private functions prototypes of ArtReader and ArtWriter to wrap in an anonymous namespace. Also, why are we wrapping MapReader/MapWriter private functions in an anonymous namespace?
 - I'm trying to remove the noise from ArtFile. So things like file headers and container sizes are not recorded in it. Instead, these are created by the writer or discarded by the reader to create ArtFile. This should reduce the amount of unneeded variables that have to be sifted through in ArtFile.

Thanks,
-Brett


DanRStevens:
> Also, why are we wrapping MapReader/MapWriter private functions in an anonymous namespace?

From a technical perspective, the anonymous namespace keeps those methods private to that compilation unit (.cpp file). There is no way to reference or link to them from elsewhere, so it prevents global namespace pollution, which otherwise might affect linking if two methods have the same name. Naturally this only applies to the private methods, since you need access to the public methods from other modules.

Remember that global functions have no concept of public versus private. They are all public. A header file might not provide prototypes to help link to them, but there is nothing stopping another module from writing its own forward declaration for what should be a private method. The anonymous namespace is like adding a unique prefix to the name, which can never be specified. Each compilation unit would have its own unique anonymous namespace.

From a design perspective, simply put, because we came up with a bad design. The use of anonymous namespaces was a bit of a hack to improve the design and reduce potential problems. Really, we should have taken it as a sign we were going down a bad road and should have refactored the code.

There really shouldn't be any split between MapData/MapReader/MapWriter. These should all be part of one class. The MapWriter methods should be regular member functions (and private member functions, rather than being wrapped in an anonymous namesapce). The MapReader methods should be static members (and private static member functions).

A reason for making the read methods static, is we don't actually have a valid complete MapData object until reading is complete. That is, they run before the object is constructed. This is similar to how you don't have a full complete object until the constructor finishes. If you throw during a constructor, the destructor will not run because the object never was, and so there is nothing to destruct. (If sub objects were constructed, those will need to be destructed, and the standard ensures they are). You can think of the static methods as factory methods, which are effectively very much like constructor methods. If there is a problem reading map data, you were never able to produce a fully functioning MapData object. It never was. It should not exist in some quasi complete state, where you might call methods on incomplete data, which might do who knows what. By making the methods static, you can choose to either return a complete object, or throw, such that no object is constructed or returned.

At the very least though, the odd design for MapData/MapReader/MapWriter was a good learning opportunity, which I think was instructive both for language features, and for design aspects.

----

Might not have time for a full review right now, but I'll at least take a peak.

src/Rect.h
```diff
@@ -6,6 +6,8 @@
 
 struct Rect
 {
+	Rect() : x1(0), y1(0), x2(0), y2(0) {}
+
```
DanRStevens: FYI: I believe the explicit constructor can affect a number of object properties that may be a bit unexpected. Such as what contexts a struct can be used, how initialization syntax works, or what default members are automatically provided. I'm a little fuzzy on the details, hence the vague warning this may have greater consequences than expected.

With that said, it's probably fine. Makes sense to initialize those values to 0.

I suppose if you really wanted to get into it, initialization could waste cycles if you're just going to immediately read into the object/overwrite it.

src/Sprites/ArtFile.h
```diff
@@ -0,0 +1,80 @@
...
+#include <vector>
+#include <cstdint>
+
+struct Color {
+	Color();
+
+	uint8_t red;
+	uint8_t green;
+	uint8_t blue;
+	uint8_t alpha;
```
DanRStevens: I believe the newer C++ standards allow `= 0` initializers in struct definitions. That might be a good alternative to supplying an explicit constructor.

src/Sprites/ArtFile.h
```diff
@@ -0,0 +1,80 @@
...
+	uint32_t unknown2; //(0x3C CC/DIRT/Garage/Std. Lab construction/dock, 0x0D spider walking)
+
+	std::vector<Frame> frames;
+
+	uint32_t unknownArtContainerCount;
+
+	std::vector<UnknownArtContainer> unknownArtContainer;
+};
+
+struct ArtFile
```
DanRStevens: The above structs could probably be moved into the ArtFile's namespace.

src/Sprites/ArtFile.h
```diff
@@ -0,0 +1,80 @@
...
+{
+public:
+	std::vector<std::array<Color, 256>> palettes;
```
DanRStevens: Might be good to set a type alias for this.

src/Sprites/ArtReader.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+namespace ArtReader {
+	// Anonymous namespace to hold private methods
+	//namespace {
+	void ReadPalette(Stream::SeekableReader& seekableReader, ArtFile& artFile);
+	void ReadImageMetadata(Stream::SeekableReader& seekableReader, ArtFile& artFile);
+	void ReadAnimations(Stream::SeekableReader& seekableReader, ArtFile& artFile);
+	Animation ReadAnimation(Stream::SeekableReader& seekableReader);
+	Frame ReadFrame(Stream::SeekableReader& seekableReader);
+	//}
```
DanRStevens: I think these methods should become private static members of ArtFile.

src/Sprites/ArtWriter.cpp
```diff
@@ -0,0 +1,71 @@
...
+#include <cstdint>
+
+namespace ArtWriter
+{
+	// Anonymous namespace to hold private methods
+	//namespace {
+	void WritePalettes(Stream::SeekableWriter& seekableWriter, const ArtFile& artFile);
+	void WriteAnimations(Stream::SeekableWriter& seekableWriter, const ArtFile& artFile);
+	void WriteAnimation(Stream::SeekableWriter& seekableWriter, const Animation& animation);
+	//}
```
DanRStevens: These should probably become private member functions of ArtFile.

src/Sprites/PaletteHeader.cpp
```diff
@@ -0,0 +1,44 @@
+#include "PaletteHeader.h"
+
+namespace PaletteTag {
+	extern const std::array<char, 4> Palette{ 'C', 'P', 'A', 'L' };
+	const std::array<char, 4> Section{ 'P', 'P', 'A', 'L' };
+	const std::array<char, 4> Header{ 'h', 'e', 'a', 'd' };
+	const std::array<char, 4> Data{ 'd', 'a', 't', 'a' };
+}
```
DanRStevens: Interesting how exactly one of these is declared `extern`.

I suspect this has something to do with taking the address of a constant, and the difference between declarations and definitions.

A declaration goes in the header file. It can and should define the value for constants. A definition goes in the source (.cpp) file. It does not need to define the value, which should be imported from the header file. The definition in the source file reserves actual space for the constant. As in, there are bytes written to the executable's image, and the constant has a place in the virtual address space of the process. This is what allows the address of the constant to be taken. Obviously you can't do this in the header file, or every compilation unit (.cpp) file that imports the header will allocate space for the constant, and you'd have multiple copies of it floating around in memory. That's generally a bad thing.

If you don't take the address of a constant, the compiler may be able to feed the defined constant value forward, and eliminate any references to an actual constant. Hence you can use a constant with only a declaration and no definition, and not get errors. You'd only get errors about a missing definition once you try to take the address of the constant.

DanRStevens: Additionally, these constants should probably be defined in the namespace of the class that uses them.

src/Rect.h
Brett208: Someone random on the internet's opinion: http://www.cs.technion.ac.il/users/yechiel/c++-faq/init-lists.html

I found when testing the version number of MapData, I would get inconsistent test results before initializing the value in the constructor. Sometimes, the compiler would chose a random number that was correct and sometimes it would chose a number that was incorrect. By explicitly setting values to 0 (or something else), it eliminates the possibility of receiving intermittent errors caused by randomly initialized values. I'm not aware of any limits this puts on the use of the class/struct.

For a struct like Rect, someone might decide to create 10,000 of them for legitimate reasons, so there could be a noticeable performance hit maybe?

I'm not hard set on doing this if it makes sense not to or only in certain cases. It just seemed better than letting the compiler chose wild and random numbers that could cause intermittent issues.

Brett

src/Sprites/ArtFile.h
Brett208: I like this idea and will try it out. It is a cleaner approach.

src/Sprites/ArtFile.h
Brett208: okay. We can try it out.

src/Sprites/PaletteHeader.cpp
Brett208: The one defined as extern is used outside of this file. The others are only used locally. It could probably use some refactoring though.

Brett208:
Thanks for looking the code over quickly.

I am not a fan of having ArtFile be responsible for the contents of ArtFile, saving an ArtFile, loading an ArtFile, and manipulating an ArtFile. It is too much. There is a nice separation of concerns between the actual contents, loading, and saving. I feel like the source file gets too large and has too many members when it does all these things. Not that a namespace separation has to be the answer, but I thought it worked well for MapReader and MapWriter.

I used a mixture of the OllyDbg notes, Cynex's readme, and BlackBox's Python script. Unfortunately, Cynex's readme contained C code that I have trouble reading, and I don't know Python. I haven't tried your Visual Basic code yet, but will probably look at that next. I don't know Visual Basic either. The loading at the end gets really tricky, so it is hard to understand in languages that I'm trying to guess on their syntax.

-Brett

src/Sprites/ArtReader.cpp
Brett208: I'd like to figure out why I'm getting compilation errors when wrapping them in the anonymous namespace.

src/Sprites/ArtFile.h
Brett208: Yeah, this code is rough because I'm having trouble getting it to read properly. It will need cleaned up once I actually know what is going on.

Brett208:
Using the Visual Basic as another resource, I was able to finally get the .prt file reading properly with some minor head bashing on the way. Fortunately, Visual Basic syntax isn't too difficult to decipher. We should consider updating the OllyDbg code to be maybe a bit more specific on what is going on.

Some improvements were noted earlier in the branch which have not been implemented yet. Just wanted to get working code online.

Also:

 - There are several unknown fields in the ArtFile struct and sub-structures. I'd like to consider removing any of these unknowns that are either completely unused by the game or are always set to the same constant value. They can just be produced in the Write code for writing to disk and checked/ignored by the read code. I would need help identifying these unless @DanRStevens wants to see them remain in ArtFile.h.

 - There is a SectionHeader class here that I designed and a SectionHeader class in the Archives code. I wanted to look at unifying them, but the VolFile SectionHeader contains padding in its length variable, so maybe they have to stay separate.

 - There may be some opportunity to define a Point struct to go along with our Rect struct. Is there any reason not to do so?

Thanks,
Brett

src/Rect.h
DanRStevens: Interesting read, thank you. I would agree that initializer lists can improve performance in some cases.

----

The point I brought up was more about having an explicit constructor, rather than the constructor using an initializer list. I did a little reading. Some relevant links are:
[Should I define the default constructor?](https://stackoverflow.com/questions/22184060/should-i-define-the-default-constructor)
[What are aggregates and PODs and how/why are they special?](https://stackoverflow.com/questions/4178175/what-are-aggregates-and-pods-and-how-why-are-they-special)

If you define a constructor, an automatic default constructor (having no parameters) is no longer provided. In this case, you're defining an explicit default constructor.

If you define a constructor, the type is no longer an aggregate. This affects brace initialization. You can initialize a struct with braces, unless you explicitly define a constructor, then it's an error:
```cpp
Rect rect { 1, 2, 5, 10 };  // Error with explicit default constructor
```

The above also affects array initialization:
```cpp
Rect rectArr[2] = {{1, 2, 5, 10}, {3, 4, 5, 10}}; // Error with explicit default constructor
```

----

The "randomly initialized" values are actually uninitialized values. The struct was being used before the fields were assigned values. That generally falls into the realm of undefined behaviour. If the compiler is able to detect use of uninitialized values, it will likely flag it with a warning or error. Unfortunately, there are cases where the compiler is not always able to detect this. In particular, the compiler likely has trouble with pointer and reference parameters, which could be in or out parameters. It's perfectly normal to pass uninitialized memory as an out parameter, as you may be expecting the function to initialize it. You wouldn't want to pass uninitialized memory as an in parameter.

Although random values from uninitialized fields is never good, I'm a little concerned this may be a sign of some deeper issue with the code. In particular, it may be trying to use a value before it should, or it may be skipping necessary error checks. Using zeros might be safer in the sense of no undefined behaviour, and likely no crash, but it may be just as wrong. Worse, it's now harder to detect because it doesn't crash or behave oddly.

src/Sprites/PaletteHeader.cpp
DanRStevens: Used outside of this file, or defined outside of this file? I would assume that it's defined elsewhere if it's declared as `extern` here. If it's used elsewhere, that may be a problem. Either way, the asymmetry is concerning.

Using `extern` will turn a variable definition into a simple declaration. It can be used to create a forward declare to some global variable that is defined elsewhere. Generally `extern` is used in a header file. You could use `extern` from a .cpp file, as a way to import a symbol from another .cpp file without having to create a header file, and that may be suitable for a one-off import, but otherwise it would be more properly placed in a header file.

If you use `extern` in all .cpp files, you may find the variable lacks a definition, which can lead to link errors if you ever take the address of it.

----

I believe `extern` is being used incorrectly here. I've tried playing with the code a bit, but have not yet managed to demonstrate any problems.

src/Sprites/ArtReader.cpp
DanRStevens: Hmm, I see. It's because you need to wrap both the forward declarations and the definitions in an anonymous namespace. The commented out lines suggest only the forward declarations were wrapped. As such, the definitions will be in a different namespace, and so be defining entirely new functions, leaving the anonymous namespace functions undefined.

Still, I disagree that anonymous namespaces are the correct solution here.

If all you want is to split up the implementation, you can already do that with a class. Define all methods in the header. Implement reading methods in one file, writing methods in another file, and data manipulations in yet another file.

DanRStevens:
I'm not opposed to splitting the implementation files by concern to keep source code size manageable. Though I do feel the data manipulation, reading, and writing, should all be part of the same class. Particularly since this isn't exactly a dumb data structure.

I think a better way to break things down is by components, such as palettes, layers, images, and animations. Maybe. Palettes definitely should be a separate concern.

----

I don't mind removing unused fields. I think it would be nice to have structs containing all fields, including unknowns, just for serialization purposes, though they don't need to stick around in memory in that form.

I would need to review old notes and possibly some assembly code to remember what some of that stuff did. It's been a long time.

It is tempting to re-use the SectionHeader. I think you're right though, in that bit field complicates things. Might be easier to keep it separate. Perhaps we could have common code to handle 4 letter section tags though.

Point and Rect seem like obvious pairs to have both of. It may even make sense to allow defining Rect in terms of Point, optionally, as overloaded methods.

src/Rect.h
Brett208: Thanks for providing the reading. I think there are pros and cons both ways, but we can stick with not defining default constructors for the POD like structs. 

While I agree that you should not rely on initializing values to 0 to hide bugs, it makes it easier to troubleshoot. Intermittent bugs are harder to deal with then repeatable results. Also, when I see a zero I know it was either not initialized (except by the constructor) or actually set to 0. If it is a random number I just don't know.

src/Sprites/ArtReader.cpp
Brett208: Okay, thanks I'll try this out to get the anonymous namespaces working. I had thought about separating the write/read into separate files but belonging to the same class/header. I'm not thrilled about all the associated function prototypes filling up the class' header file. This might be a good compromise though.

src/Sprites/PaletteHeader.cpp
Brett208: That variable is also defined as extern in the header file. Adding the extern in the 
.cpp file as well was a logic error on my part and I'll correct. Sorry for the confusion. We can move the variables' location out of this file when we now where we want it as well.

Brett208:
I'll make another pass on the code soon and clean it some more. Thanks for the expanded comments.

Brett208:
I did some refactoring of the ArtFile code and filled in more of the ArtReader code.

Some of the current head hurters are:

When reading in animation data, 4 counts are read at once. I know what the animationCount, frameCount, and subframeCounts are. However, the final count, which I labeled unknown here I do not know what it is counting. So, I don't know what to add up to check it against or to apply in the write function.

``` c++
uint32_t animationCount;
seekableReader.Read(animationCount);
artFile.animations.resize(animationCount);

uint32_t frameCount;
seekableReader.Read(frameCount);

uint32_t subframeCount;
seekableReader.Read(subframeCount);

uint32_t unknownCount;
seekableReader.Read(unknownCount);
```

In the image metadata, I wanted to validate the variable scanlineByteWidth. I apparently am not sure what this property represents as my attempt didn't work.
```
for (const auto& imageMeta : imageMetas) {
	//if (imageMeta.scanlineByteWidth % 32 != 0) {
	//	throw std::runtime_error("Image scan line byte width is not divisible by 32.");
	//}

	if (imageMeta.paletteIndex > palettes.size()) {
		throw std::runtime_error("Image palette index is out of range of available palettes.");
	}
}
```

Reading in the frame data is a bit complex requiring some bit manipulation. I've check the code below, and it seems to work correctly, not that I understand why this is the route they went. Anyways, I have not idea how to translate this into the write algorithm.

```
uint8_t subframeCount;
seekableReader.Read(subframeCount);
seekableReader.Read(frame.unknown);

if (subframeCount & 0x80) {
	seekableReader.Read(frame.optional1);
	seekableReader.Read(frame.optional2);
}
if (frame.unknown & 0x80) {
	seekableReader.Read(frame.optional3);
	seekableReader.Read(frame.optional4);
}

frame.subframes.resize(subframeCount & 0x7F);
for (int i = 0; i < (subframeCount & 0x7F); ++i) {
	seekableReader.Read(frame.subframes[i]);
}
```

Thanks,
Brett

Brett208:
Okay, I just figured out how to calculate the scan line size properly, so that issue is now resolved.

src/Sprites/ArtFile.h
DanRStevens: FYI: On a hunch, I just did a bit of testing. It seems defaulting fields to some value in a struct definition is the same as providing an explicit constructor. That is, the type is no longer an aggregate, and you'll get errors from brace initialization of instance variables.

src/Sprites/ArtFile.h
```diff
@@ -1,23 +1,17 @@
...
-	uint8_t alpha;
-};
-
 struct ImageMeta {
 	uint32_t scanlineByteWidth; //number of bytes in each scanline of image (this should be the width of the image rounded up to a 32 bit boundary)
 	uint32_t pixelDataOffset; // Offset of the pixel data in the .bmp file
 	uint32_t height; // Height of image in pixels
 	uint32_t width; // Width of image in pixels
-	uint16_t type;
+	ImageType type;
```
DanRStevens: Hmm, should we be explicit about the size of `ImageType` here? All the other fields are quite explicit. As it stands now, the size of this struct is dependent on external information, namely the base type of `ImageType`. We could potentially write:
```cpp
ImageType type:16;
```

I think that would make the serialization code a little more robust in regards to any changes to the ImageType enum.

src/Sprites/Animation.h
```diff
@@ -0,0 +1,42 @@
+#pragma once
+
+#include "../Point.h"
+#include "../Rect.h"
+#include <vector>
+#include <cstdint>
+
+struct Animation {
+	struct Frame {
+		struct SubFrame {
```
DanRStevens: I suggest renaming `SubFrame` to `Layer`. That would more closely match terminology in popular graphics packages.

Back in the day, I was never very happy with the original naming of that field. At the time, I didn't feel I had proper terminology to really describe what it was about. Now I'm thinking `Layer` is a much more obvious name.

src/Sprites/Animation.h
```diff
@@ -0,0 +1,42 @@
...
+		};
+
+		uint8_t unknown;
+		uint8_t optional1;
+		uint8_t optional2;
+		uint8_t optional3;
+		uint8_t optional4;
+
+		// Limited to size of uint16_t MAX
+		std::vector<SubFrame> subFrames;
```
DanRStevens: I assume  you mean the vector index is limited to fit within a `uint16_t`. I had to think about the meaning here, and make sure it wasn't about the vector occupying 0, 1, or 2 bytes only. Seeing the optional fields above put me in that mindset.

src/Sprites/ArtReader.cpp
```diff
@@ -90,7 +90,7 @@ namespace ArtReader {
 		for (Animation animation : artFile.animations) {
 			actualFrameCount += animation.frames.size();
 
-			for (Frame frame : animation.frames) {
+			for (Animation::Frame frame : animation.frames) {
```
DanRStevens: This would be a good place to use `auto`.

src/Sprites/ArtReader.cpp
```diff
@@ -13,6 +15,7 @@ namespace ArtReader {
 	void ReadAnimations(Stream::SeekableReader& seekableReader, ArtFile& artFile);
 	Animation ReadAnimation(Stream::SeekableReader& seekableReader);
 	Animation::Frame ReadFrame(Stream::SeekableReader& seekableReader);
+	void VerifyFrameCount(const ArtFile& artFile, std::size_t frameCount, std::size_t subframeCount, std::size_t unknownCount);
```
DanRStevens: What is this doing? The name is not so obvious.

src/Sprites/ArtWriter.cpp
```diff
@@ -86,6 +87,37 @@ namespace ArtWriter
...
+		seekableWriter.Write(animation.unknown);
+		seekableWriter.Write(animation.selectionRect);
+		seekableWriter.Write(animation.pixelDisplacement);
+		seekableWriter.Write(animation.unknown2);
+		
+		std::size_t frameCount = animation.frames.size();
+		if (frameCount > UINT32_MAX) {
+			throw std::runtime_error("There are too many frames in animation to write");
+		}
+		seekableWriter.Write(frameCount);
```
DanRStevens: Here `frameCount` is a `std::size_t`. I think it needs a `static_cast<uint32_t>`.

src/Sprites/ArtWriter.cpp
```diff
@@ -86,6 +87,37 @@ namespace ArtWriter
...
+
+	void WriteFrame(Stream::SeekableWriter& seekableWriter, const Animation::Frame& frame) 
+	{
+		std::size_t subframeCount = frame.subframes.size();
+
+		if (subframeCount > UINT8_MAX) {
+			throw std::runtime_error("Too many subframes in frame.");
+		}
+
+		seekableWriter.Write(subframeCount);
```
DanRStevens: Here `subframeCount` is a `std::size_t`. I think it need a `static_cast<uint8_t>`.

src/Sprites/ImageType.h
```diff
@@ -0,0 +1,15 @@
...
+{
+	MenuGraphic = 0,
+	GameGraphic = 1,
+	UnitShadow1 = 4,
+	UnitShadow2 = 5,
+	Unknown1 = 33,
+	TruckBed = 65,
+	Unknown3 = 97,
+	Unknown4 = 129
+};
```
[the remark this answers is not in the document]
Brett208: I think you mean enum instead of struct in your comment? 

Nice deductive work. I pulled those values out of ArtFile by tabulating all recorded values for the uint16_t in the existing .prt file. Maybe now that we know it is a bitfield, we could figure out what some of the other values mean. 

I'll update the code base to be this bitfield shortly.

src/Sprites/ArtFile.h
Brett208: I think this is a mute point as ImageType should actually be a bitfield as mentioned earlier. I'll update shortly.

I made ImageType a uint16_t scoped enum, so it should have always been the same size as uint16_t.

src/Sprites/Animation.h
Brett208: Layer sounds good to me. I'll update shortly. The other variant I saw was FrameComponent.

src/Sprites/ArtReader.cpp
Brett208: This loop and it's parent loop should both actually be const auto&. Thanks for pointing it out.

src/Sprites/ArtReader.cpp
Brett208: It is verifying the number of read frames, subframes (layers), and whatever the third optional count is counting equals the counts recorded in the .prt file. None of these values are needed for parsing the data, but since they are reported in the prt file, I figured they should be checked and ensure they match what was read.

I can add a comment explaining what it does or perhaps there is a better name?

src/Sprites/ArtWriter.cpp
Brett208: Nice catch, I'll fix it.

src/Sprites/ArtWriter.cpp
Brett208: Good catch.

src/Sprites/ArtFile.h
DanRStevens: Hmm, very good point. I didn't realize I was commenting on the same field there.

As for the `uint16_t` base type on the enum, the part I didn't like so much about it is the base type is specified with the enum declaration, separate from the struct declaration. It's the struct declaration where the size really matters, and the other fields all had explicit sizes in the struct. I suppose if the enum is changed to a struct of bitfields, it's the same issue, where the size of the sub-component is controlled by the separate declaration of that subcomponent.

I suppose the other option is to embed the bitfield struct within the `ImageMeta` namespace, or embed the fields themselves, into the `ImageMeta` struct. Not sure what is better. I suppose it depends on if those bit fields have use outside the context of the `ImageMeta` struct.

src/Sprites/Animation.h
DanRStevens: Good catch on the bitfield controlled size. I forgot about the masking and the 7-bit bitfield part.

src/Sprites/ArtReader.cpp
DanRStevens: I think this is a case of using `Verify<NounPhrase>` rather than `Verify<PropertyOfNounPhrase>`. I think a better name would be better than a comment. A better name makes the code more self documenting, whereas a comment has to be kept in sync with the code.

Maybe `VerifyCountsMatchHeader`?

src/Sprites/Animation.h
```diff
@@ -20,8 +20,8 @@ struct Animation {
 		uint8_t optional3;
 		uint8_t optional4;
 
-		// Limited to size of uint16_t MAX
-		std::vector<Subframe> subframes;
+		// Maximum count of items in container is 128
```
DanRStevens: I suppose technically the range is 0..127 inclusive, so the "maximum count" is `127`.

src/Sprites/ArtFile.h
```diff
@@ -43,7 +46,17 @@ struct ArtFile
...
+
+	struct LayerMetadata {
+		uint8_t layerCount : 7;
+		uint8_t bReadOptionalData : 1;
+	};
+
+	struct UnknownBitfield {
+		uint8_t unknown : 7;
+		uint8_t bReadOptionalData : 1;
+	};
```
DanRStevens: `LayerMetadata` and `UnknownBitfield` structs can probably be collapsed into a single struct. They have the same layout. I think they might really be two separate instances of the same struct.

src/Sprites/ArtWriter.cpp
```diff
@@ -88,16 +88,23 @@ void ArtFile::WriteAnimation(Stream::SeekableWriter& seekableWriter, const Anima
 
 void ArtFile::WriteFrame(Stream::SeekableWriter& seekableWriter, const Animation::Frame& frame)
 {
-	std::size_t subframeCount = frame.subframes.size();
+	std::size_t layerCount = frame.layers.size();
+
+	//TODO: FIX FACT THAT BITFIELDS ARE NOT BEING WRITTEN CORRECTLY.
```
DanRStevens: Ahh, do you mean the bit is not set if optional fields are present? Is this information still available here? It might be more than just checking if the optional fields are zero or not (and they might not even be initialized if they were not loaded).

It looks like you're writing raw casted `uint8_t` count fields, rather than using a bitfield struct which would force the count into a 7-bit field.

src/Sprites/BitCount.h
```diff
@@ -0,0 +1,13 @@
...
+
+enum class BitCount : uint16_t
+{
+	One = 1,
+	Four = 4,
+	Eight = 8,
+	Sixteen = 16,
+	TwentyFour = 24,
+	ThirtyTwo = 32
+};
```
DanRStevens: This seems more like a `uint` field than an `enum`. Granted, the actual expected values tend to be somewhat restricted, so I can kind of see why an `enum` was used here.

In terms of bitmap header values, and other image format headers, I think they typically use a `uint`.

src/Sprites/ImageMeta.h
```diff
@@ -1,10 +1,20 @@
...
+	struct ImageType {
+		uint16_t bGameGraphic : 1;  // 0 = MenuGraphic, 1 = GameGraphic
+		uint16_t unknown1 : 1; // 2
+		uint16_t bShadow : 1; // 4
+		uint16_t unknown2 : 1; // 8
+		uint16_t unknown3 : 1; // 16
+		uint16_t unknown4 : 1; // 32
+		uint16_t bTruckBed : 1; // 64
+		uint16_t unknown5 : 1; // 128
+	};
```
DanRStevens: Ahh, excellent. Exactly what I was suggesting.

src/Sprites/Animation.h
Brett208: Wouldn't 0-127 be a count of 128 items? Maybe our terminology isn't agreeing here?

src/Sprites/ArtFile.h
Brett208: It is hard for me to say sense we don't know enough about the structs. I think I would prefer leaving them as separate structs until we know more. Otherwise we will have 2 layer counts if we just go with 2 LayerMetadatas.

src/Sprites/ArtWriter.cpp
Brett208: Yeah, anything below this TODO is pretty much wrong right now.

src/Sprites/BitCount.h
Brett208: I thought about both. I liked the enum because it provides some automatic error checking if someone tries to use a bit count of 3 or 99. It also makes it arguably cleaner if you want to use a switch statement.

Brett208:
Okay, I'll modify the ArtFile to include the unknown data fields so we can write an ArtFile. Later we can clean up these fields if we ever figure out what they mean.

Thanks,
Brett

src/Sprites/ArtReader.cpp
Brett208: Sounds good. I'll update.

Brett208:
Just finished updating the codebase. I added some of the suggestions and spent some time using HxD to troubleshoot the ArtFile Write process. I teased out various bugs. The output of ArtFile::Write is now identical to the original copy from 0 to 0x1CA56. This is somewhere in the middle of the first Animation where it goes off the tracks for the rest of the file. 

At this point, I'm probably unable to finish troubleshooting the writer without some help or a long time sink as I've exhausted my normal tricks. While the FileFormat from the SVN repo was very helpful, I think we have learned some new data that should be added and it is vague as to what happens next.

At this point, I would be happy merging in current state. My next project is probably working on the homerolled BMP writer unless we make headway with the ArtFile shortly. I have a BMP writing to disk that can be opened. The pixels are mostly correct, but they have artifacts that appear to be caused by incorrect width or something which I'm thinking may be related to ScanLine issues. Since we are reading all the data to animation correctly, we should be able to extract images with the code we have.

Thanks,

-Brett

src/Sprites/Animation.h
DanRStevens: Correct, there are 128 distinct bit combinations, but one of those counts is 0, hence the maximum is 127.

src/Sprites/BitCount.h
DanRStevens: Sort of yes. An enum field could hold arbitrary integer values outside of what are simply declared in the `enum` though. Granted, that probably falls into the realm of undefined behaviour if you store and use an undeclared value.

For `enum class`, you may get a compiler error on the assignment line for an undeclared name (such as an explicit value), unless it is silenced with a cast. But the same is true in the other direction too. You can't convert from a strongly typed enum back to an integer without a cast. Considering the bit depth is often used in calculations, that could prove to be an annoying hassle.

You can get around the need for a cast when converting back to an integer by using a regular enum, but then you also lose the assignment safety, where arbitrary integer values could be assigned to the enum without error.

I'd say enums make more sense when the distinctions are merely symbolic, with the only restriction on the underlying bit pattern being that they are mutually exclusive of each other.

DanRStevens: Additionally, I'd like to add these values are loaded from a file, and so only known at runtime. The enum assignment checking is a static check, done at compile time. That means defeating the check with an explicit cast, or maintaining a big switch statement to convert numbers to their identical symbolic name form. The explicit cast, or explicit switch statement must be done in both directions with an `enum class`. Additionally, any switch statements must be maintained if the enum is every updated.

It would be more appropriate to use a uint field with an explicit runtime check and throw an exception when setting this field. You'd only need to worry about assignment, not reading the field.

src/Sprites/ArtFile.h
DanRStevens: Ahh, but I'm suggesting dropping the "Layer" part of the name in the struct, and make that part of the variable instance name.

That leaves open the question of what do we name the struct. I don't have a suggestion there:
```cpp
struct ThisStructNeedsABetterName {
	uint8_t count : 7;
	uint8_t bHasOptionalData : 1;
};
```

test/Sprites/ArtWriter.test.cpp
```diff
@@ -21,3 +23,40 @@ TEST(ArtWriter, BlankFilename)
 {
 	EXPECT_THROW(ArtFile::Write("", ArtFile()), std::runtime_error);
 }
+
+TEST(ArtWriter, scanLineByteWidth)
```
DanRStevens: The name of this test could be a little more clear about what is being tested. It not immediately obvious from looking at the code. In particular, I don't know why writing should throw.

I have noticed the `scanlineByteWidth` is 2 byte aligned, but not 4 byte aligned. It also seems kind of arbitrary in relation to the `width`. Not sure if that is related to the throwing though.

Additionally, is it necessarily to cleanup the output file if writing throws, or should the writing routine cleanup any partially written files? I suppose for debugging purposes we want those partial files around. Maybe something to think about for later once everything is tested and working.

test/Sprites/ArtWriter.test.cpp
```diff
@@ -21,3 +23,40 @@ TEST(ArtWriter, BlankFilename)
...
+	artFile.imageMetas.push_back(imageMeta);
+
+	std::string filename = "Sprites/data/test.prt";
+
+	EXPECT_THROW(ArtFile::Write(filename, artFile), std::runtime_error);
+
+	XFile::DeletePath(filename);
+}
+
+TEST(ArtWriter, paletteIndex)
```
DanRStevens: I'm also uncertain what this is trying to test, and would appreciate a more detailed test name here. The only difference I spot is the `scanlineByteWidth`, which is now a multiple of 4.

src/Sprites/ArtWriter.cpp
```diff
@@ -28,7 +29,16 @@ void ArtFile::WritePalettes(Stream::SeekableWriter& seekableWriter, const ArtFil
 	}
 
 	seekableWriter.Write(SectionHeader(TagPalette, static_cast<uint32_t>(artFile.palettes.size())));
-	seekableWriter.Write(PaletteHeader(artFile));
+
+	// Intentially do not pass palette as reference to allow local modification
+	// Switch red and blue color to match Outpost 2 custom format.
+	for (auto pallete : artFile.palettes) {
```
DanRStevens: I like this comment. This is a very good point about modifying a copy versus modifying in place. Though I'm a little uncertain what the comment is referring to here. Is this the use of `auto` without `&`? I suppose the array is of actual palette objects, and not simple references, so yes, I think you're correct here. Hmm, that's actually a nice way of handling it.

I think this warrants a test, where a write is performed and then the palette is checked to make sure it has not been modified. Alternatively, two back to back writes can be performed and checked they produce the same results.

test/Sprites/ArtWriter.test.cpp
Brett208: Since I am using the same filename for each test file, it seemed to make sense to delete them as I went. If I want to inspect the contents of a file, I typically just set a breakpoint before the delete call. It sort of keeps the clutter down when committing by not having all the produced test files laying around. All minor preference things. If you want to change this behavior to leaving the artifacts around, I don't mind.

test/Sprites/ArtWriter.test.cpp
Brett208: This is testing to ensure an error is thrown when writing if an ImageMeta palette index is outside the range of palettes being saved. I'll update to a little more descriptive test name. Let me know if it is still too vague.

src/Sprites/ArtWriter.cpp
Brett208: Thanks for looking this over. I think it is a little unusual way to solve the problem, hence the comment. 

I will try to add a test for this shortly as it should be pretty easy.

-Brett

Brett208:
Thanks for pointing out the weakness in the test code. Besides the naming being vague, there was an error in the logic that also probably contributed to confusion. Check the new tests out and let me know if they are easier to follow. They are more robust as well.

I think the next step is moving a lot of these write test cases over to ArtFile itself. Many are testing ArtFile validation functions that do not require actually writing to disk to test. This should save some overhead in not constant opening file streams for the tests.

If you are up for it, message me times you are thinking of working with ollyDbg and we can pair program through it. Probably more like demo to me what you are doing as I haven't played with it. 

Once I'm happier with the tests cases, I'll plan to merge even though the write is a bit broke so we can move on without this branch getting too long. I had some thoughts about the Bit Count enum but no time right now to devote.

Thanks,
-Brett

test/Sprites/ArtWriter.test.cpp
DanRStevens: Err, my meaning was should the core library code (not the test code) delete the output file if writing fails part way through? I think that would make sense to do.

I suppose that would mean debugging requires setting a breakpoint in the error handler that cleans up the file, within the library, though that would likely trigger before the error was reported in any console messages. Though if the only console error reporting we do is the default stack trace when an exception bubbles up to the top and terminates the program, then nothing would really have changed debugging wise.

test/Sprites/ArtWriter.test.cpp
DanRStevens: Ahh, ok, I missed some of the detail there.

Hmm, if ArtFile is a class with encapsulation, you can error check when setting the data, rather than when writing the data. Checking a palette index can be done long before the data is written. That moves the error reporting closer to the point that caused the error. It also means simpler testing as you don't have to generate and delete a temporary file.

For now though, I think a better name would suffice.

I think test names should follow the advice for "Verify" methods. They should describe the property being tested, rather than a noun phrase that has something tested about it. For example: `ScanlineByteWidthNonAlignedThrows`, or `PaletteIndexOutOfBoundsThrows`, or `PaletteColorsNotSwappedAfterWrite`.

DanRStevens:
I was just thinking we should do a pair program session again sometime. Not tomorrow though, I'll be living in a pagoda with a bunch of monks. Tuesday might be better.

The test code looks like it's testing the right thing.

I'm curious what you mean by moving the tests over. Do you mean validation within the class itself, done while setting fields, rather than delayed until write time? That would make sense, and would be consistent with black box testing. It's a more obvious way to test a public API, rather than trying to trigger errors indirectly through writing files with bad data. Trying to trigger errors with bad data likely has the same difficulties as trying to test private methods.

test/Sprites/ArtWriter.test.cpp
Brett208: Oh okay. Yeah, I agree that if an exception/failure occurs during writing, the file should be deleted. I'll tease this out and see if it is happening or not and fix if needed. We can worry about the test peace later if we notice a problem with getting the output for test purposes. So far it hasn't been a problem.

Brett208:
I could be available
 * 18 Sep Tuesday 11PM - Wednesday 3AM UTC
 * 19 Sep Wednesday 11PM - Thursday 3AM UTC

I was planning on testing the verify functions after adding data to ArtFile. This would ensure the validation worked without dealing with actual writing files. However, I forgot these are private functions and cannot be accessed directly. So scrap that thought.

I'm not sure what form we want ArtFile to take. Currently it doesn't validate data unless reading or writing. This makes it more of a data structure than a class. I'm okay with this. So we could leave it alone for now. 

Would you prefer we mark the containers as private and start controlling adding/removing the data through public function calls that validate the data on each call. Like:
 * AddPalette(Palette palette)
    * Ensure max palette count is not exceeded
 * UpdatePalette(int index, Palette palette)
 * RemovePalette(int index, Palette palette)
    * Ensure palette index not contained in any Image metas before removal
  * AddImage()
    * Ensure palette index is within palettes range
    * Ensure ScanLineWidth is valid (or calculate automatically?)
    * Ensure image container count is not exceeded

etc

Thanks,
Brett

PR #184, opened 2018-09-18
Refactor InitializeWaveHeader

DanRStevens:


DanRStevens:
Sounds like a good plan. I'll try to roll that into a next set of updates.

PR #185, opened 2018-09-18
Replace hardcoded value with `sizeof`

DanRStevens:


DanRStevens:
Yes, that can happen. Even more annoying, MSVC and g++ have different ways of specifying alignment. With MSVC there is a pragma for the entire struct. For g++ it is an attribute on each field. You can take a look at NetFix for details. It looks kind of ugly to add all the annotations. Most of the time it doesn't matter so much though, since usually structs are designed to follow natural alignment of fields, which for most purposes here they are.

If we want to be paranoid, we can add static asserts about the struct sizes right after each struct definition.

ldicker83:
Seems like you two may have already addressed this but wanted to note that sizeof(), while generally recommended, can cause unexpected problems on different platform/compiler configurations. I'd be sure to thoroughly test this on any new configuration that might crop up (especially macOS as Apple likes to do weird things).

DanRStevens:
Yes, we should probably add `static_assert`s for structs used directly in serialization. I added Issue #191 to track this.

PR #186, opened 2018-09-18
Remove base type on `VolPadding` enum

DanRStevens:
This enum is a two valued bit flag. It probably doesn't make sense to explicitly declare the base type as a 32-bit quantity.


DanRStevens:
Indeed, I started that trend thinking it was a good idea, but also with no research behind it. I think it would make more sense for non-bitfield enums, where the defined size is actually correct. More so though, it seems most useful for opaque enums.

PR #188, opened 2018-09-18
Use `vector::at`

DanRStevens:
Side note: It seems a bit weird to require users of the Sprite API to manually bounds check things. This may be a sign we should make the API a little more class like, with better encapsulation.

For now, perhaps we should consider using [`std::vector::at(size_type index)`](https://en.cppreference.com/w/cpp/container/vector/at) which does implicit bounds checking, rather than [`std::vector::operator[](size_type index)`](https://en.cppreference.com/w/cpp/container/vector/operator_at), which does not. This is still a burden on the caller, though it means less explicit code to write and maintain on our part.

### DIFF
--- a/src/Sprites/ArtFile.cpp
+++ b/src/Sprites/ArtFile.cpp
@@ -12,13 +12,6 @@ BitCount ArtFile::GetBitCount(std::size_t imageIndex)
 	return imageMetas[imageIndex].type.bShadow ? BitCount::One : BitCount::Eight;
 }
 
-void ArtFile::CheckImageIndex(std::size_t index)
-{
-	if (index > imageMetas.size()) {
-		throw std::runtime_error("An index of " + std::to_string(index) + " exceeds range of images");
-	}
-}
-
 void ArtFile::ValidateImageMetadata() const
 {
 	for (const auto& imageMeta : imageMetas) {
--- a/src/Sprites/ArtFile.cpp
+++ b/src/Sprites/ArtFile.cpp
@@ -5,11 +5,7 @@ const std::array<char, 4> ArtFile::TagPalette{ 'C', 'P', 'A', 'L' };
 
 BitCount ArtFile::GetBitCount(std::size_t imageIndex) 
 {
-	if (imageIndex > imageMetas.size()) {
-		throw std::runtime_error("Index exceeds image count");
-	}
-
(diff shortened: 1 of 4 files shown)


Brett208:
After thinking about this some, I would recommend moving the function `GetBitCount` from `ArtFile` to `ImageMeta`. This way you can retrieve the bit count if you are passed an `ImageMeta` without the `ArtFile`.  

As a side effect, it punts the problem of what to do with the container range problem.

I tested this branch to see what error message MSVC would create. It writes "invalid vector<T> subscript". This doesn't mention anything about what vector the problem came from.

-Brett

ldicker83:
What about checking yourself and throwing a std::runtime_error exception with a custom message? Or is it really an error that needs an exception thrown? Could it default to a safe default and just set a warning flag?

(I know I know, already reviewed and checked but thought I'd put it out there).

DanRStevens:
Hmm, good point about moving that method, that makes much more sense. Bounds checking can be done when we get access to the ImageMeta.

I've noticed you're removing the BitCount enum in another branch, so there will be some merge conflicts around that part. To ease merging, I've also removed the use of BitCount in the moved code.

----

Sorry if there's a bit of noise from updates. I used `git rebase` to try and fix an error, and accidentally changed more history than intended. It should be changed back now.

----

For the exception message, does it report the current index that caused the error, or the size of the vector? Sounds like you're not satisfied with the error messages produced by MSVC for `vector::at`. Would you prefer we do manual bounds checking and throw with custom messages? It's more code, but the result might be better.

----

Hello and welcome Leeor! I almost didn't realize that comment was from you.

Brett208:
Leeor, I think moving the function out of ArtFile solved the issue here.

Dan, The default message MSVC produces when vector::at encounters an index greater than its current range is verbatim: `invalid vector<T> subscript`. No index is provided. The name of the vector is also not provided. Let's move the discussion on how to handle this into the other branch since it doesn't really pertain here anymore.

DanRStevens:
Alright, I resolved the merge conflict with master. This brings in a renamed bounds checking method, which will be used instead of `vector::at`. That means this branch simply amounts to moving the `BitCount` method from `ArtFile` to `ImageMetas`, which removes the need to pass an index to the method.

PR #189, opened 2018-09-18
Use `vector::at` in archive code

DanRStevens:
There are some slightly controversial changes in this branch. By using `vector::at` for bounds checking rather than a custom method, it simplifies the code, but we also lose explicit control over the exception messages. In particular, we lose explicit reporting of the archive filename in the event a bounds check throws.

This branch is based on changes from other recent branches, which it would otherwise have merge conflicts with. For this reason, I added explicit merge commits to bring in changes from the other branches, so you can verify the changes unique to this branch by viewing commits added after the two merge commits.

### DIFF
--- a/src/Archives/ClmFile.cpp
+++ b/src/Archives/ClmFile.cpp
@@ -60,8 +60,7 @@ namespace Archives
 	{
 		VerifyIndexInBounds(index);
 
-		WaveHeader header;
-		InitializeWaveHeader(header, index);
+		auto header = InitializeWaveHeader(clmHeader.waveFormat, indexEntries[index].dataLength);
 
 		try
 		{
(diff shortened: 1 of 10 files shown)


src/Archives/ClmFile.cpp
```diff
@@ -54,9 +54,9 @@ namespace Archives
 	// Extracts the internal file corresponding to index
 	void ClmFile::ExtractFile(std::size_t index, const std::string& pathOut)
 	{
-		VerifyIndexInBounds(index);
+		const auto& indexEntry = indexEntries.at(index);
```
Brett208: While I'm not a fan of removing the custom error messaging, we could still return a reference here to reduce index lookups.

Brett208:
I'd like to discuss how this branch will affect the use of the library. Currently, OP2Archive and OP2MapImager use the provided error messaging when reporting problems to the end user. If we eliminate the custom error messages, on any of these range errors, the user would just see `invalid vector<T> subscript`. I think the custom message will provide a much better experience. 

One option would be to manually catch the range error in OP2Archive/OP2MapImager and then display a custom error message. But, this will require duplicating effort for each application that consumes the library in developing custom messages. 

Also, it is more difficult to write about the details of the error after catching than it would be just to provide a nice message from the point of throw in the library. There could be several sources of the error from the library and you have to think about where it could throw from to provide amplifying information.

I think the library is more robust with the custom error messages and would like to see the extra lines of code to provide them remain. Granted, I'm probably missing some errors and some of the messaging could be improved on from where it is.

Happy to consider any counterpoints to this.

-Brett

DanRStevens:
Lets close this then. The changes to reduce index lookups is easier to apply on a fresh branch than to try to gut and fixup this branch.

PR #190, opened 2018-09-19
Refactor create wave header

DanRStevens:
Implementation of suggested changes.

Side note: This is likely to have a minor merge conflict with other open branches that update indexing where the wave header is created.


DanRStevens:
Ahh, thought I'd missed something. Thanks for fixing that.

PR #192, opened 2018-09-21
Reduce index lookups

DanRStevens:


DanRStevens:
Ahh yes, auto build hiccup. I managed to re-run the build, and it passed. I had to log in to see the re-run button.

PR #198, opened 2018-09-24
Add static assert for sizeof structs used in serialization

### LINKED ISSUE #191: Add `static_assert` for `sizeof` structs used directly in serialization
The size of a struct is not well specified by the C++ standard. It often depends on compiler and platform settings. In particular, alignment requirements of the platform may cause the compiler to insert padding bytes either before fields in a struct, or at the end up a struct, which may affect spacing in arrays of structs. This padding affects the output returned by the `sizeof` operator.

To guard against possible problems, we can add `static_assert` lines which verify an expected `sizeof` for the struct. Such checks can be placed directly after the struct definition so they can be kept in sync and verified easily.

The `static_assert`s should provide early warnings of problems on new platforms or compilers, and give people a chance to insert the necessary size or alignment directives (which are compiler specific).

----

Example:
```cpp
struct SomeStruct {
  uint32_t a;
  uint32_t b;
  uint32_t c;
};
static_assert(12 == sizeof(SomeStruct), "Unexpected size of SomeStruct");
```

We might also consider using a(n evil) macro with the string pasting operator to report the name of the struct without duplicating it. Or we could omit the name and rely on the compiler reporting the line number to make it obvious. An example macro use might be:
```cpp
VerifyStructSize(12, SomeStruct);
```

Brett208:
Closes #191 

Also contains some cleanup for non-fixed width integer types used in serialization, missing preprocessor directives to ensure 1 byte aligned structs, etc.

Low priority for merging, but @DanRStevens probably worth a quick look when you have a chance to make sure it all looks good as I've never done anything like this before.

Good news is, Visual Studio would immediately flag warnings for wrong sized structs using static_assert without even making me compile the code. So it definitely works properly on MSVC.

### DIFF
--- a/src/Archives/VolFile.h
+++ b/src/Archives/VolFile.h
@@ -53,6 +53,8 @@ namespace Archives
 			CompressionType compressionType;
 		};
 
+		static_assert(14 == sizeof(IndexEntry), "VolFile::IndexEntry is an unexpected size");
+
 		// Specify boundary padding for a volume file section
 		enum class VolPadding
 		{
@@ -64,13 +66,15 @@ namespace Archives
 		{
 			SectionHeader();
 			SectionHeader(std::array<char, 4> tag, uint32_t length, VolPadding padding = VolPadding::FourByte);
+
 			std::array<char, 4> tag;
 			uint32_t length : 31;
 			VolPadding padding : 1;
 		};
-#pragma pack(pop)
 
-		static_assert(sizeof(SectionHeader) == 8, "SectionHeader not of required size");
+		static_assert(8 == sizeof(SectionHeader), "VolFile::SectionHeader is an unexpected size");
+
+#pragma pack(pop)
 
 		struct CreateVolumeInfo
 		{
--- a/src/Archives/ClmFile.h
+++ b/src/Archives/ClmFile.h
@@ -49,6 +49,8 @@ namespace Archives
 			void VerifyUnknown() const; // Exception raised if invalid version
 		};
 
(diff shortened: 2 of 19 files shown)


src/Archives/VolFile.h
```diff
@@ -64,13 +66,15 @@ namespace Archives
...
 			SectionHeader(std::array<char, 4> tag, uint32_t length, VolPadding padding = VolPadding::FourByte);
+
 			std::array<char, 4> tag;
 			uint32_t length : 31;
 			VolPadding padding : 1;
 		};
-#pragma pack(pop)
 
-		static_assert(sizeof(SectionHeader) == 8, "SectionHeader not of required size");
+		static_assert(8 == sizeof(SectionHeader), "VolFile::SectionHeader is an unexpected size");
```
DanRStevens: I approve of the order swap here. That reads better with the hardcoded size on the left, and matches typical assert order.

DanRStevens:
All tests are passing. I'm going to go ahead and merge this in.

Brett208:
Thanks for fixing the signed variable.

I didn't mention earlier, but this did catch a mismatch on the size of ImageMeta::ImageType between MSVC and G++. I had to add another field to the struct to make it line up on both compilers. This may have been why you could not properly compile the code on Linux on our last pair programming session.

-Brett

DanRStevens:
Ahh yes, I saw the bug fix. Good that we caught that.

A bit fuzzy now, but what I remember of the pair programming session was I didn't have debugger installed. Apparently I love bugs :wink:

PR #199, opened 2018-09-24
Resolve Compiler Warnings within IndexedBmpWriter

### LINKED ISSUE #196: G++ signed/unsigned warnings in IndexedBmpWriter
The following two warnings are produced by g++ and should be reviewed at some point:

src/Sprites/IndexedBmpWriter.cpp:137:47: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
  if (pixelCountIncludingPitch * pixelsPerByte != CalcScanlinePitch(bitCount, width) * std::abs(height)) {
     
src/Sprites/IndexedBmpWriter.cpp: In static member function ‘static void IndexedBmpWriter::VerifyPixelsContainedInPalette(uint16_t, std::size_t, const std::vector<unsigned char>&)’:
src/Sprites/IndexedBmpWriter.cpp:145:24: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
  if (paletteEntryCount == 1 << bitCount) {

Brett208:
- MSVC: C4334 'operator' : result of 32-bit shift implicitly converted to 64 bits (was 64-bit shift intended?)
 - G++: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]

Closes #196 

@DanRStevens, Another low priority branch. Worth you looking over since I'm not great with bitwise operations.

### DIFF
--- a/src/Sprites/IndexedBmpWriter.cpp
+++ b/src/Sprites/IndexedBmpWriter.cpp
@@ -138,8 +138,6 @@ void IndexedBmpWriter::VerifyPixelBufferSizeMatchesImageDimensions(uint16_t bitC
 
 void IndexedBmpWriter::VerifyPixelBufferSizeMatchesImageDimensionsWithPitch(uint16_t bitCount, int32_t width, int32_t height, std::size_t pixelsWithPitchSize)
 {
-	const uint16_t pixelsPerByte = 8 / bitCount;
-
 	// G++ will flag warning -Wsign-compare if comparing a signed and unsigned value
 	if (pixelsWithPitchSize != static_cast<uint32_t>(CalculatePitchSize(bitCount, width) * std::abs(height))) {
 		throw std::runtime_error("An incorrect number of pixels were passed.");
--- a/src/Sprites/IndexedBmpWriter.cpp
+++ b/src/Sprites/IndexedBmpWriter.cpp
@@ -38,7 +38,7 @@ void IndexedBmpWriter::Write(std::string filename, uint16_t bitCount, int32_t wi
 void IndexedBmpWriter::WriteHeaders(Stream::SeekableWriter& seekableWriter, uint16_t bitCount, int width, int height, const std::vector<Color>& palette)
 {
 	std::size_t pixelOffset = sizeof(BmpHeader) + sizeof(ImageHeader) + palette.size() * sizeof(Color);
-	std::size_t fileSize = pixelOffset + CalcScanlinePitch(bitCount, width) * std::abs(height);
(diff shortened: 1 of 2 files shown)


src/Sprites/IndexedBmpWriter.cpp
```diff
@@ -130,19 +136,20 @@ void IndexedBmpWriter::VerifyPixelBufferSizeMatchesImageDimensions(uint16_t bitC
...
 	}
 }
 
 void IndexedBmpWriter::VerifyPixelsContainedInPalette(uint16_t bitCount, std::size_t paletteEntryCount, const std::vector<uint8_t>& pixels)
 {
 	// Check if palette is full
-	if (paletteEntryCount == 1 << bitCount) {
+	// G++ will flag warning -Wsign-compare if comparing a signed and unsigned value
+	// MSVC will flag warning C4334 on x64 compilation if bit shift not set to 64 bit integer
+	if (paletteEntryCount == static_cast<uint64_t>(1) << bitCount) {
```
DanRStevens: Using a suffix of `u` here would make `1u` an unsigned value. That eliminates the warning on g++ and clang. Alternatively a suffix of `ull` can also be used to be explicit about the size, which also eliminates warnings on both g++ and clang. I hadn't seen the `ui64` suffix before. I suspect it's less common, or part of a newer compiler standard.

My recommendation here is to use the `u` suffix. In particular, we are dealing with a `std::size_t` value on the left, so the size may vary with compiler settings.

We should prefer suffixes over casts. Using a cast defeats many compiler checks, and so is more prone to bugs.

DanRStevens:
Seems the `1u` solution was producing warnings on 64-bit Windows. I found a workaround on StackOverflow about [Initializing an integer literal to `std::size_t`](https://stackoverflow.com/questions/22346369/initialize-integer-literal-to-stdsize-t)

Interestingly, there is a proposed extension to include a suffix for `std::size_t` literals.

Worth pointing out, in the accepted answer it was mentioned there is a difference between parenthesis initialization `()`, and brace initialization `{}`. The parenthesis initialization supposedly silences errors like a cast, while the brace initialization doesn't.

Brett208:
I didn't prefer ui64 or a static_cast, but there were the only solutions I was aware of. I've never seen a type constructed using std::size_t { X }. The complexity of C++ continues to grow. I tried ui64 first since it was recommended in the warning's official documentation.

I had CalculatePitch return an int32_t since this is a similar value to width, which is an int32_t. I suppose it doesn't need to be the same type though.

Thanks for the thorough review. The code base is better for it.

-Brett

PR #203, opened 2018-09-30
Bmp header validation

Brett208:
Basic validation code for when we start reading Bitmaps from file.

### DIFF
--- a/src/Sprites/ImageHeader.cpp
+++ b/src/Sprites/ImageHeader.cpp
@@ -58,19 +58,14 @@ void ImageHeader::VerifyIndexedBitCount(uint16_t bitCount)
 void ImageHeader::Validate()
 {
 	if (headerSize != sizeof(ImageHeader)) {
-		throw std::runtime_error("Image Header must be equal to " + std::to_string(sizeof(ImageHeader)));
+		throw std::runtime_error("Image Header size must be equal to " + std::to_string(sizeof(ImageHeader)));
 	}
 
 	if (planes != DefaultPlanes) {
-		throw std::runtime_error("Planes must be equal to " + std::to_string(DefaultPlanes));
+		throw std::runtime_error("Image format not supported: only single plane images are supported, but this image has " + std::to_string(DefaultPlanes));
 	}
 
 	VerifyValidBitCount(bitCount);
-
-	auto compressionCast = static_cast<uint32_t>(compression);
-	if (compressionCast > 13) {
-		throw std::runtime_error("A compression value of " + std::to_string(compressionCast) + " is invalid");
-	}
 
 	if (usedColorMapEntries > std::size_t{ 1 } << bitCount) {
 		throw std::runtime_error("Used color map entries is greater than possible range of color map (palette)");
--- a/src/Sprites/ImageHeader.cpp
+++ b/src/Sprites/ImageHeader.cpp
@@ -62,7 +62,7 @@ void ImageHeader::Validate()
 	}
 
 	if (planes != DefaultPlanes) {
-		throw std::runtime_error("Image format not supported: only single plane images are supported, but this image has " + std::to_string(DefaultPlanes));
+		throw std::runtime_error("Image format not supported: only single plane images are supported, but this image has " + std::to_string(planes));
 	}
 
 	VerifyValidBitCount(bitCount);
--- a/src/Sprites/ImageHeader.cpp
+++ b/src/Sprites/ImageHeader.cpp
@@ -54,3 +54,29 @@ void ImageHeader::VerifyIndexedBitCount(uint16_t bitCount)
 		throw std::runtime_error("A bit count of " + std::to_string(bitCount) + " does not support an indexed palette");
 	}
 }
+
+void ImageHeader::Validate()
+{
+	if (headerSize != sizeof(ImageHeader)) {
+		throw std::runtime_error("Image Header must be equal to " + std::to_string(sizeof(ImageHeader)));
+	}
+
+	if (planes != defaultPlanes) {
+		throw std::runtime_error("Planes must be equal to " + std::to_string(defaultPlanes));
+	}
+
+	VerifyValidBitCount(bitCount);
+
+	auto compressionCast = static_cast<uint32_t>(compression);
+	if (compressionCast > 13) {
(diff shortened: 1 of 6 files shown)


src/Sprites/ImageHeader.cpp
```diff
@@ -54,3 +54,29 @@ void ImageHeader::VerifyIndexedBitCount(uint16_t bitCount)
...
 }
+
+void ImageHeader::Validate()
+{
+	if (headerSize != sizeof(ImageHeader)) {
+		throw std::runtime_error("Image Header must be equal to " + std::to_string(sizeof(ImageHeader)));
+	}
+
+	if (planes != defaultPlanes) {
+		throw std::runtime_error("Planes must be equal to " + std::to_string(defaultPlanes));
```
DanRStevens: If a user gets this error, they may question why. Is this a format restriction, or is this a support restriction? I think this is due to limited support.

src/Sprites/ImageHeader.cpp
```diff
@@ -54,3 +54,29 @@ void ImageHeader::VerifyIndexedBitCount(uint16_t bitCount)
...
+	}
+
+	VerifyValidBitCount(bitCount);
+
+	auto compressionCast = static_cast<uint32_t>(compression);
+	if (compressionCast > 13) {
+		throw std::runtime_error("A compression value of " + std::to_string(compressionCast) + " is invalid");
```
DanRStevens: As the format could be expanded, and new compression codes added, I think this should be about lack of support.

src/Sprites/ImageHeader.cpp
Brett208: Wikipedia says: The number of color planes, must be 1. This is consistent with other sources. When I change the color planes to 0 of a monochrome BMP, MS paint still seems to read it fine.

As far as I know, planes isn't used in any bitmap specification (which makes me wonder what it exists for). So, if we allow it to be a different value, it will not affect anything we are doing. 

It might be an indicator that someone is doing something weird with the file or that the file is damaged? 

I'm happy leaving this in or deleting this check. @DanRStevens, what do you prefer. 

-Brett

src/Sprites/ImageHeader.cpp
Brett208: I would like to leave information on forcing the compression to be 0 (uncompressed) out of ImageHeader as this is our implementation concern and not specific to an ImageHeader. Would you prefer this check just be removed? I'm okay either way. 

It was also sort of an experiment for me to see what would happen if I casted an integer outside of the scoped enum's range. C++ doesn't mind. I think in C# this would have caused an exception, but I would need to check for sure.

-Brett

Brett208:
I'll wait for feedback on the 2 responses and then merge.

Thanks for the quick review.

-Brett

src/Sprites/ImageHeader.cpp
DanRStevens: You mention the Wikipedia article states that value must be 1, but then you have to wonder, why even have that field defined if it can only ever take 1 value? My guess is, it does or did have a use that is now uncommon.

----

If you had a 3 plane image, the red, green, and blue components might be stored in separate arrays, rather than have all 3 color components merged within a single entry within 1 array. Beyond graphics, this has uses for scientific sensors that measure wavelengths outside the visible spectrum, and may have more than 3 bands of data. This would be common for weather satellites, which often capture all kinds of raster data outside the realm of visible light.

Of course if the data represents a raster image outside the visible spectrum, you're going to have to fudge it to display it. Most raster packages are only designed for working with visible image data, so they ignore such concerns.

Though scientific raster data is not typically stored in BMP files. They typically come in alternate compressed raster formats.

----

I think we should definitely have that check. It's pretty clear we don't support multi-plane images.

My concern is about the wording of the message, and how the end user could make sense of it. Right now it would probably make them feel dumb and frustrated. We've effectively passed of responsibility for understanding this problem to the user. That's fine during development, but we need to clean it up. The only time you'd want to leave a message like that in the code is in contexts where the user is more of an expert in their domain than the programmer writing code for them.

Perhaps:
"Image format not supported: Only single plane images are supported, but this image has X planes."

src/Sprites/ImageHeader.cpp
DanRStevens: Hmm, if we just copy the pixel data around, it's probably fine if we don't understand the compression code. If we access the pixel data in any way at all, then it should be a compression format that we support.

I think we should avoid making assumptions about what codes are valid as per the specification, as the specification can be updated and extended. Writing the message as "is invalid" seems to suggest it is outside the specification of the file format, but we don't know for certain the specification won't change.

----

I suppose the question here is what do the validity checks represent. Are they checks that must be true for all bitmap files to be valid and conform to the specification, or are they about what we support. Sounds like you want this to be about the specification itself, rather than what we support. In which case, we should be more careful about what specification we are validating against. That is, we should include some kind of version number for the spec, and name the validate method accordingly. If we want to get really picky, even the size of the header structure can change between versions. That is one of the ways it determines which version the file format is. If you want to make this solely about the specification, than we may need to adjust the header size check, the plane check, and the compression code check. 

I would caution against checking for validity against the spec, since that could easily turn into a big hairy beast. I'm not prepared to put that much effort into understanding the intricacies of the evolution of the bitmap file format as it's progressed over the years from OS/2 to modern versions of Windows. I'm not even sure it's fully documented anywhere.

----

We should have some checks somewhere about what we support. If we only support uncompressed, and throw an error if we encounter any other type, I'm fine with that. That would be easy enough to do.

If we want to work a little harder at it, the error might be delayed to accessing the pixel data in a manner other than copying it all around as one big chunk. I'm not sure we need or want to go there.

----

I think I'm fine with keeping this check somewhere, but we may be going down a path that is much more work than we want to do. Your call.

If you keep the check, I suggest rewording the "is invalid" to "is unsupported".

Brett208: I'll delete the check here and it can be handled int he Reader/Writer as it is encountered based on what is supported.

-Brett

src/Sprites/ImageHeader.cpp
Brett208: I'll update the exception method to what is suggested.

-Brett

DanRStevens:
I think this should probably be merged now, unless you're still working on something.

PR #204, opened 2018-09-30
Add color comparison

DanRStevens:
I felt like doing something easy, that would also get me thinking about a TDD workflow. This is the result.

### DIFF
--- a/test/Sprites/Color.test.cpp
+++ b/test/Sprites/Color.test.cpp
@@ -1,0 +1,18 @@
+#include "../src/Sprites/Color.h"
+#include <gtest/gtest.h>
+
+TEST(Color, CompareEquality) {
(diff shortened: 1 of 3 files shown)


test/Sprites/Color.test.cpp
```diff
@@ -0,0 +1,67 @@
+#include "../src/Sprites/Color.h"
+#include <gtest/gtest.h>
+
+TEST(Color, CompareEqual) {
+	// Note: Internally EXPECT_EQ uses `operator ==`
+	//       EXPECT_EQ has better error formatting than calling == explicitly
+
```
Brett208: Thank you for adding this comment. I was wondering why the code was structured how it was before seeing it.

PR #208, opened 2018-10-01
Add OpenMode parameter to FileWriter constructor

### LINKED ISSUE #22: Allow controls for how FileStreamWriter manipulates files

DanRStevens:
This implements open mode support for the constructor of the `FileWriter` stream class.

This is based partly on the old branch to add open modes. The old branch has grown stale, being half a year old, and files and classes have been moved and renamed since then. I figured a fresh branch would be easier to work with.

I think this matches the overall structure of what we need, though I would like a bit of feedback on some of the details.

I'm not entirely happy with the `OpenMode` names, and would appreciate some feedback there.

We should add documentation on race conditions with the file existence checks. Any thoughts on where or how to document this? Should this be a function level comment in the header file? For reference: [Time of Check to Time of Use (TOCTTOU)](https://en.wikipedia.org/wiki/Time_of_check_to_time_of_use)

Once we get this polished, this should close #22.

### DIFF
--- a/src/Streams/FileWriter.cpp
+++ b/src/Streams/FileWriter.cpp
@@ -21,7 +21,7 @@ namespace Stream
 
 		// Translate flags to ios_base flags
 		std::ios_base::openmode iosOpenMode = std::ios_base::out | std::ios_base::binary;
-		if ((openMode & OpenMode::PreserveContents) == 0) {
+		if ((openMode & OpenMode::Truncate) != 0) {
 			iosOpenMode |= std::ios_base::trunc;
 		}
 		if ((openMode & OpenMode::Append) != 0) {
--- a/src/Streams/FileWriter.cpp
+++ b/src/Streams/FileWriter.cpp
@@ -6,13 +6,20 @@ namespace fs = std::experimental::filesystem;
 namespace Stream
 {
 	std::ios_base::openmode FileWriter::TranslateFlags(const std::string& filename, FileWriter::OpenMode openMode) {
-		if ((openMode & FileWriter::OpenMode::FailIfExist) != 0 && fs::exists(filename)) {
-			throw std::runtime_error("File OpenMode::FailIfExist specified, but file exists: " + filename);
-		}
-		if ((openMode & FileWriter::OpenMode::FailIfNoExist) != 0 && !fs::exists(filename)) {
-			throw std::runtime_error("File OpenMode::FailIfNoExist specified, but file does not exist: " + filename);
+		// Check for bad flag combinations
+		if ((openMode & FileWriter::OpenMode::FailIfExist) != 0 && (openMode & FileWriter::OpenMode::FailIfNoExist) != 0) {
+			throw std::runtime_error("Bad open mode for file write. Can not specify file be both pre-existing and not pre-existing. Filename: " + filename);
 		}
 
+		// File existance checks
+		if ((openMode & FileWriter::OpenMode::FailIfExist) != 0 && fs::exists(filename)) {
+			throw std::runtime_error("Bad open mode for file write. File required to be not pre-existing, but already exists. Filename: " + filename);
+		}
+		if ((openMode & FileWriter::OpenMode::FailIfNoExist) != 0 && !fs::exists(filename)) {
+			throw std::runtime_error("Bad open mode for file write. File required to be pre-existing, but does not currently exist. Filename: " + filename);
+		}
+
+		// Translate flags to ios_base flags
 		std::ios_base::openmode iosOpenMode = std::ios_base::out | std::ios_base::binary;
 		if ((openMode & OpenMode::PreserveContents) == 0) {
 			iosOpenMode |= std::ios_base::trunc;
@@ -29,7 +36,7 @@ namespace Stream
 		file(filename, TranslateFlags(filename, openMode))
 	{
 		if (!file.is_open()) {
-			throw std::runtime_error("File could not be opened.");
+			throw std::runtime_error("File could not be opened. Filename: " + filename);
 		}
 	}
 
--- a/src/Streams/FileWriter.cpp
+++ b/src/Streams/FileWriter.cpp
@@ -1,11 +1,32 @@
 #include "FileWriter.h"
 #include <stdexcept>
+#include <experimental/filesystem>
+namespace fs = std::experimental::filesystem;
 
 namespace Stream
 {
-	FileWriter::FileWriter(const std::string& filename) :
+	std::ios_base::openmode FileWriter::TranslateFlags(const std::string& filename, FileWriter::OpenMode openMode) {
+		if ((openMode & FileWriter::OpenMode::FailIfExist) != 0 && fs::exists(filename)) {
+			throw std::runtime_error("File OpenMode::FailIfExist specified, but file exists: " + filename);
+		}
+		if ((openMode & FileWriter::OpenMode::FailIfNoExist) != 0 && !fs::exists(filename)) {
+			throw std::runtime_error("File OpenMode::FailIfNoExist specified, but file does not exist: " + filename);
+		}
+
+		std::ios_base::openmode iosOpenMode = std::ios_base::out | std::ios_base::binary;
+		if ((openMode & OpenMode::PreserveContents) == 0) {
+			iosOpenMode |= std::ios_base::trunc;
+		}
+		if ((openMode & OpenMode::Append) != 0) {
+			iosOpenMode |= std::ios_base::ate;
+		}
+		return iosOpenMode;
+	}
+
+
+	FileWriter::FileWriter(const std::string& filename, OpenMode openMode) :
 		filename(filename),
-		file(filename, std::ios::trunc | std::ios::out | std::ios::binary)
+		file(filename, TranslateFlags(filename, openMode))
 	{
 		if (!file.is_open()) {
 			throw std::runtime_error("File could not be opened.");
--- a/src/Streams/FileWriter.cpp
+++ b/src/Streams/FileWriter.cpp
@@ -5,21 +5,21 @@ namespace fs = std::experimental::filesystem;
 
 namespace Stream
 {
-	std::ios_base::openmode FileWriter::TranslateFlags(const std::string& filename, FileWriter::OpenMode openMode) {
+	std::ios_base::openmode FileWriter::TranslateFlags(const std::string& filename, OpenMode openMode) {
 		// Check for bad flag combinations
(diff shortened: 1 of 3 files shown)


src/Streams/FileWriter.h
```diff
@@ -11,7 +11,16 @@ namespace Stream
...
 	public:
-		FileWriter(const std::string& filename);
+		// Open mode bit flags
+		enum OpenMode {
+			Default = 0,
+			FailIfExist = 0b0000'0001,
+			FailIfNoExist = 0b0000'0010,
+			PreserveContents = 0b0000'0100,
+			Append = 0b0000'1000,
+		};
```
Brett208: I'm a little confused by append and preserve contents. I expected both of these words to sort of mean the same thing in that the file would be opened and then seek to the end of the file.

-Brett

src/Streams/FileWriter.cpp
```diff
@@ -1,11 +1,32 @@
...
 #include <stdexcept>
+#include <experimental/filesystem>
+namespace fs = std::experimental::filesystem;
 
 namespace Stream
 {
-	FileWriter::FileWriter(const std::string& filename) :
+	std::ios_base::openmode FileWriter::TranslateFlags(const std::string& filename, FileWriter::OpenMode openMode) {
+		if ((openMode & FileWriter::OpenMode::FailIfExist) != 0 && fs::exists(filename)) {
+			throw std::runtime_error("File OpenMode::FailIfExist specified, but file exists: " + filename);
```
Brett208: For a console application, I would probably just want to display the error message to the end user. Using `OpenMode::FailIfExist` in the name would likely not be intuitive to a non C++ programmer. 

Maybe: Writing to non-existing filename " + filename + " failed due to a file with this name already existing"

This comment would also apply to `OpenMode::FailIfNoExist` error message as well.

-Brett

Brett208:
What do you think of putting the race condition information within the enum:

```
enum OpenMode {
	Default = 0,
        // FailIfExist creates a race condition.
	FailIfExist = 0b0000'0001,
	FailIfNoExist = 0b0000'0010,
	PreserveContents = 0b0000'0100,
	Append = 0b0000'1000,
};
```

This will allow intellisense to pick up the comment when typing the option using Visual Studio. It should also draw the user to the condition when they are reviewing the enum for the open mode they want to use.

src/Streams/FileWriter.cpp
```diff
@@ -1,11 +1,32 @@
 #include "FileWriter.h"
 #include <stdexcept>
+#include <experimental/filesystem>
+namespace fs = std::experimental::filesystem;
 
 namespace Stream
 {
-	FileWriter::FileWriter(const std::string& filename) :
+	std::ios_base::openmode FileWriter::TranslateFlags(const std::string& filename, FileWriter::OpenMode openMode) {
+		if ((openMode & FileWriter::OpenMode::FailIfExist) != 0 && fs::exists(filename)) {
```
Brett208: Should there be a check here that raises and exception if both FailIfNoExist and FailIfExist are both specified together?

src/Streams/FileWriter.h
DanRStevens: The names need work. As for the meaning, basically don't truncate. You might want to update a file, without appending new data to it.

I suppose I could name it `Truncate`, and make it part of the default flag set (which is currently all bits clear).

src/Streams/FileWriter.cpp
DanRStevens: You're right, the language should be updated. Mention of the flag should be replaced with a description.

src/Streams/FileWriter.cpp
DanRStevens: I had considered that. An exception will be raised in that case, since the file must either exist or not exist, and so one of those two exception cases will be hit.

Unless of course a race condition allows the check to sneak past. I suppose I could do a single read of the file status, and use it with both flag checks. Or yes, I could test explicitly for both flags being set simultaneously and call the programmer an idiot. :wink: 

I'll look into cleaning it up a bit.

DanRStevens:
Putting the race condition comment by the enum may be convenient. Though the problem is not with the enum itself, but rather how it is used with the filesystem API. The `fstream` library doesn't allow checking file existence as a condition of `open`, and so there must be a separate time of check versus time of use. The Windows API does allow checking this with `dwCreationDisposition`, and so code written directly to the Windows API would not suffer from a race condition. Similarly, Linux has the `O_EXCL` flag when opening files.

The problem is more that there is no single API we can code to that is platform independent and allows checking this. The existing C++ standard library is too limited in this regards. We could write platform specific code into the `FileWriter` class, but that's more work than I want to put into it.

Effectively, this is a weakness of the `FileWriter` implementation.

Though, now that I think about it, the flag enum is contained in the `FileWriter` class, so perhaps that would be an ok place to document this. I guess I'm used to having a generic flag set used by both readers and writers, which we are under no obligation to do.

What about the constructor for `FileWriter`. I assume a function level comment would be visible intellisense there. File opening is only ever done in the constructor.

src/Streams/FileWriter.h
Brett208: I think that truncate works well as the default behavior. It is typically what I want when saving to disk. Occasionally I might want to append, but that seems to be the not normal or default in work I've done.

src/Streams/FileWriter.cpp
Brett208: Something like that???

For a console application consuming this library, I might add options to the user that line up with these flags when saving data. In that case, a nice error message would come in handy telling the end user they need to only use one of the flags. Also not too big a deal since they will probably figure it out on their own as you mentioned.

src/Streams/FileWriter.h
DanRStevens: Yes, it was setup so the default would truncate.

It would be unusual to specify both truncation and appending, though I don't see that as necessarily wrong, just kind of odd. Sort of like specifying both needing to open an existing file, and needing to open a new file that doesn't previously exist. The result is quite predictable, maybe even useful in some rare cases (where the flags are variable parameters rather than hardcoded constants), but probably a bit silly most of the time.

Brett208:
Sorry for such a late reply on the new commit.

Reading through the exception messages and they are great now. They should be usable as is by a console application.

I am mulling over the lines below. I think we want to rename `OpenMode::PreserveContents` to `OpenMode::Truncate`. Or maybe something similar to truncate if we want to use a different word. `PreserveContents` and `Truncate` seem mutually exclusive to me. Perhaps I'm missing something though? 

``` C++
if ((openMode & OpenMode::PreserveContents) == 0) {
	iosOpenMode |= std::ios_base::trunc;
}
if ((openMode & OpenMode::Append) != 0) {
	iosOpenMode |= std::ios_base::ate;
}
```

Thanks,
Brett

DanRStevens:
Sorry I kind of left this open and hanging for some time.

I think you're right about using `Truncate` rather than `PreserveContents`. I've made the change.

I'd like someone to suggest flag names for `FileIfExists` and `FailIfNotExists`. Those names don't sound entirely right. The grammar seems off. I'm not certain the pluralization is obvious. I'm not certain "Not" is the appropriate negation. I'm wondering if they should also be complemented, to something like `OpenExisting` and `CreateNew`, where at least one of them needs to be set, rather then not being allowed to have both set. I'm open to other suggestions though.

After coming back to this, I think you're also right about not allowing both `Truncate` and `Append`. That is weird, and I decided to add a check to disallow it.

I still haven't made the documentation update.

Brett208:
Thanks for continuing to work on this.

Some other suggestions, although none seem too great. The middle seems my favorite.

* MustExist
* MustNotExist

--

* Preexisting
* NonExisting

--

* MustOverwrite
* NonOverwrite

-Brett

Brett208:
If I saw OpenMode::Preexisting then I would assume the file needed to be preexisting for overwrite/opening.

If I saw OpenMode::NonExisting then I would assume a new file needed to be made without overwriting an existing file.

So I guess these would be allowed conditions.

After thinking about it, I think it would be better to specify all the flags as allowed conditions in general. When I read OpenMode, I would assume on context that we are talking about what the specified open condition is.

I think FailIfExist is okay. It isn't tough to read or reason what is meant. Although it does specify failure condition where truncate and append specify what will happen. So maybe a little disjointed in that both allowed condition and failure conditions are specified in the same flag set?

FailIfNoExist gets a bit tough to reason about including both fail and no as qualifiers. I have to think about it for a second to know what is going on. It is clear what it does though after you think about it for a minute.

Hope the random musings are helping?

-Brett

DanRStevens:
The first part of your reply sounds like required conditions (MustBe), while the later part sounds like allowed conditions (CanBe).

How would you read: `OpenMode::Preexisting | OpenMode::NonExisting`
Does that make perfect sense, or is that a silly error?

Hmm, we should probably have consistent capitalization between the two.

You bring up a good point about being consistent with usage concerning Truncate and Append. I suppose there is a certain sense of wanting both to be positive, as well as a double negative being hard to read. That didn't really stand out to me as Truncate and Append are operations to perform upon opening, whereas Preexisting and NonExisting are conditions to check before allowing open to succeed. You're right though, in that the happy case is when a file is opened.

For reference, Windows uses allowed conditions. There it would be something like: `OpenExisting | CreateNew`

DanRStevens:
How about `CanOpenExisting` and `CanOpenNew`. They are both positive sounding. There should be less ambiguity over use as they both imply these are allowed conditions, rather than required conditions.

At least one of the two flags must be specified. Default would be to have both specified:
```cpp
OpenMode::CanOpenExisting | OpenMode::CanOpenNew
```

Brett208:
Sounds good. These names are easy to understand. Would like to get this merged.

DanRStevens:
Ok, I think that about wraps this up.

Tests show it behaves correctly. I'll maybe give you a chance to review the test code, but otherwise I plan to merge this.

I noticed I needed some casting to combine bit flags though. I thought that would only be needed for `enum class`, but it seems to require it for a regular `enum` too. Not too sure why, but I don't think I'll fight with it.

Please review test case names. I grouped them all under the name `FileWriterOpenMode`. I'm a little uncertain if the "OpenMode" part should be part of the first or second parameter.

PR #212, opened 2018-10-08
Add bitmap file and indexed bitmap reader

Brett208:
I apologize for the significant code change in this branch. This branch does the following:

 - Improve our use of Bit Count checks and Indexed Bit Count Checks
 - Move bitmap functions out of Write that should be shared with Reading or other tasks
 - Create non-static/static pair versions of many bitmap functions
 - Add a BitmapFile struct
 - Add ability to read a standard indexed bitmap file

Testing needs to be improved in the Bitmap Reader. I held off on writing more tests for now because I'd like to create a way to check two in memory BitmapFiles for equaility. Then we can do round-robin checks, where we create an in memory bitmap file, save it to disk, load it, and check to make sure it is equal to the initially created struct.

I'm excited about these changes although I'm sure there is room for improvement.

-Brett


Brett208:
I believe since we are forcing the padding to be byte aligned on BmpHeader and ImageHeader, we can use `std::memcmp` on them when comparing equality.

Next step is to actually read/write a file and compare the results in memory.

src/Bitmaps/BitmapFile.cpp
```diff
@@ -32,3 +32,14 @@ void BitmapFile::VerifyIndexedImageForSerialization(uint16_t bitCount)
 		throw std::runtime_error("Unable to read/write a non-indexed bitmap file. Bit count is " + std::to_string(bitCount) + " but must be 8 or less");
 	}
 }
+
+bool operator==(const BitmapFile& lhs, const BitmapFile& rhs) {
+	return lhs.bmpHeader == rhs.bmpHeader && 
+		lhs.imageHeader == rhs.imageHeader && 
+		lhs.palette == rhs.palette &&
+		lhs.pixels == rhs.pixels;
+}
```
DanRStevens: After the stuff I watched a little while ago on code formatting, I think I might recommend:
```cpp
return
    lhs.bmpHeader == rhs.bmpHeader && 
    lhs.imageHeader == rhs.imageHeader && 
    lhs.palette == rhs.palette &&
    lhs.pixels == rhs.pixels;
```

Related, should the `&&` go at the end of the line above, or the beginning of the line below? I was tempted to move them to the beginnings of the next lines, though I'm not sure that's a clear advantage.

DanRStevens:
I believe you are correct, in that `memcpy` should work given the assumptions we are already making here. There are no hidden fields or padding space that needs to be excluded from the comparison.

test/Bitmaps/BitmapFile.test.cpp
```diff
@@ -1,17 +1,60 @@
...
+void RoundTripSub(uint16_t bitCount, int32_t width, int32_t height)
+{
+	const std::string filename("Sprites/data/BitmapTest.bmp");
+
+	BitmapFile bitmapFile = BitmapFile::CreateDefaultIndexed(bitCount, width, height);
+	BitmapFile bitmapFile2;
+
+	EXPECT_NO_THROW(BitmapFile::WriteIndexed(filename, bitmapFile));
+	EXPECT_NO_THROW(bitmapFile2 = BitmapFile::ReadIndexed(filename));
+	EXPECT_EQ(bitmapFile, bitmapFile2);
```
DanRStevens: I would probably have used a memory stream here, rather than an actual disk file. Less cleanup, faster tests, and more robust in case files liter the filesystem or there are permission issues.

One downside is the filename variant of the Read/Write methods tends to wrap the raw stream methods, so technically you'd be testing a bit less. That means more tests to exercise the wrappers. At least, assuming you care about 100% code coverage. Might be sensible to aim for more of an 80% solution.

Anyway, curious what you think on the matter. I could see arguments for both ways.

test/Bitmaps/BitmapFile.test.cpp
```diff
@@ -1,17 +1,60 @@
...
+		RoundTripSub(4, 1, 2);
+		RoundTripSub(4, 8, 1);
+		RoundTripSub(4, 9, 1);
+	}
+	{
+		SCOPED_TRACE("8 Bit");
+		RoundTripSub(8, 1, 1);
+		RoundTripSub(8, 1, 2);
+		RoundTripSub(8, 2, 1);
+	}
```
DanRStevens: It's actually pretty neat how you did this. Interesting use of Google Test features.

PR #219, opened 2018-11-19
Disable minimum rebuild (/Gm) property on OP2UtilityTest project due to deprecation

Brett208:
This project setting is now deprecated and isn't essential to our project.

https://docs.microsoft.com/en-us/cpp/build/reference/gm-enable-minimal-rebuild?view=vs-2017


Brett208:
Good point about just deleting the property if we don't need it and it is deprecated. Just implemented in another commit.

-Brett

PR #220, opened 2018-12-11
Make Rect methods const

DanRStevens:


Brett208:
Good catch on the rect const being in the wrong place!

PR #222, opened 2018-12-13
Add TrimTilesetSources

DanRStevens:
Add `MapData::TrimTilesetSources()`.

### DIFF
--- a/src/Maps/MapData.cpp
+++ b/src/Maps/MapData.cpp
@@ -38,6 +38,8 @@ void MapData::TrimTilesetSources()
 		),
 		tilesetSources.end()
 	);
+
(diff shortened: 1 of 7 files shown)


DanRStevens:
Ahh, good catch on the bug.

You're probably right about removing the `tilesetCount` field from the map struct, as it has become a derived attribute by virtue of using vectors. Though in the interests of serialization, we probably want to keep the field in the `MapHeader` struct. The change might be to not use the `MapHeader` struct inside of the `MapData` struct. We could also normalize the map `height` field in the process, so access is similar to `width`.

This sounds like a separate larger issue. I'll open a new issue to record it.

As always, thank you for the project file updates.

PR #225, opened 2018-12-20
Update CircleCI Dockerfile

DanRStevens:
Add Clang to the Docker image, and switch to Clang as the default compiler in the CircleCI config.yml.

Image supports building with either Clang or GCC. Image includes Google Test with Google Mock.

This image is a bit larger than before, so that may slow builds slightly, though more can be done with this single image.


DanRStevens:
Heh. I was debating if I should add more changes to this branch, but I think I'll delay to a future branch. In particular, doing builds for both GCC and Clang. I believe that requires Workflows, which I'm still reading the CircleCI documentation on.

PR #229, opened 2019-01-06
Add Google Mock on Travis

DanRStevens:
This adds support for using the older `google-mock` Apt package on Ubuntu 14.04. This package contains both Google Mock and Google Test.

The newer `libgtest-dev` package can continue to be used on newer distributions (Ubuntu 16.04, 18.04). This package also contains both Google Test and Google Mock, and in a more standard layout. Unfortunately this newer package does not exist on Ubuntu 14.04, which TravisCI still uses.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,7 @@ script:
   # Use -k to continue after errors, ensuring full build log with all errors
   - make -k
   - make gtest
+  - make gmock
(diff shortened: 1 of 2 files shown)


DanRStevens:
Tests are passing, and this is a Linux only change. I'm going to go ahead and merge.

PR #231, opened 2019-01-08
Fix XFile::RemoveFilename so it does not return the filename when no …

Brett208:
…directory is passed

Also add unit tests for RemoveFilename


DanRStevens:
This looks like the same issue I ran into in #216. It was solved in https://github.com/OutpostUniverse/OP2Utility/pull/216/commits/214a657acd174e8edfa6d7a5803ef5fb102099ff by using `.generic_string()` instead of `.string()`.

DanRStevens:
Interesting. It seems the platforms have an inconsistency regarding the trailing slash.

DanRStevens:
I'm just realizing we have some duplication here. What's the difference between `GetDirectory` and `RemoveFilename`? They appear to both do the same thing.

Brett208:
Now that you mention it, both functions accomplish the same thing. I'm okay deleting one. Do you prefer one name over the other? I would probably prefer just keeping RemoveFilename, but not a big deal either way.

We may want to consider forcing a trailing slash in both builds so we have consistent results.

Thanks,
Brett

DanRStevens:
Agreed about forcing a trailing slash.

I think the current implementation of `GetDirectory` already forces a trailing slash, based on the test code already added and passing in the other pull request.

For the name, I would prefer going with `GetDirectory`, or maybe rename to `GetDirName`. The name `RemoveFilename` sounds very misleading to me.

PR #234, opened 2019-01-12
Improve replace filename

Brett208:


src/XFile.cpp
```diff
@@ -122,9 +122,12 @@ bool XFile::IsRootPath(const std::string& pathStr)
 std::string XFile::ReplaceFilename(const std::string& pathStr, const std::string& filenameStr)
 {
 	fs::path p(pathStr);
+
+#if _WIN32 || _WIN64
```
DanRStevens: According to Microsoft's list of [Predefined Macros](https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=vs-2017), the `_WIN32` macro is defined for both 32-bit and 64-bit builds. Typically I see only the `_WIN32` macro used when doing a Windows platform detect.

Brett208:
Added a short comment as well. Thanks for the suggestions.

-Brett

PR #236, opened 2019-01-12
Rename MapData.h/.cpp/.test.cpp to Map & rename class MapData to Map

### LINKED ISSUE #139: Better names for Tiles
The name "tile" is used in several different contexts, which could be confusing. Currently "Tile" is used to describe:
- An (x, y) location on a map
- A data structure describing the graphic and animation of the above, from the set of all tilesets
- A graphical representation within a single tile set.

Similarly, "tileIndex" is used in all 3 contexts, making it harder to tell what the value actually represents.

We should rename some variables to make a better distinction between these cases. Currently using "TileData" versus "TileInfo" just doesn't explain the difference.

Example code:
https://github.com/OutpostUniverse/OP2Utility/blob/0a8389e532cbe21193c1b654b6036c2695e0affa/src/Maps/MapData.h#L28-L52

Brett208:
We can open a new issue to finalize a global OP2Utility namespace as it is getting fairly far off topic from the title "Better names for Tiles".

Thanks,
Brett


Brett208:
Okay, I won't close the issue then. Just trying to clear out some of the longstanding issues if I could. 

Thanks,
Brett

PR #238, opened 2019-01-14
Make all directories singular

Brett208:
Next up was supposed to be adding OP2Utility namespace. Unfortunately, I cannot use the namespace1::namespace2 nested convention out of the box. Apparently that was a feature added in C++17.

I have to target C++17 on our VS projects to use. (Apparently we have not been targeting C++17 even though we are using the filesystem module?)

When I enabled C++17, it broke the Microsoft provided Google Test nuget package. I think maybe we want to target the nuget package mentioned here: https://github.com/google/googletest/issues/1111. It lists a macro that must be set to use C++17 on MSVC and I think maybe it is the official googletest nuget package??? (https://www.nuget.org/packages/googletest)

If these changes look good, we can merge now or hammer through the namespace changes if they fall into place easily.

-Brett


Brett208:
@DanRStevens since you are good with these changes, I'll merge and work the C++17/googletest issue in a different branch. 

Thanks,
Brett

Issue #239, opened 2019-01-17
Broken Saved Game Parsing


DanRStevens:
We may have broken saved game parsing. I received the following errors when trying to run the OP2MapImager on my Outpost 2 folder:
```
Render initialized (May take up to 45 seconds): SGAME6.OP2
All instances of version tag in .map and .op2 files should be greater than 4112Error reading from file

Render initialized (May take up to 45 seconds): SGAME1.OP2
All instances of version tag in .map and .op2 files should be greater than 4112std::bad_alloc

Render initialized (May take up to 45 seconds): SGAME0.OP2
All instances of version tag in .map and .op2 files should be greater than 4112Error reading from file
```

This might happen if reading has been offset by an incorrectly sized field.

Brett208:
I looked at this today. I believe it is caused by:
```
---------------------------
	4	tag - must match tag at first DWORD of .map file else error (see Header Section)
---------------------------
		Note: extra data can be read here, depending on the parameters to the load routine.
		.map files shouldn't load any data here, probably only saved games (which do seem to load extra data here)
```

For a saved game, we are not accounting for the extra data between the 2 version tag reads.

```C++
streamReader.Read<uint32_t>(map.terrainTypes);
ReadVersionTag(streamReader);
// TODO: Account for saved game data
ReadVersionTag(streamReader);
// TODO: Check if tile group data exists on a saved game.
ReadTileGroups(streamReader, map);
```

-Brett

DanRStevens:
I think you're right. Thanks for figuring that out.

Perhaps we should write a method to read the unknown counts, and advance the stream by the calculated size of that data. We probably don't need to worry about storing the data at this point.

```
---------------------------
	4	tag - must match tag at first DWORD of .map file else error (see Header Section)
---------------------------
		Note: extra data can be read here, depending on the parameters to the load routine.
		.map files shouldn't load any data here, probably only saved games (which do seem to load extra data here)
	4	numUnits
	4	**TODO** figure out what this is
	4	**TODO** figure out what this is
	4	**TODO** figure out what this is
	4	sizeofUnit - 0x78 (120 bytes) if sizeofUnit != 0x78 && numUnits != 0 then Error
- - - - - - - - - - - - - -
		Note: Reading continues here in a subroutine at 0x00446CF0
	4	numObjects1 (number of 512 byte objects)
	4	numObjects2 (number of 4 byte objects)
		Note: if (numObjects1 == 0) return (successfully)
[...]

PR #240, opened 2019-01-17
Make map version check throw instead of log to cerr

DanRStevens:
Parsing should stop when a bad version number is found, so throwing is more appropriate. It also allows the caller to control how and where output is displayed, which logging to `cerr` doesn't.

Added additional diagnostic output to error message.

This update relates to issue #239, but doesn't solve it. This change should be considered separate.

### DIFF
--- a/src/Map/MapReader.cpp
+++ b/src/Map/MapReader.cpp
@@ -95,7 +95,11 @@ void Map::ReadVersionTag(Stream::Reader& streamReader)
 
 	if (versionTag < MapHeader::MinMapVersion)
 	{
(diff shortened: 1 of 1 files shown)


DanRStevens:
You're probably right. I would prefer being able to load a map with an incorrect version and display a warning, though we have no proper mechanism to communicate warnings that allow the caller to choose how the warning is displayed.

I suppose one method is to not check during load, but rather to validate afterwards. That's a topic for another day though.

And I guess we don't really need to support maps with invalid version numbers at this point. We shouldn't encounter any in the wild.

Ehh, I guess I've changed my mind.
¯\_(ツ)_/¯

PR #241, opened 2019-01-18
Allow Map::Read to read saved game portions without crashing

Brett208:
I'm not sure our documentation is correct for a saved game after unitID2 is read. However, this will parse without throwing an error and properly reads in map data from a saved game.

-Brett

### DIFF
--- a/src/Map/Map.h
+++ b/src/Map/Map.h
@@ -97,7 +97,7 @@ private:
 	static void WriteContainerSize(Stream::Writer& streamWriter, std::size_t size);
 
 	// Read
-	static Map Map::ReadMapBeginning(Stream::Reader& streamReader);
+	static Map ReadMapBeginning(Stream::Reader& streamReader);
 	static void SkipSaveGameHeader(Stream::SeekableReader& streamReader);
 	static void ReadTilesetSources(Stream::Reader& streamReader, Map& map, std::size_t tilesetCount);
 	static void ReadTilesetHeader(Stream::Reader& streamReader);
--- a/src/Map/Map.h
+++ b/src/Map/Map.h
@@ -34,7 +34,7 @@ public:
 	Map();
 
 	static Map ReadMap(std::string filename);
-	static Map ReadMap(Stream::SeekableReader& seekableReader);
+	static Map ReadMap(Stream::Reader& seekableReader);
 	static Map ReadSavedGame(std::string filename);
 	static Map ReadSavedGame(Stream::SeekableReader& seekableReader);
 
@@ -95,11 +95,12 @@ private:
 	static void WriteContainerSize(Stream::Writer& streamWriter, std::size_t size);
 
 	// Read
+	static Map Map::ReadMapBeginning(Stream::Reader& streamReader);
 	static void SkipSaveGameHeader(Stream::SeekableReader& streamReader);
 	static void ReadTilesetSources(Stream::Reader& streamReader, Map& map, std::size_t tilesetCount);
 	static void ReadTilesetHeader(Stream::Reader& streamReader);
 	static void ReadVersionTag(Stream::Reader& streamReader);
-	static void ReadSavedGameSecondChunk(Stream::SeekableReader& streamReader);
+	static void ReadSavedGameSection2(Stream::SeekableReader& streamReader);
 	static void ReadTileGroups(Stream::Reader& streamReader, Map& map);
 	static TileGroup ReadTileGroup(Stream::Reader& streamReader);
 };
--- a/src/Map/Map.h
+++ b/src/Map/Map.h
@@ -101,7 +101,7 @@ private:
 	static void SkipSaveGameHeader(Stream::SeekableReader& streamReader);
 	static void ReadTilesetSources(Stream::Reader& streamReader, Map& map, std::size_t tilesetCount);
 	static void ReadTilesetHeader(Stream::Reader& streamReader);
-	static void ReadVersionTag(Stream::Reader& streamReader);
+	static void ReadVersionTag(Stream::Reader& streamReader, uint32_t lastVersionTag);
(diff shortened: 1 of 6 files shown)


src/Map/Map.h
```diff
@@ -34,7 +34,7 @@ class Map
 	Map();
 
 	static Map ReadMap(std::string filename);
-	static Map ReadMap(Stream::Reader& seekableReader);
+	static Map ReadMap(Stream::SeekableReader& seekableReader);
```
DanRStevens: Would be nice to not need this change, though I can see why we might want to do this for now. Perhaps at some point we could factor out the common part into a method. The `if` controlled part could then be unconditionally called from the `ReadSavedGame` method.

src/Map/MapReader.cpp
```diff
@@ -34,8 +35,17 @@ Map Map::ReadMap(Stream::Reader& streamReader)
...
+	if (map.IsSavedGame()) {
+		ReadSavedGameSecondChunk(streamReader);
+	}
+
 	ReadVersionTag(streamReader);
-	ReadTileGroups(streamReader, map);
+	
+	if (!map.IsSavedGame()) {
+		ReadTileGroups(streamReader, map);
+	}
```
DanRStevens: I had never really thought about this before, but yeah, I guess the tile groups wouldn't be stored in saved game files.

src/Map/MapReader.cpp
```diff
@@ -99,6 +109,31 @@ void Map::ReadVersionTag(Stream::Reader& streamReader)
...
+	savedGameData.objects1.resize(512 * savedGameData.objectCount1);
+	streamReader.Read(savedGameData.objects1);
+	savedGameData.objects2.resize(savedGameData.objectCount2);
+	streamReader.Read(savedGameData.objects2);
+
+	streamReader.Read(savedGameData.unitID1);
+	streamReader.Read(savedGameData.unitID2);
+
+	//streamReader.Read(savedGameData.unitRecord);
+	streamReader.SeekRelative(0x3BF88);
```
DanRStevens: This is likely affected by the unit limit patch, which I believe invalidated old saved game files, due to the expansion in size of the unit array.

It would be nice if we could distinguish files for this. Perhaps I should have bumped the file format version number.

src/Map/SavedGameDataSection2.h
```diff
@@ -0,0 +1,21 @@
...
+	uint32_t unknown3;
+	uint32_t sizeOfUnit;
+	uint32_t objectCount1;
+	uint32_t objectCount2;
+	std::vector<uint8_t> objects1;
+	std::vector<uint32_t> objects2;
+	uint32_t unitID1;
+	uint32_t unitID2;
+	std::vector<uint8_t> unitRecord;
+};
```
DanRStevens: I'm wondering if we should avoid storing `vector` fields in the same struct as fixed length fields. The main use of fixed length fields is serialization, which we can't do if vectors are embedded in the same struct.

----

We could create dummy structs for object1 and object2, which contain just an array of `uint8_t`. That would give a better type and automatic sizing for `objects1` and `objects2`. Same for the `unitRecord` field.

src/Map/Map.h
Brett208: Good idea. Since we are peppering the code with if statements anyways, might as well separate the code.

src/Map/MapReader.cpp
Brett208: I would say better late than never on bumping the version, but we might have a lot of angry fans if we bump it for no gain. If you have time, it might be nice to update the documentation file to represent the new expected value.

src/Map/SavedGameDataSection2.h
Brett208: I think line 20, unit record could just be a std::array if it is always the same fixed size which should play nicer with serialization? I'm not quite following what you are proposing for the other 2 vectors though. I'm happy leaving this alone until (if ever) we want to polish reading a saved game. Unless someone wants to research the new size of the vector unitRecord and then I could quickly update that.

Thanks,
Brett

src/Map/MapReader.cpp
DanRStevens: Yeah, I think it's a bit too late to bump the version number for that change. At this point, there will be more new saved game files with the original version number than old saved game files.

It may be possible to distinguish them through other means. Perhaps the field for the number of units, or by using the version tag.

For now, let's just assume the newer version with the expanded unit array and ignore the problem.

I'll update the documentation to add a note. Though I think the SVN is down again, so I won't be able to commit the change yet.

src/Map/SavedGameDataSection2.h
DanRStevens: You're probably right about using a `std::array` for the last field, at least if we only ever want to support the newer version of the saved game files. However, with the other two fields remaining as vectors, there isn't much point. For serialization, it's an all or nothing thing.

I'd say ignore it for now. We use some fixed sized fields, but this isn't really a serialization struct. Each field is serialized individually.

Brett208:
Okay, the placeholder struct makes more sense seeing it in the code. Thanks for fixing up. I went ahead and made UnitRecord a std::array to document that it is a fixed length field. Will merge shortly.

-Brett

DanRStevens:
I have mixed feeling if `unitRecord` should be an `std::array` or a `std::vector`. If we want to be able to support the older saved game file format, then it's not really fixed size anymore, at least not in the compile time sense. Though I'm also not looking to worry about that level of complexity at this time.

On the other hand, if it's fixed at the larger size, we could just do a partial load of data for older saved game files. Plus, there's no room left to expand the bitfields that index into that array, so it's unlikely the array will ever grow to be any larger.

Ok, I think I just talked myself into liking the change.

DanRStevens:
Tagging issue #239.

PR #242, opened 2019-01-19
Rename tile data to tile

### LINKED ISSUE #139: Better names for Tiles

Brett208:
Looking to close #139.

I think since we extracted the global namespace into its own issue, this should currently solve all but one part of #139. 

In the issue it mentions tileIndex being a confusing name. @DanRStevens, when you have some time, could you elaborate a little on what you are looking to fix? Feel free to hijack this branch and make a change if it makes demonstrating easier. 

Thanks,
Brett


DanRStevens:
Ok, I think that addresses all the points of the issue. I noticed I was a bit inconsistent about `tileGraphicIndex` versus `tileGraphicsIndex` (note the "s") in previous discussions. Feel free to adjust if desired.

Brett208:
Thanks for putting time into fixing this up. I'll merge shortly and we can finally push that issue off the list.

-Brett

PR #244, opened 2019-01-21
Separate seek into seek forward and seek backward

### LINKED ISSUE #160: Split SeekRelative to SeekForward and SeekBackward
In the Stream API, perhaps we should split `SeekRelative` to `SeekForward` and `SeekBackward`.

Currently the bounds checking is proving to be a nuisance. Some of that comes from trying to figure out all the corner cases when mixing a signed offset with an unsigned position.

Another difficulty, perhaps more theoretical, is seeking all the way across a long stream, which might need the full unsigned range. By using a signed offset, a relative seek might require two seeks to reach the intended position.

Finally, I think it's generally clear when making a seek call if you intend to go forward or backward, so perhaps that decision should be encoded in the function name, rather than the parameter. This is similar to the decision to split `Seek` from `SeekRelative`, rather than encode the starting position as a parameter (which usually ends up being a hardcoded constant, or default value).

----

Additionally, this might be a good place to break up the interface. Some stream, notably compression/decompression streams, or encryption/decryption streams have complicated state, and transitions that are only intended to go one direction. It may be difficult (decompression) to impossible (decryption) to seek in the opposite direction. It might make sense to have a `ForwardSeekableStream` and a `BidirectionalSeekableStream`.

Brett208:
Should close #160

I didn't spend any time improving or updating the unit tests beyond changing them to bi-directional.


src/Stream/FileReader.cpp
```diff
@@ -60,25 +60,32 @@ namespace Stream
...
 
-			if (offsetAbsValue > Position()) {
-				throw std::runtime_error("Change in offset puts read position before beginning bounds of file.");
-			}
+		if (newPosition < Position()) {
+			throw std::runtime_error("Change in offset puts read position beyond possible bounds of file " + filename);
 		}
 
-		file.seekg(offset, std::ios_base::cur);
+		file.seekg(Position() + offset);
```
DanRStevens: We should probably use `newPosition` here.

src/Stream/FileSliceReader.cpp
```diff
@@ -72,24 +72,33 @@ namespace Stream
...
 				" would place the position of the FileSliceReader outside the slice bounds of file " + fileStreamReader.GetFilename());
 		}
 
-		fileStreamReader.SeekRelative(offset);
+		fileStreamReader.SeekForward(offset);
+	}
+
+	void FileSliceReader::SeekBackward(uint64_t offset)
+	{
+		if (Position() + offset < startingOffset) {
```
DanRStevens: I think this should be `-` here.

src/Stream/BiDirectionalSeekableReader.h
```diff
@@ -1,23 +1,14 @@
...
+	class BiDirectionalSeekableReader : public ForwardSeekableReader {
 	public:
-		// Get the size of the stream
-		virtual uint64_t Length() = 0;
-
-		// Get the current position of the stream
-		virtual uint64_t Position() = 0;
-
 		// Seek to absolute position, given as offset from beginning of stream
 		virtual void Seek(uint64_t position) = 0;
```
DanRStevens: We should probably move absolute seek last. Having random access is more specific than only forward or backward seeking. We could even create a `RandomAccessReader` class if we wanted.

src/Stream/ForwardSeekableReader.h
```diff
@@ -0,0 +1,19 @@
...
+
+namespace Stream
+{
+	class ForwardSeekableReader : public Reader {
+	public:
+		// Get the size of the stream
+		virtual uint64_t Length() = 0;
+
+		// Get the current position of the stream
+		virtual uint64_t Position() = 0;
```
DanRStevens: I suppose technically the `Length` and `Position` attributes are independent of seeking. If we wanted we could have a `FiniteStream` class that has a well defined length and position, but no ability to seek. Not sure how granular we want to get though.

DanRStevens:
Hmm, was just thinking "Bi" probably shouldn't be capitalized separately from "Bidirectional".

src/Stream/FileSliceReader.cpp
```diff
@@ -85,7 +85,7 @@ namespace Stream
 
 	void FileSliceReader::SeekBackward(uint64_t offset)
 	{
-		if (Position() + offset < startingOffset) {
+		if (Position() - offset < startingOffset) {
 			throw std::runtime_error("A backward offset of " + std::to_string(offset) +
```
Brett208: Good catch

DanRStevens:
I think you have a point, we should delay making the change until we have a need for it. Splitting a class interface like that should be a backwards compatible binary change, as long as the order of the methods doesn't change. It may introduce some source incompatibilities though due to the renaming.

Let's defer on further class splitting.

----

I'll leave the Bidirectional rename to you to avoid untested/inconsistent changes in the project files.

Brett208:
@DanRStevens, if this looks good to you, feel free to merge. 

Thanks,
Brett

PR #248, opened 2019-01-27
Switch to vcpkg for gtest

Brett208:
Branch does not work on my computer in debug mode. Took a stab at updating appveyor but not tested.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,13 +1,16 @@
 image: Visual Studio 2017
 configuration:
   - Release
+  - Debug
 platform:
   - x86
   - x64
-cache:
-  # Cache Nuget packages in "packages" folder. Break cache if "packages.config" is updated.
-  - packages -> **\packages.config
-before_build: nuget restore
+install:
+  - cmd: cd C:\Tools\vcpkg
+  - cmd: git pull
+  - cmd: .\bootstrap-vcpkg.bat
+  - cmd: vcpkg install gtest:x86-windows gtest:x64-windows
+  - cmd: cd %APPVEYOR_BUILD_FOLDER%
 build:
   project: OP2Utility.sln
 test_script:
--- a/OP2Utility.vcxproj
+++ b/OP2Utility.vcxproj
@@ -77,6 +77,7 @@
       <SDLCheck>false</SDLCheck>
       <DisableSpecificWarnings>4996;%(DisableSpecificWarnings)</DisableSpecificWarnings>
       <RuntimeLibrary>MultiThreadedDebug</RuntimeLibrary>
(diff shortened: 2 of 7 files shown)


.appveyor.yml
```diff
@@ -1,13 +1,16 @@
...
-cache:
-  # Cache Nuget packages in "packages" folder. Break cache if "packages.config" is updated.
-  - packages -> **\packages.config
-before_build: nuget restore
+install:
+  - cmd: cd C:\Tools\vcpkg
+  - cmd: git pull
+  - cmd: .\bootstrap-vcpkg.bat
+  - cmd: vcpkg install gtest:x86-windows gtest:x64-windows
+  - cmd: cd %APPVEYOR_BUILD_FOLDER%
```
DanRStevens: Hmm, using this tool buries the setup/installation into an AppVeyor specific file. That doesn't make it easy for other developers to download the code from GitHub and get building. It would be very manual to get dependencies installed.

From what I understand, with NuGet, Visual Studio would know how to download and use the dependencies.

----

Small question, why the `git pull`?

Brett208: vcpkg comes preinstalled on appveyor. You could remove the Git pull and maybe some other lines. It just means that you wouldn't have the newest packages when downloading. This would of course improve your build times.

We would have to tell someone how to pull vcpkg and install gtest in the readme. A batch script could be developed since it all happens on the command line.

It takes about 3 console commands, so manual but I wouldn't say very manual.

Once a package is installed, it is available to all Visual Studio projects on the computer. So if you were using gtest in multiple projects, you would only download once.

You might be able to add installing gtest as a prebuild event in Visual Studio to automate. Not sure it is worth it either.

Brett

DanRStevens:
I wanted to bump this PR as the Google Test NuGet package was recently update. Additionally `stdcpp17` has been enabled for the `LanguageStandard` setting. That covers some of the changes done in this PR.

Not addressed, is the use of Google Mock.

I would really like to avoid burdening new developers with complex setup requirements to build and test the project. As `vcpkg` isn't as automated as NuGet for local builds, perhaps we could add a feature detect for Google Mock, and use that to replace the `#ifndef _WIN32` platform detect. I believe we could do this using ['__has_include'](https://en.cppreference.com/w/cpp/preprocessor/include), or perhaps Google Mock itself defines a macro that can be tested.

With such a detect in place, I'd feel more comfortable about using `vcpkg` on AppVeyor to install a Google Mock dependency.

Brett208:
Just adding the gmock nuget package that is officialy released by Google allows us easy, standardized access to gmock on MSVC.

With this revelation, I see no reason to continue experimenting with vcpkg at this time and am closing this branch.

Issue #72, opened 2018-06-21
Stream Slicing


DanRStevens:
I wanted to start a discussion on stream slicing. In particular, I wanted to go over why stream slicing could be helpful, as that has a major impact on how it is designed, implemented, and used.

The major benefit of slicing, is isolation. It's an API enforced boundary to data access.

I think the clearest example of why stream slicing would be useful is to consider a compressed ZIP/RAR/7Z file with various types of files. Each file type may have its own stream parser to handle the data, and those stream parsers may be written later than the ZIP parser, or be unknown to it.

Some of the file parsers may be a bit buggy or ill behaved. In particular, some of them may have been written to only consider the case of reading a single file from disk, and processing the entire contents. There may be no header data to indicate the data length, or the header data is ignored. This might actually be typical behaviour for some movie file parsers, particularly ones that play past the stated end of the file. These parsers might rely on hitting the end of stream to know it is done processing. Others might try to read a file from disk, such as a graphics file format, but the parser is old and can't handle a newer extension to the file format, and so aborts with an error before reading all of the data. Any data unexpectedly left in the stream might have unintended consequences for future stream processing.

There might also be security concerns, where some internal files contain sensitive info, perhaps passwords, and you want to protect that data from unknown parsers that shouldn't have access to it. Perhaps one buggy or ill behaved parser reads past the end of its data, and fails on the contents at the start of the next file, containing passwords, and those passwords end up getting dumped to a publicly readable log file detailing the parse error.

This is why you want isolation. It puts a limit on the damage that bad parsers can do. It prevents problems processing one internal file, from spreading to the processing of other internal files.

Slicing is able to enforce this isolation without doing wasteful intermediate steps, such as extracting to a temporary file and then running the parser on that. It sidesteps the issue of cleaning up temporary files, particularly after bugs, crashes, or power loss.

Side note: Slicing maps nicely over the concept of solid archives. In a solid archive, it's as if all files are concatenated into one big stream before being compressed, with only the index containing information to split that data apart into individual files. In such an archive, the compressed form of a file might not even end on a byte boundary. The compression scheme might use a variable number of bits for various codes, and end 3 bits into a byte, with the remaining 5 bits being part of the compressed stream for the next file. In this case, one solid stream is fed into a decompressor, and the output stream is sliced into files.

Brett208:
[...]

PR #249, opened 2019-02-02
Generalize sliced streams

DanRStevens:
This lays some of the foundation for generalizing stream slicing. The `FileSliceReader` class doesn't really depend much on the specifics of the `FileReader` class. The class could eventually be refactored into a template that handles slices of arbitrary stream types.

This pull request aims to reduce some of the dependence of `FileSliceReader` on the specifics of `FileReader`. Most such dependencies were not particularly fundamental, but rather more in the wording of exception messages and names of variables and types.

This addresses some of the aspects brought up in Issue #72.


DanRStevens:
Ok, I think I've hit a good stopping point for today.

The design is still a little rigid. In particular, the `GetFilename` method is still a little too specific for the template. That might be found on a few different streams, but not every stream. Perhaps that method can be removed, or only conditionally provided if the wrapped stream also provides it. Similarly, the `IdentifySource()` method used in exception messages could be specialized to provide info on other stream types, or as a default, just return an empty string if there is no specific info it knows how to return.

Another point is using `BidirectionalSeekableReader` as the base class. Not all streams will derive from this class. With a bit of work, and some template guards, it may be possible to derive from the same base class as the wrapped stream, and also only provide the appropriate method overrides.

That may be work for a separate branch though.

Brett208:
I read through the commits. It all seems like positive changes to me. I'll carve out time later to download and play with the code and review in more depth.

src/Stream/FileSliceReader.h
```diff
@@ -4,41 +4,165 @@
...
 
 namespace Stream
 {
-	// Opens a new file stream that is limited to reading from the provided file slice.
-	class FileSliceReader : public BidirectionalSeekableReader
+	// Creates a stream which is a slice of an existing stream.
+	// Access is bounds checked according to the slice parameters.
+	// The underlying stream must be copy constructible, so that an independent stream can be created.
+	template<class WrappedStreamType>
+	class StreamSlice : public BidirectionalSeekableReader
```
Brett208: What do you think of renaming this to Stream::ReaderSlice? To me, Stream::StreamSlice would indicate a read and write capability (or just be vague on what it does?).

If this sounds good, I can update the filename and the Visual Studio project files.

DanRStevens: The name probably does need some work. I was purposefully vague on the name since I haven't really decided if this will apply only to read streams, or if it will generalize to write streams as well. I suppose I could just have two templates, though there would be a lot of code in common.

In terms of supporting different levels of the stream interface (Reader, Seekable, BidirectionalSeekable), there may need to be a lot of conditional code added to the template. As such, it may be a simple matter to conditionalize Read/Write, and have the template work for both stream types.

With that said, you're welcome to rename things. We can always rename again later if we come up with a better name.

DanRStevens:
Hmm, I probably should have put more thought into this earlier before asking you to rename. Instead of `ReaderSlice`, maybe we should have used the name `SliceReader`. I was just looking at the other class names, and I think `SliceReader` would be more consistent.

Another point of clarification, I think we renamed the wrong thing. The template is currently named `StreamSlice`, and I think that construct should get the name `SliceReader`:
```cpp
	template<class WrappedStreamType>
	class SliceReader : public BidirectionalSeekableReader
```

Meanwhile, the specific instantiation of that template, which wraps a `FileReader` should probably still be named `FileSliceReader`.
```cpp
	using FileSliceReader = SliceReader<FileReader>;
```

The name of the header should probably be named after the template, hence `SliceReader.h`.

----

The specific instance `using FileSliceReader = SliceReader<FileReader>;` doesn't really need to be in the template's header file. I just put it there for backwards compatibility reasons during the transition period. It could get it's own header file, or it could just be declared where it is used, namely in `FileReader.h`, where there is already a forward declare for it. Actually, I think that forward declaration is sufficient. We should probably just remove the `using FileSliceReader = ...` line from the template header file.

----

Sounds like we should just largely ignore the stream size concern. You're right in that stream sizes can always change outside the control of our API. Namely, the filesystem API can change file sizes which we have readers for. We just have to accept the possibility of this happening.

I say let's just stick with what we have for now. What we have now fits our needs. Even if stream resizing was a bigger concern, I don't think it's a show stopper for the current API.

Brett208:
Okay, no problem with the renames. I can try to fix this hopefully today or tomorrow. I probably didn't put enough time into thinking through the template verses the class.

- Brett

PR #250, opened 2019-02-06
Fix GetDirectory

DanRStevens:
Seems I wrote a bug into `GetDirectory` when I (unconditionally) added the trailing "/". I also missed the test case that checks only a filename with no path component, which highlights this bug.

I believe this is the root cause of the problem from PR #243.


Brett208:
Oh cool. I'm glad that you found this. 

I wrote the response on the other PR before reading this. I should have time later today (maybe tomorrow if I can't fence off the time) to test this and report back on what MSVC does.

PR #253, opened 2019-02-08
Strip directory from GetFilenamesFromDirectory Return

### LINKED ISSUE #252: Should XFile::GetFilesFromDirectory return full paths or only filenames
Based on use case in OP2MapImager, it would be better if this function returned the filename only and not the full path. IE if passed the directory C://test, just return ashes.map instead of C://test/ashes.map.

Maybe better if we rename the function GetFilenamesFromDirectory to be more specific in what it does.

Brett208:
Closes #252

Also add usage comment to GetFilenamesFromDirectory

There was some discussion of recursively searching for filenames in the issue. I think it is a good idea in generally, but I don't have a current use case, so I was just proceeding without it.

Thanks,
Brett

### DIFF
--- a/src/XFile.cpp
+++ b/src/XFile.cpp
@@ -75,22 +75,22 @@ std::string XFile::AppendToFilename(const std::string& filename, const std::stri
 	return newPath.string();
 }
 
-std::vector<std::string> XFile::GetFilesFromDirectory(const std::string& directory)
+std::vector<std::string> XFile::GetFilenamesFromDirectory(const std::string& directory)
 {
 	// Brett208 6Aug17: Creating a path with an empty string will prevent the directory_iterator from finding files in the current relative path.
 	auto pathStr = directory.length() > 0 ? directory : "./";
 
 	std::vector<std::string> filenames;
 	for (const auto& entry : fs::directory_iterator(pathStr)) {
-		filenames.push_back(entry.path().string());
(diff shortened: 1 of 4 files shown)


src/XFile.cpp
```diff
@@ -75,22 +75,22 @@ std::string XFile::AppendToFilename(const std::string& filename, const std::stri
...
-std::vector<std::string> XFile::GetFilesFromDirectory(const std::string& directory)
+std::vector<std::string> XFile::GetFilenamesFromDirectory(const std::string& directory)
 {
 	// Brett208 6Aug17: Creating a path with an empty string will prevent the directory_iterator from finding files in the current relative path.
 	auto pathStr = directory.length() > 0 ? directory : "./";
 
 	std::vector<std::string> filenames;
 	for (const auto& entry : fs::directory_iterator(pathStr)) {
-		filenames.push_back(entry.path().string());
+		filenames.push_back(GetFilename(entry.path().generic_string()));
```
DanRStevens: Ahh yes, this would be the simple way to do it for the non-recursive case. For some reason I was expecting this line to be a little more complicated. In the past, I've done this by cropping the first `n` characters from the path, where `n` is the number of characters in the search base folder. That would work for the recursive case. It does however have the assumption that all paths returned are prefixed with the search base folder, which is generally true.

PR #254, opened 2019-02-09
Fix resource manager constructor bug

Brett208:
Hopefully this is the last bug to fix before releasing OP2MapImager. There always seems to be another problem with finding filenames/directories in that project.


DanRStevens:
Looks like you've been busy. As I'm eager to get OP2MapImager working with all the recent bug fixes, I'm going to go ahead and merge this.

PR #256, opened 2019-02-10
Allow custom pitch when writing bitmaps

Brett208:
This is a pretty extensive rollup of changes. I've been playing with extracting images from op2_art. These changes were a result of that work.

Doesn't need to be reviewed quickly.

-Brett

### DIFF
--- a/src/Bitmap/BitmapFile.cpp
+++ b/src/Bitmap/BitmapFile.cpp
@@ -38,9 +38,9 @@ void BitmapFile::VerifyPixelSizeMatchesImageDimensionsWithPitch() const
 	BitmapFile::VerifyPixelSizeMatchesImageDimensionsWithPitch(imageHeader.bitCount, imageHeader.width, imageHeader.height, pixels.size());
 }
 
-void BitmapFile::VerifyPixelSizeMatchesImageDimensionsWithPitch(uint16_t bitCount, int32_t width, int32_t height, std::size_t pixelsWithPitchSize)
+void BitmapFile::VerifyPixelSizeMatchesImageDimensionsWithPitch(uint16_t bitCount, std::size_t pitch, int32_t height, std::size_t pixelsWithPitchSize)
 {
-	if (pixelsWithPitchSize != ImageHeader::CalculateDefaultPitch(bitCount, width) * std::abs(height)) {
+	if (pixelsWithPitchSize != pitch * std::abs(height)) {
 		throw std::runtime_error("The size of pixels does not match the image's height time pitch");
 	}
 }
--- a/src/Bitmap/BitmapFile.cpp
+++ b/src/Bitmap/BitmapFile.cpp
@@ -7,7 +7,7 @@ BitmapFile BitmapFile::CreateDefaultIndexed(uint16_t bitCount, uint32_t width, u
 	BitmapFile bitmapFile;
 	bitmapFile.imageHeader = ImageHeader::Create(width, height, bitCount);
 	bitmapFile.palette.resize(bitmapFile.imageHeader.CalcMaxIndexedPaletteSize());
-	bitmapFile.pixels.resize(bitmapFile.imageHeader.CalculatePitch() * height);
+	bitmapFile.pixels.resize(bitmapFile.imageHeader.CalculateDefaultPitch() * height);
 
 	const std::size_t pixelOffset = sizeof(BmpHeader) + sizeof(ImageHeader) + bitmapFile.palette.size() * sizeof(Color);
 	const std::size_t bitmapFileSize = pixelOffset + bitmapFile.pixels.size() * sizeof(uint8_t);
@@ -40,7 +40,7 @@ void BitmapFile::VerifyPixelSizeMatchesImageDimensionsWithPitch() const
 
 void BitmapFile::VerifyPixelSizeMatchesImageDimensionsWithPitch(uint16_t bitCount, int32_t width, int32_t height, std::size_t pixelsWithPitchSize)
 {
-	if (pixelsWithPitchSize != ImageHeader::CalculatePitch(bitCount, width) * std::abs(height)) {
+	if (pixelsWithPitchSize != ImageHeader::CalculateDefaultPitch(bitCount, width) * std::abs(height)) {
 		throw std::runtime_error("The size of pixels does not match the image's height time pitch");
 	}
 }
--- a/src/Bitmap/BitmapFile.h
+++ b/src/Bitmap/BitmapFile.h
@@ -22,14 +22,15 @@ public:
 	std::vector<Color> palette;
 	std::vector<uint8_t> pixels;
 
+	// Uses a default 4 byte pitch
 	static BitmapFile CreateDefaultIndexed(uint16_t bitCount, uint32_t width, uint32_t height);
 
-	// BMP Reader only supports Indexed Color palettes (1, 2, and 8 bit BMPs).
+	// BMP Reader only supports indexed color palettes (1, 2, and 8 bit BMPs).
 	static BitmapFile ReadIndexed(const std::string& filename);
 	static BitmapFile ReadIndexed(Stream::BidirectionalSeekableReader& seekableReader);
 
-	// BMP Writer only supporting Indexed Color palettes (1, 2, and 8 bit BMPs).
-	// @indexedPixels: Must include padding to fill each image row out to the next 4 byte memory border (pitch).
+	// BMP Writer only supports indexed color palettes (1, 2, and 8 bit BMPs).
+	// @indexedPixels: Must include padding to fill each image row out to the next byte memory border (pitch). 4 byte pitch is typical.
 	static void WriteIndexed(std::string filename, uint16_t bitCount, int32_t width, int32_t height, std::vector<Color> palette, const std::vector<uint8_t>& indexedPixels);
 	static void WriteIndexed(Stream::BidirectionalSeekableWriter& seekableWriter, uint16_t bitCount, int32_t width, int32_t height, std::vector<Color> palette, const std::vector<uint8_t>& indexedPixels);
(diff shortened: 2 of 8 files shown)


Brett208:
Forgot to commit unit tests and out of time for now. Will try and fix tomorrow.

-Brett

src/Sprite/OP2BmpLoader.cpp
```diff
@@ -29,7 +31,31 @@ void OP2BmpLoader::ExtractImage(std::size_t index, const std::string& filenameOu
...
+{
+	std::vector<Color> palette;
+
+	if (imageMeta.GetBitCount() == 1) {
+		palette.push_back(Color{ 0, 0, 0 });
+		palette.push_back(Color{ 255, 255, 255 });
+	}
+	else {
+		palette.resize(artFile.palettes[imageMeta.paletteIndex].size());
+		std::copy(artFile.palettes[imageMeta.paletteIndex].begin(), artFile.palettes[imageMeta.paletteIndex].end(), palette.begin());
```
DanRStevens: I haven't looked at related code yet, but my first thought upon reading this, is why are the palettes copied, rather than using a read-only reference to an existing palette?

Related, the monochrome palette could reference a global/static `PaletteMonochromeWhiteBlack` palette.

src/Bitmap/BitmapFile.h
```diff
@@ -22,14 +22,15 @@ class BitmapFile
...
 
-	// BMP Reader only supports Indexed Color palettes (1, 2, and 8 bit BMPs).
+	// BMP Reader only supports indexed color palettes (1, 2, and 8 bit BMPs).
 	static BitmapFile ReadIndexed(const std::string& filename);
 	static BitmapFile ReadIndexed(Stream::BidirectionalSeekableReader& seekableReader);
 
-	// BMP Writer only supporting Indexed Color palettes (1, 2, and 8 bit BMPs).
-	// @indexedPixels: Must include padding to fill each image row out to the next 4 byte memory border (pitch).
+	// BMP Writer only supports indexed color palettes (1, 2, and 8 bit BMPs).
+	// @indexedPixels: Must include padding to fill each image row out to the next byte memory border (pitch). 4 byte pitch is typical.
```
DanRStevens: This comment sounds a bit confusing. I think it's made more confusing because the parameters don't allow you to specify the pitch. I assume the pitch comes from the `BmpHeader` struct stored as a member variable of the class.

Edit: Ok, I see the `pitch` parameter was added later. The comment is still a little confusing though. Maybe phrase as:
"@indexedPixels: Size must be padded to a multiple of pitch. (4 byte pitch is typical)"

It would also help to be consistent about specifying pitch as either bytes or bits. I thought I saw somewhere else use 32-bits, whereas here it's using 4 bytes.

src/Bitmap/ImageHeader.h
```diff
@@ -42,8 +42,9 @@ struct ImageHeader
 	void VerifyValidBitCount() const;
 	static void VerifyValidBitCount(uint16_t bitCount);
 
-	std::size_t CalculatePitch() const;
-	static std::size_t CalculatePitch(uint16_t bitCount, int32_t width);
+	// Default pitch interval is 32
```
DanRStevens: Suggested phrasing: "Default pitch granularity is 32 bits"
Edit: or rather, 4 bytes

src/Bitmap/BitmapFile.h
Brett208: I'll update the comment as suggested. I'll try to standardize on using the term 4 bytes for standard. The variable takes bytes, so it makes more sense to reference bytes in the comments to not confuse the issue.

-Brett

src/Sprite/OP2BmpLoader.cpp
Brett208: I added a default monochrome palette. Good idea.

I tried adding the ability to return a const reference to a palette to ArtFile.h. Code doesn't seem to want to compile:

```C++
const std::vector<Color>& ArtFile::GetPalette(std::size_t index) const 
{
	if (index >= palettes.size()) {
		throw std::runtime_error("Passed index does not correspond to a palette (too large)");
	}

	return palettes[index];
}
```

Error (active)	E0312	no suitable user-defined conversion from "const Palette" to "const std::vector<Color, std::allocator<Color>>" exists	OP2Utility	\OP2Utility\src\Sprite\ArtFile.cpp	12

Ran out of time to troubleshoot.

Thanks,
Brett

DanRStevens: You probably need to use `const std::vector<const Color>&` as the return type. Note the `const` within the `<>`.

Brett208: Yeah, I realized this yesterday as well. I think we should just stick to copy for now. This branch is ballooning too much. We can write an issue to decide what to do with the Palette type which will then allow taking a reference if we want.

src/Bitmap/BitmapFile.cpp
```diff
@@ -35,16 +35,36 @@ void BitmapFile::VerifyIndexedPaletteSizeDoesNotExceedBitCount(uint16_t bitCount
...
+	return FindPitch(imageHeader.width, imageHeader.height, pixels.size());
+}
+
+std::size_t BitmapFile::FindPitch(std::size_t width, std::size_t height, std::size_t pixelCount)
+{
+	if (pixelCount % height != 0) {
+		throw std::runtime_error("Unable to calculate a valid pitch based on height and pixel count");
+	}
+
+	const std::size_t pitch = pixelCount / height;
```
Brett208: This hack of mine is causing the unit tests to fail. They happen to be fairly robust for this small section of the code compared to others.

The next step is probably to figure out a proper bit twiddle or calculation that determines pitch without assuming it is a 4 byte interval. @DanRStevens, feel free to chime in if you have something up your sleeve.

-Brett

src/Bitmap/ImageHeader.h
```diff
@@ -42,8 +42,9 @@ struct ImageHeader
 	void VerifyValidBitCount() const;
 	static void VerifyValidBitCount(uint16_t bitCount);
 
-	std::size_t CalculatePitch() const;
-	static std::size_t CalculatePitch(uint16_t bitCount, int32_t width);
+	// Default pitch granularity is 4 bytes
```
Brett208: I had intentionally left this as interval. Perhaps it is a regional language different but interval seems a better and easier to understand choice than granularity to me. Being a comment, it probably isn't worth arguing over now that it has been changed.

DanRStevens: A disagreement in terminology might signal a disagreement about what is intended here.

I would define an interval as having a start and end point. Granularity is about the minimum spacing of values.

Here the pitch is rounded to some multiple of a certain value. That is defining the granularity of the value. As in, values are a multiple of 4 bytes. Is does not define the length of an interval as 4 bytes. For example, the spacing between scanlines could easily be 12 bytes, even though the granularity of that value is 4 bytes.

More deeply, I'm not sure I would really consider pitch to be an interval. It's more of a spacing between scanlines. If pitch were an interval, I'd be uncertain how to define the start and end points.

Looking back at `std::size_t CalculateDefaultPitch() const;`, I'm actually a little uncertain what the comment is trying to tell me, or why it is important. Is it trying to tell me the value is a multiple of 4 bytes? From the example test code, that would seem to be the case.

I vaguely recall some requirement that pitch is a minimum of 4 bytes, though can be a larger multiple of it.

I might also consider the phrasing:
Default pitch is the scanline byte width rounded up to a multiple of 4 bytes

src/Bitmap/BitmapFile.cpp
DanRStevens: Pitch is either assumed, or given. It's not generally calculated (other than in an assumption, such as rounding the scanline byte width up to a multiple of 4 bytes). The problem with trying to calculate it, is the data source could always have chosen a larger pitch than is necessary.

I find it a little non-intuitive to use `pixelCount` here. When I think pixel could, I generally assume it's simply `width * height`, which is not particularly useful for calculating `pitch` if you already have the width. Plus, `pitch` is a byte measurement, not a pixel measurement. The `bitDepth` needs to be accounted for.

Brett208: If you are wanting to work on this branch feel free. I'm not going to pick it up again for a long time.

-Brett

src/Bitmap/IndexedBmpReader.cpp
```diff
@@ -62,8 +62,9 @@ void BitmapFile::ReadPalette(Stream::BidirectionalSeekableReader& seekableReader
 
 void BitmapFile::ReadPixels(Stream::BidirectionalSeekableReader& seekableReader, BitmapFile& bitmapFile)
 {
-	std::size_t pixelContainerSize = bitmapFile.bmpHeader.size - bitmapFile.bmpHeader.pixelOffset;
-	BitmapFile::VerifyPixelSizeMatchesImageDimensionsWithPitch(bitmapFile.imageHeader.bitCount, bitmapFile.imageHeader.width, bitmapFile.imageHeader.height, pixelContainerSize);
+	const std::size_t pixelContainerSize = bitmapFile.bmpHeader.size - bitmapFile.bmpHeader.pixelOffset;
```
DanRStevens: I think this might be more of a maximum upper bound, rather than an exact size. There may be other sections stored in the bitmap file, after the pixel data, before the end of the file.

src/Bitmap/BitmapFile.cpp
DanRStevens: I'm not sure I quite have the context to pick this one up right now.

Brett208:
I am attempting to pull some of the better ideas out of this branch and commit them as smaller, separate branches and then close this pull request.

I think the custom pitch thrust itself went poorly, but there was some supporting code that I didn't want to toss when terminating the branch.

PR #263, opened 2019-02-13
Add DynamicMemoryWriter class

DanRStevens:
I'm still working on the test code for this one, but I thought I'd get the code up early for review purposes.

I was at first thinking of just making this a `Writer`, then thought a `ForwardWriter` would also be easy, and slightly more useful. Then I thought why not make it `BidirectionalWriter`, though I took one slight shortcut in the implementation, where seeking sets the length of the stream (either by truncating, or 0 filling).

----

This writes to a dynamically allocated and resizable memory buffer. After writing, the `GetReader()` method will return a `MemoryReader` object with a view of the memory.

This class can be used to aid writing test code. Writing can be done in-memory, and then verified with the associated `MemoryReader` object. There is no need to write to a slow disk, nor any need to clean up temporary files. This makes writing test code simpler and faster.

### DIFF
--- a/src/Stream/DynamicMemoryWriter.h
+++ b/src/Stream/DynamicMemoryWriter.h
@@ -17,11 +17,11 @@ namespace Stream
 		uint64_t Length() override;
 		uint64_t Position() override;
 
-		// Note: SeekForward will set a new stream size and 0 fill the buffer gap
+		// SeekForward will set a new stream size and 0 fill the buffer gap
 		void SeekForward(uint64_t offset) override;
-		// Note: SeekBackward will set a new stream size and truncate the stream
+		// SeekBackward will set a new stream size and truncate the stream
 		void SeekBackward(uint64_t offset) override;
-		// Note: Seek will set a new stream size (either by 0 filling or by truncating)
+		// Seek will set a new stream size (either by 0 filling or by truncating)
 		void Seek(uint64_t position) override;
 
 		// Get read access to the written data
--- a/src/Stream/DynamicMemoryWriter.h
+++ b/src/Stream/DynamicMemoryWriter.h
@@ -8,6 +8,14 @@
 
 namespace Stream
 {
+	/**
+	 * \brief Memory backed stream, which grows dynamically in size as data is added
+	 *
+	 * An internal memory buffer is managed by DynamicMemoryWriter, which grows as needed to accommodate new data.
+	 * This is in contrast to MemoryWriter, which is a view to a non-owned pre-set sized buffer, which can not grow.
+	 *
+	 * \note Adding additional data to the stream after calling GetReader may invalidate existing readers.
+	 */
 	class DynamicMemoryWriter : public BidirectionalWriter
 	{
 	public:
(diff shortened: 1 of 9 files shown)


src/Stream/DynamicMemoryWriter.h
```diff
@@ -0,0 +1,36 @@
...
+	class DynamicMemoryWriter : public BidirectionalSeekableWriter
+	{
+	public:
+		DynamicMemoryWriter();
+		DynamicMemoryWriter(std::size_t preallocateSize);
+
+		uint64_t Length() override;
+		uint64_t Position() override;
+
+		// Note: SeekForward will set a new stream size and 0 fill the buffer gap
```
Brett208: What do you think about pulling 'Note:' out of these comments? I think it is superfluous in context.

DanRStevens: You're probably right. I do overuse "Note:" a bit.

DanRStevens:
Ok, I've added the tests I wanted to add. I also updated a few existing test cases you use the `DynamicMemoryWriter`. Two cases for update were:
1) Using a `MemoryWriter` with a statically sized buffer.
2) Using a `FileWriter`, and then deleting the file afterwards. (But not explicitly testing a `filename` path)

I think this is ready to merge now.

Brett208:
Should have time to review later today or tomorrow. Sorry for the excessive delay. Been trying to focus on NetFix/OP2Internal.

src/Stream/DynamicMemoryWriter.h
```diff
@@ -0,0 +1,36 @@
...
+
+#include "BidirectionalWriter.h"
+#include "MemoryReader.h"
+#include <cstddef>
+#include <cstdint>
+#include <vector>
+
+namespace Stream
+{
+	class DynamicMemoryWriter : public BidirectionalWriter
```
Brett208: Recommend adding a short description right above the class declaration explaining what the DynamicMemoryWriter does and why. Doesn't need to get into specifics and details, but I don't feel like DynamicMemoryWriter by itself fully explains the purpose of the class (and I don't know that it can be explained succinctly by just the name of the class?)

DanRStevens: Good idea about leaving some documentation.

DanRStevens:
Thanks for the updates. Just realized that although the AppVeyor build appeared to have passed for the code I'd pushed, it never actually ran the new tests since they hadn't been added to the test project file yet.

----

I've made some slight updates to the `static_cast` code to use a type consistent with the range check. It's possible for a standard container to use a type other than `std::size_t`, so they have their own `size_type` type alias, though it's usually just an alias for `std::size_t`.

I was tempted to update the type in the `DynamicMemoryWriter` constructor as well. The `preallocateSize` parameter still uses `std::size_t`. We could try to DRY things up a bit more, though not without moving the private member variable ahead of the constructor, which I didn't want to do. Probably time to move on.

Brett208:
Thanks for the documentation on Dynamic Memory Writer. Good point on size_type.

PR #269, opened 2019-02-24
Fix static_assert

DanRStevens:
A non-zero value is true, which allowed this static_assert to pass without actually checking what it intended to check.

----

Related to this change, though not because of this change, I think the `static_assert` checks on struct sizes used in serialization should always use a literal constant for the size, rather than a constant expression which sums the component parts. The single literal is easier to reason about.

### DIFF
--- a/src/Sprite/Animation.h
+++ b/src/Sprite/Animation.h
@@ -23,7 +23,7 @@ struct Animation {
 			Point16 pixelOffset;
 		};
 
-		static_assert(4 + sizeof(Point16), "Animation::Frame::Layer is an unexpected size");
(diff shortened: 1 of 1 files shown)


DanRStevens:
That is true, yes. Though as we are dealing with serialization of binary files, there are strong implications if the size of a substructure changes. Additionally, we don't control the source format, and simply have to build to it, so it's unlikely we would change the size of a substructure (which should also have it's own static size check).

I suppose maybe my preference comes from looking at assembly code, where the size of the entire aggregate is often known well before the constituent parts.

Issue #135, opened 2018-08-08
Improve StreamReader/Writer static checks


DanRStevens:
We can improve the static checks for the StreamReader and StreamWriter template methods. In particular, the length prefixed data types can use some bounds checking on the length. The checks depend on Read/Write, as well as the relative data sizes between the chosen type and `std::size_t`.

For reading:
With signed length values, negative values should result in an exception.
For positive values (either signed or unsigned types), exceeding the bounds of a `std::size_t` should result in an exception.

For writing:
No bounds checking is needed for negative values, as an object can never have negative size.
If the size of the object (bounded by the range of `std::size_t`) exceeds the range of the chosen length type, an exception should be raised.

Depending on the type parameter, either case may be impossible. It is desirable to not incur any runtime cost for these checks when the condition is impossible. For instance, a `uint32_t` can never be negative, and can never overflow the bounds of a `std::size_t`.

It may be possible to implement this with template specialization using some combination of the following methods:

From `<type_traits>`:
[`std::integral_constant`](https://en.cppreference.com/w/cpp/types/integral_constant)
[`std::is_signed`](https://en.cppreference.com/w/cpp/types/is_signed)
[`std::is_unsigned`](https://en.cppreference.com/w/cpp/types/is_unsigned)

From `<numeric_limits>`:
[`std::numeric_limits::is_integer`](https://en.cppreference.com/w/cpp/types/numeric_limits/is_integer)
[`std::numeric_limits::is_signed`](https://en.cppreference.com/w/cpp/types/numeric_limits/is_signed)

DanRStevens:
Saw this article about `if constexpr(expression)` in C++17:
https://www.codingame.com/playgrounds/2205/7-features-of-c17-that-will-simplify-your-code/constexpr-if

This could be used to ensure a zero-cost runtime check when `SizeType` parameters are set such that overflow is impossible, or such that negative values are impossible.

This could be used to wrap the existing runtime checks, such that they are only present in the compiled code if some `SizeType` range check determines a runtime check is necessary.

```cpp
if constexpr(compileTimeSizeTypeRangeCheck) {
  if (existingRuntimeOverflowCheck) {
    throw std::runtime_error("Overflow ...");
  }
}
```

Here's some additional update info detailing `if constexpr(...)`:
https://en.cppreference.com/w/cpp/language/if

PR #270, opened 2019-02-26
Add stream size checks

DanRStevens:
This addresses Issue #135.

Ideally, some test code should be added here to exercise the overflow checks.

Additional checks were added for Reader. Writer checks were already sufficient. Only commenting was adding for Writer.

Not directly addressed in the issue is if trivially false checks are zero cost at runtime. That may require examining assembly output to determine. It may be true, but hasn't been verified.

### DIFF
--- a/src/Stream/Reader.h
+++ b/src/Stream/Reader.h
@@ -43,6 +43,7 @@ namespace Stream
 		// passing vector into this function to ensure proper vector size is read
 		template<typename T, typename A>
 		inline void Read(std::vector<T, A>& vector) {
+			// Size calculation can't possibly overflow since the vector size necessarily fits in memory
 			ReadImplementation(vector.data(), vector.size() * sizeof(T));
 		}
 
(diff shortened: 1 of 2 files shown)


DanRStevens:
I did some checking of assembly output with Clang, to determine if the checks are indeed zero-cost when they're not actually needed. I noticed the string for negative values check appeared in the output assembly listing using default build settings. It didn't seem to be used in the main code path though. Turning on optimizations with `-O2`, or guarding the if statement using a constexpr-if (and no optimization) caused the string to be omitted from the assembly output.

I'm thinking perhaps constexpr-if guards should be added around some of the checks.

To enable constexpr-if checks, we should switch compiler settings from C++14 to C++17.

Should such checks be added? Should they be added now before merging, or done as a new branch?

Brett208:
Our current nuget package for gtest will fail if we switch to C++17 explicitly in the MSVC build. Maybe preprocessor guard it for Linux only until we figure out a different gtest solution or just push an issue for it.

Brett

DanRStevens:
Ok, I created issue #275 to track possible future work here. I think we can merge what we have now.

PR #280, opened 2019-03-07
Add guard to vector write in Stream code

### LINKED ISSUE #276: Stream code should not direct serialize vectors of non-trivial types
For reference, the following `enable_if` check is used to only directly serialize trivial objects:
```cpp
		// Trivially copyable data types
		template<typename T>
		inline std::enable_if_t<std::is_trivially_copyable<T>::value> Write(const T& object) {
			WriteImplementation(&object, sizeof(object));
		}
```

However, when serializing a vector, currently no check is done for the underlying type of the vector:
```cpp
		// Vector data types
		template<typename T, typename A>
		inline void Write(const std::vector<T, A>& vector) {
			// Size calculation can't possibly overflow since the vector size necessarily fits in memory
			WriteImplementation(vector.data(), vector.size() * sizeof(T));
		}
```

This could potentially fail in bad ways for vectors of non-trivial types. We should add an appropriate `enable_if` guard on the vector's underlying type `T`, to ensure it is trivially-copyable.

In particular, this unguarded method should fail for `std::vector<std::string>` (unless the strings happen to all use the short-string optimization).

----

See the proposals in Issue #277 and Issue #278 for how we can serialize vectors of non-trivial types.

DanRStevens:
This fixes #276.

Trying to pass a vector of non-trivially-copyable types will now result in a compile time error.

### DIFF
--- a/src/Stream/Writer.h
+++ b/src/Stream/Writer.h
@@ -31,13 +31,17 @@ namespace Stream
 
 		// Trivially copyable data types
 		template<typename T>
-		inline std::enable_if_t<std::is_trivially_copyable<T>::value> Write(const T& object) {
+		inline
+		std::enable_if_t<std::is_trivially_copyable<T>::value>
+		Write(const T& object) {
 			WriteImplementation(&object, sizeof(object));
 		}
 
 		// Vector of trvially copyable data types
 		template<typename T, typename A>
-		inline std::enable_if_t<std::is_trivially_copyable<T>::value> Write(const std::vector<T, A>& vector) {
+		inline
+		std::enable_if_t<std::is_trivially_copyable<T>::value>
+		Write(const std::vector<T, A>& vector) {
 			// Size calculation can't possibly overflow since the vector size necessarily fits in memory
 			WriteImplementation(vector.data(), vector.size() * sizeof(T));
 		}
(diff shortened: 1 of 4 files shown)


DanRStevens:
Having unit tests to exercise the code being edited would be good.

In terms of this change, you can't really write a unit test for it, since the addition is a compile time check. If you failed the check, the code wouldn't compile. In particular, if unit tests attempted to fail this check, the unit test code wouldn't compile.

I suppose there are template based SFINAE ways of checking that certain code doesn't compile, so maybe it can be done, but it would require extra support code and it won't be pretty. It would also likely take days of research and testing. If we really want to do this, we should probably open an issue and defer it.

Additionally, `Stream::Writer` is an abstract class, so you could only test this change through a derived class. Perhaps `DynamicMemoryWriter` would be a good candidate to test from.

We should at least make sure the code is exercised for the happy path. The happy path being serializing of trivial data types. That's likely already covered indirectly in other unit tests. We should probably have a direct unit test though, as that would provide better error messages if something went wrong.

DanRStevens:
I was thinking about the SFINAE check that something wouldn't compile, and realized that would be silly, since some of the other open changes are about ways to support serializing vectors of non-trivial types. Hence, code will compile and run once those changes are complete.

For now I've added some test code that exercises the happy path.

Issue #277, opened 2019-03-07
Extend Stream library to support non-trivially-copyable objects


DanRStevens:
I noticed there are a number of structs being used for serialization, but due to complex members, such as `std::vector`, they are not `std::is_trivially_copyable`. As such, they can not be automatically serialized.

To get around this, such objects often have a custom `Write(Stream::Writer&)` style method, which is manually called. I was thinking perhaps the Stream library could be updated to support auto calling such a member function on non-trivially-copyable objects. This would make serialization of such complex objects more consistent with traditional trivially-copyable objects.

Example:
```cpp
struct ComplexObjectType {
  std::vector<std::string> nonTrivialField;

  Write(Stream::Writer& writer);
};

void f(Stream::Writer& writer) {
  ComplexObjectType object;
  writer.Write(object);  // Auto calls object.Write(writer)
}
```

----

On it's own, this small syntax adjustment doesn't buy much. However, combine this with the update proposed in #276, and we would be able to serialize `std::vector` of non-trivial types, provided they define an appropriate `Write(Stream::Writer& writer)` method.

DanRStevens:
This article has some details about how to detect the presence of a given member:
https://en.wikibooks.org/wiki/More_C%2B%2B_Idioms/Member_Detector

I was hoping there would be a built in [`type_traits`](https://en.cppreference.com/w/cpp/header/type_traits) method, but I'm not seeing an exact match. The closest looking one I see is [`is_member_function_pointer`](https://en.cppreference.com/w/cpp/types/is_member_function_pointer). Maybe it can be used.

Brett208:
I generally support this and #278. 

In map, we have both a WriteMap and WriteSavedGame. I'm guessing that class would not be a candidate for this improvement unless we separated into two classes, Map and SavedGame.

Also, for Bitmap, we have named the write function WriteIndexed. I like this because it explicitly tells the user they cannot write a non-indexed bitmap. This function could just be renamed to write, and a comment added about only supporting index bitmaps. 

In C# you can decorate functions although I haven't researched the concept in C++. There might be a way to decorate a different named function to serve as the 'Write' function, although I don't know if we want to support it or not. It starts adding more complexity to the project since less people will be familiar with decorations.

The proposed syntax will certainly improve the readability when reading/writing.

-Brett

DanRStevens:
[...]

PR #282, opened 2019-03-07
Allow Write of non-trivial types with Write method

DanRStevens:
Non-trivial data types may supply their own Write method.

This addresses Issue #277. Test code should be written to demonstrate this capability before it is merged in. Tests should also verify the new method does not conflict with existing uses of `Write`. The `enable_if` guards may need to be checked and tightened up. In particular, it's not currently checking the exact signature of the `Write` method.


DanRStevens:
Ahh yes, thank you for poking this. It's been left open for a very long time. I don't really remember what it was waiting on anymore.

Brett208:
I added a basic test to prove that it works properly. I'm ok if it merges now. Could probably be some more rigorous testing but I struggled to design the tests.

DanRStevens:
Ok, I've had a chance to re-read the associated issue. I think I'm fine with merging this now. I had wanted to tighten up the `enable_if` expression, though it should be good enough for now. We can always update later. Besides, as C++20 becomes mainstream, the new concepts feature may provide a better easier way to tighten up when the `Write` method can be used.

PR #285, opened 2019-03-09
Add rvalue read maps

Brett208:
The branch name is incorrect. It is for Reading maps/saved games, not writing.

We don't have a default way to create a saved game, so it isn't really tested yet. I think this is okay since we don't actually support fully reading/writing saved games.

Inspired by @DanRStevens work on Rvalue in a different branch.

-Brett

### DIFF
--- a/test/Map/MapReader.test.cpp
+++ b/test/Map/MapReader.test.cpp
@@ -1,4 +1,5 @@
 #include "Map/Map.h"
+#include "Stream/DynamicMemoryWriter.h"
 #include <gtest/gtest.h>
 #include <string>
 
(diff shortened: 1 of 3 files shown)


test/Map/MapReader.test.cpp
```diff
@@ -15,3 +16,20 @@ TEST(MapReader, EmptyFile) {
 	EXPECT_THROW(Map::ReadMap("Map/data/EmptyMap.map"), std::runtime_error);
 	EXPECT_THROW(Map::ReadSavedGame("Map/data/EmptySave.OP2"), std::runtime_error);
 }
+
+TEST(MapReader, ReadMap) {
+	// Simple valid source to load from
+	Stream::DynamicMemoryWriter writer;
+	Map map;
+	map.Write(writer);
```
DanRStevens: I'd suggest shortening to:
```cpp
	Map().Write(writer);
```

Brett208: Good catch.

Issue #278, opened 2019-03-07
Extend Stream library to support serializing vectors of non-trivial types


DanRStevens:
Instead of trying to copy the whole vector to the stream at once (only possible for non-trivial types), we could instead iterate over each element of the vector, and serialize the element individually. This is possible if there is an appropriate `Write(T& object)` method. Such is already the case for `std::string`, as the Stream library natively supports this. The code of course generalizes to any non-trivial type, given the existence of an appropriate serialization method.

Example vector write method for non-trivially-copyable objects:
```cpp
  Write(const std::vector<T, A>& vector) {
    for (const auto element : vector) {
     writer.Write(element);  // Call custom serializer for element type
    }
  }
```

Example usage:
```cpp
void f(Stream::Writer& writer) {
  std::vector<std::string> nonTrivial;
  // ...
  writer.Write(nonTrivial);
}
```

In the above example, the serialization code would loop over the vector's elements, and call the custom `std::string` serialization code for each element.

Combine this with the proposal in Issue #277, and it would also add support for other arbitrary non-trivial types, provided they supply an appropriate `T::Write(Stream::Writer& writer)` method.

DanRStevens:
I was just thinking, the above method generalizes to any container type, even ones that use non-contiguous memory. The specific reference to `std::vector` can be removed, and replaced with a generic template type parameter. That could simplify the allocator `A` type parameter part as well.

I suppose technically you might want to guard the template with the notion of the container being iterable (and the element type being writable by the stream library). There's no built in type_trait check for iterable, but a solution can be found here:
https://stackoverflow.com/questions/13830158/check-if-a-variable-is-iterable

DanRStevens:
I came across `void_t`, which uses an example that checks for iterable:
https://en.cppreference.com/w/cpp/types/void_t

Another point worth mentioning, which is touched on in that, is using `cbegin` and `cend`, rather than `begin` and `end`. The "c" prefix is for `const`, and produces a const iterator, which can not modify the object being iterated over. Reference:
https://en.cppreference.com/w/cpp/iterator/begin

----

I came across `std::data`:
https://en.cppreference.com/w/cpp/iterator/data

It allows for consistent treatment of containers with a `.data()` method, as well as arrays, and initializer lists. In terms of built in containers, and container like objects with `.data()`, that includes `std::array`, `std::vector`, and `std::string`. In particular, the method to write strings and vectors can be collapsed, as the method body is effectively the same.

[...]

PR #288, opened 2019-03-25
Cleanup stream read method guards

DanRStevens:
This is the read update corresponding to PR #287. Tagging Issue #278 as related. Tagging PR #281 as related for simplifying the template parameters.

There was one slight difference to the write case. I didn't collapse one of the string read methods into the generalized container form, since the `string.data()` method didn't have a non-const overload until C++17. Reference:
https://en.cppreference.com/w/cpp/string/basic_string/data

The current Linux flags are set for C++14. I'm uncertain if there are blocking issues with the Windows project, particularly regarding the Google Test unit test project. If it's fine to update to C++17, then we can collapse those two methods into one.

### DIFF
--- a/src/Stream/Reader.h
+++ b/src/Stream/Reader.h
@@ -82,25 +82,5 @@ namespace Stream
 			container.resize(containerSize);
 			Read(container);
 		}
-
-		// Size prefixed string data types
-		// Ex: Read<uint32_t>(string); // Read 32-bit string length, allocate space, then read string data
-		template<typename SizeType, typename CharT, typename Traits, typename Allocator>
-		void Read(std::basic_string<CharT, Traits, Allocator>& string) {
-			SizeType stringSize;
-			Read(stringSize);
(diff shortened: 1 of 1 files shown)


Brett208:
If PR #290 is merged, this branch can use C++17 features for MSVC builds.

DanRStevens:
I just took another look at this. It should be fine to merge now.

Once we enable C++17 on all systems we could then remove the one largely redundant method. If we want to do the method removal as part of this PR, then we should wait until the other changes are merged first.

Brett208:
Unless we plan to update to requiring C++17 now on all Linux compilers, I would prefer merging and writing up an issue for later work. Better to close this branch out instead of letting it grow stale over time.

We could create a C++17 tag for issues as a reminder on which ones are waiting on C++17 as standard to complete.

DanRStevens:
Ok, very good suggestion. I wrote up an issue. I'll merge this now.

PR #290, opened 2019-04-17
Update nuget gtest

Brett208:
Should be able to use all C++17 features on MSVC builds now. Also removed suppression of some warnings that gtest was producing.

It looks like gmock is still not installed. Initially I thought it was, but turns out it was pulling it from my vcpkg install of gtest/gmock.

-Brett


Brett208:
I like __has_include and __cpp_lib_filesystem. Probably best if I'm not the one testing since I don't have ready access to Linux.

Whenever you think we should go prime with C++17 on Linux builds, just let me know. I guess until then I'll try to stay away except for filesystem.

Brett

PR #291, opened 2019-04-18
Filter out non files resource manager

Brett208:


test/ResourceManager.test.cpp
```diff
@@ -17,3 +18,7 @@ TEST(ResourceManager, IgnoreNonFilenamesOnConstruction) {
 	// Attempt to load Resource Manager where directories exist with .vol and .clm 'extensions'
 	EXPECT_NO_THROW(ResourceManager resourceManager("./data"));
 }
+
+TEST(ResourceManager, AbortsWithCreationOnFilename) {
+	EXPECT_THROW(ResourceManager("./data/Directory.clm/.keep"), std::runtime_error);
```
DanRStevens: I was going to say this test name is a little confusing, though I see it was changed in a later commit. I really had to think about this one. I assume it's throwing during construction because you have to pass a folder name, but `.keep` is not a folder name. This is perhaps made more confusing because of the special semantics around the name `.keep`, which is being conveniently re-used for another purpose here.

test/ResourceManager.test.cpp
```diff
@@ -3,22 +3,20 @@
...
+TEST(ResourceManager, ConstructResourceManager) {
 	ResourceManager resourceManager("");
 
+	// Refuse absolute paths
 #ifdef _WIN32
 	EXPECT_THROW(resourceManager.GetResourceStream("C:/Archive.vol"), std::runtime_error);
 	EXPECT_THROW(resourceManager.GetResourceStream("C:/PathTo/Archive.vol"), std::runtime_error);
 #endif
 	EXPECT_THROW(resourceManager.GetResourceStream("/Archive.vol"), std::runtime_error);
 	EXPECT_THROW(resourceManager.GetResourceStream("/PathTo/Archive.vol"), std::runtime_error);
```
DanRStevens: These tests don't seem related to `ConstructResourceManager`. It's testing the `GetResourceStream` method on an already constructed `ResourceManager` object. They should probably be kept separate from the other tests, with the old test name restored for this portion (and possibly also mentioning the method name somewhere).

test/ResourceManager.test.cpp
```diff
@@ -3,22 +3,20 @@
...
 
-TEST(ResourceManager, IgnoreNonFilenamesOnConstruction) {
-	// Attempt to load Resource Manager where directories exist with .vol and .clm 'extensions'
+	// Load Resource Manager where directories exist with .vol and .clm 'extensions'
 	EXPECT_NO_THROW(ResourceManager("./data"));
-}
 
-TEST(ResourceManager, AbortsWithCreationOnFilename) {
+	// Refuse to load when passed a filename
 	EXPECT_THROW(ResourceManager("./data/Directory.clm/.keep"), std::runtime_error);
```
DanRStevens: I'm maybe a little weary of re-using `.keep` for another purpose here. It's semantically confusing.

----

I'm thinking these tests are maybe a little too different to merge into one. One is about a file or folder name, the other is about the contents of a folder and what happens with later processing.

test/ResourceManager.test.cpp
Brett208: Oh yeah. I misread what these tests were for. You are right they should be separated.

test/ResourceManager.test.cpp
```diff
@@ -2,10 +2,9 @@
 #include <gtest/gtest.h>
 
 
-TEST(ResourceManager, GetResourceStream) {
+TEST(ResourceManager, RefusesAbsolutePaths) {
```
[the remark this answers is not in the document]
Brett208: I like having ResourceManager on the left since it groups the ResourceManager tests together. Adding the function name starts to bin the tests into too many small categories. For this function, I like having the function name on the right and the description of what is being tested. Not sure beyond that yet though.

test/ResourceManager.test.cpp
Brett208: Added an empty.txt file to use instead of .keep to make easier to track. Good point.

We have a couple of empty.txt files now in the different test data directories. Maybe worth looking at consolidating? Maybe nice to keep things separate too though.

PR #292, opened 2019-04-18
Prevent resource manager from returning directories

Brett208:
This prevents returning directories from the resource manager. I'm ok with the results. 

We had discussed returning directories from the ResourceManager. I have to check if directory contents are filenames or not before returning results from ResourceManager. Otherwise it would invalidate all the filenames returned from the archives.

I we want that functionality currently, perhaps two sets of functions:
 - ResourceManager::GetFilenames
 - ResourceManager::GetSubdirectories

I don't think we can use something like GetDirectoryContents because the user wouldn't know if the returned items were directories or files contained in an archive. I guess they could run IsDirectory and IsFilename and anything remaining has to be from an archive, but that seems strange.

I'm fine with this code as is unless there is a desire to solve the issue at this juncture.

Thanks,
-Brett

### DIFF
--- a/test/ResourceManager.test.cpp
+++ b/test/ResourceManager.test.cpp
@@ -30,7 +30,5 @@ TEST(ResourceManager, GetFilenames)
 
 	EXPECT_EQ(0, filenames.size());
 
-	filenames = resourceManager.GetAllFilenames("Directory.vol");
-
-	EXPECT_EQ(0, filenames.size());
+	EXPECT_EQ(0, resourceManager.GetAllFilenames("Directory.vol").size());
 }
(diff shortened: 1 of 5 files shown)


Brett208:
Fixed merge conflict with master

DanRStevens:
This looks like code I've reviewed elsewhere. If so, can you please tag the base branch's PR?

Brett208:
@DanRStevens, my first time rebasing a branch onto current master manually. Hopefully it all worked out.

-Brett

test/ResourceManager.test.cpp
```diff
@@ -21,3 +21,16 @@ TEST(ResourceManager, GetResourceStream_RefuseAbsolutePath) {
...
+{
+	// Ensure Directory.vol is not returned
+	ResourceManager resourceManager("./data");
+	auto filenames = resourceManager.GetAllFilenamesOfType(".vol");
+
+	EXPECT_EQ(0, filenames.size());
+
+	filenames = resourceManager.GetAllFilenames("Directory.vol");
+
+	EXPECT_EQ(0, filenames.size());
```
DanRStevens: This test can probably remove the temporaries. Ex:
```cpp
EXPECT_EQ(0, resourceManager.GetAllFilenames("Directory.vol").size());
```

Brett208: I don't have a real preference either way. Will make the change since you brought it up though.

Thanks for the quick review after I straightened out the branch rebase. Had to use the command line for that instead of tortoiseGit.

PR #297, opened 2019-08-03
Write files in new directories

Brett208:
I think there is an issue with XFile::GetDirectory

It returns the filename without the extension in some cases instead of the directory.

### DIFF
--- a/src/Stream/FileWriter.cpp
+++ b/src/Stream/FileWriter.cpp
@@ -40,7 +40,7 @@ namespace Stream
 	{
 		// Create directory if it does not exist. ofstream will fail if directory does not exist.
 		auto directory = XFile::GetDirectory(filename);
-		if (!XFile::PathExists(directory)) {
+		if (directory != "" && !XFile::PathExists(directory)) {
 			XFile::NewDirectory(directory);
 		}
 
--- a/src/Stream/FileWriter.cpp
+++ b/src/Stream/FileWriter.cpp
@@ -40,7 +40,7 @@ namespace Stream
 	{
 		// Create directory if it does not exist. ofstream will fail if directory does not exist.
 		auto directory = XFile::GetDirectory(filename);
-		if (directory != "" && !XFile::PathExists(directory)) {
+		if (!directory.empty() && !XFile::PathExists(directory)) {
 			XFile::NewDirectory(directory);
 		}
 
--- a/src/Stream/FileWriter.cpp
+++ b/src/Stream/FileWriter.cpp
@@ -42,7 +42,7 @@ namespace Stream
 		auto directory = XFile::GetDirectory(filename);
 		if (!directory.empty() && !XFile::PathExists(directory)) 
 		{
-			if (openMode != OpenMode::CanOpenNew) {
+			if (openMode & OpenMode::CanOpenNew) {
 				throw std::runtime_error("The directory does not exist. The write command was restricted from creating new files or directories.");
 			}
 
--- a/src/Stream/FileWriter.cpp
+++ b/src/Stream/FileWriter.cpp
@@ -1,4 +1,5 @@
 #include "FileWriter.h"
+#include "../XFile.h"
 #include <stdexcept>
 #include <experimental/filesystem>
 namespace fs = std::experimental::filesystem;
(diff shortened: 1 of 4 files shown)


test/Stream/FileWriter.test.cpp
```diff
@@ -65,3 +65,13 @@ TEST(FileWriter, MoveConstructible) {
...
 }
+
+TEST(FileWriter, DirectoryDoesNotExist) {
+	EXPECT_NO_THROW(Stream::FileWriter writer("../NewDirectory/TestFile.temp"));
+	EXPECT_NO_THROW(Stream::FileWriter writer("./NewDirectory/TestFile.temp"));
+	EXPECT_NO_THROW(Stream::FileWriter writer("NewDirectory/TestFile.temp"));
+
+	EXPECT_NO_THROW(Stream::FileWriter writer("..\\NewDirectory\\TestFile.temp"));
+	EXPECT_NO_THROW(Stream::FileWriter writer(".\\NewDirectory\\TestFile.temp"));
+	EXPECT_NO_THROW(Stream::FileWriter writer("NewDirectory\\TestFile.temp"));
```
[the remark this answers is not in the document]
DanRStevens: Yes, on Linux, the Windows path separator leaves a file literally named "NewDirectory\TestFile.temp". The "\" is treated just like any other character as part of the filename. As such the tests pass, but aren't really doing what they were expected to do.

DanRStevens:
The new tests should probably clean up the temporary files they are generating.

src/Stream/FileWriter.cpp
```diff
@@ -40,7 +40,7 @@ namespace Stream
 	{
 		// Create directory if it does not exist. ofstream will fail if directory does not exist.
 		auto directory = XFile::GetDirectory(filename);
-		if (!XFile::PathExists(directory)) {
+		if (directory != "" && !XFile::PathExists(directory)) {
```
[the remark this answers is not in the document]
Brett208: I also like that string::empty reads easier than doing a not equal comparison.

test/Stream/FileWriter.test.cpp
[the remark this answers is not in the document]
Brett208: I wanted to be thorough by testing both cases. However, we are just testing that a file is created in the new directory, so testing using both directory separators on Windows seems redundant.

Brett208:
@DanRStevens would you be willing to help resolve the package install issues going on with the travis-ci tests?

Thanks,
Brett

src/Stream/FileWriter.cpp
```diff
@@ -40,7 +40,12 @@ namespace Stream
 	{
 		// Create directory if it does not exist. ofstream will fail if directory does not exist.
 		auto directory = XFile::GetDirectory(filename);
-		if (!directory.empty() && !XFile::PathExists(directory)) {
+		if (!directory.empty() && !XFile::PathExists(directory)) 
+		{
+			if (openMode != OpenMode::CanOpenNew) {
```
DanRStevens: This should test for the specific flag with `&`, rather than the combination of all flags with exact equality/inequality.
```
# Short form
if (openMode & OpenMode::CanOpenNew) {
```
```
# Explicit form
if ((openMode & OpenMode::CanOpenNew) != 0) {
```
```
# Overly explicit form
if ((openMode & OpenMode::CanOpenNew) == OpenMode::CanOpenNew) {
```

I think I would prefer the first form.

DanRStevens: I think I might also suggest testing the `OpenMode` flag before checking for a non-empty directory component to the path, or checking the filesystem itself.

DanRStevens:
I'm checking into the TravisCI stuff.

src/Stream/FileWriter.cpp
Brett208: Thanks for catching my error on the flag check. Should be fixed with newest push.

If we check the OpenMode flag first, if a new directory is required, the code would bypass our custom error message and attempt to call ofstream. This would produce an error, but not as descriptive. Maybe I'm missing your point though...

-Brett

DanRStevens: Ahh, I didn't notice the custom error message.

DanRStevens:
Should we re-purpose the `CanOpenNew` flag with this new meaning, or should we add a new flag to allow automatic directory creation? I don't mind using the `CanOpenNew` flag to guard this behaviour, as it would seem odd to create new directories if creating new files wasn't allowed. Though it seems conceivable that someone might want to allow creating new files, but not allow creating new directories.

I may be overthinking this though.

Brett208:
I'm happy with the existing flag structure although I wouldn't oppose the new addition if you thought it was important.

Brett208:
@DanRStevens are you okay if I merge this branch into master now.

Thanks,
Brett

PR #301, opened 2019-08-10
Allow empty volumes again

### LINKED ISSUE #298: Allow creating empty volume archives
OP2Utility cannot create an empty volume right now. If I recall correctly, we had designed this feature in, so I think it has been broken since. I'd like to test re-add the feature and add unit tests to keep it from breaking unknown again.

Might be good to test CLM files at some point again as well.

Brett208:
Closes #298

I suspect there is a better way to organize the Empty Archive test but this was the best I could do.

When this is merged in, I'm going back to testing OP2Archive.

### DIFF
--- a/src/Archive/VolFile.cpp
+++ b/src/Archive/VolFile.cpp
@@ -283,6 +283,11 @@ namespace Archive
 		// Calculate the zero padded length of the string table and index table
 		volInfo.paddedStringTableLength = (volInfo.stringTableLength + 7) & ~3;
 		volInfo.paddedIndexTableLength = (volInfo.indexTableLength + 3) & ~3;
+
+		if (volInfo.indexEntries.size() == 0) {
+			return;
+		}
+
 		volInfo.indexEntries[0].dataBlockOffset = volInfo.paddedStringTableLength + volInfo.paddedIndexTableLength + 32;
 
 		// Calculate offsets to the files
--- a/test/Archive/ArchiveFile.test.cpp
+++ b/test/Archive/ArchiveFile.test.cpp
@@ -1,0 +1,69 @@
+#include "Archive/VolFile.h"
+#include "Archive/ClmFile.h"
+#include "XFile.h"
+#include <gtest/gtest.h>
+#include <vector>
+#include <string>
+
+void TestEmptyArchive(Archive::ArchiveFile& archiveFile, const std::string& archiveFilename);
+
+
+TEST(VolFile, EmptyArchive) 
+{
+	const std::string archiveFilename("EmptyArchive.vol");
+
+	{
+		// Create an empty archive
+		const std::vector<std::string> filesToPack;
+		Archive::VolFile::CreateArchive(archiveFilename, filesToPack);
+
+		Archive::VolFile archiveFile(archiveFilename);
+
+		SCOPED_TRACE("Empty Volume File");
+		TestEmptyArchive(archiveFile, archiveFilename);
+
+		EXPECT_THROW(archiveFile.GetCompressionCode(0), std::runtime_error);
+	}
+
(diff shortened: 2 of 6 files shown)


test/Archive/ArchiveFile.test.cpp
```diff
@@ -0,0 +1,69 @@
...
+		// Create an empty archive
+		const std::vector<std::string> filesToPack;
+		Archive::VolFile::CreateArchive(archiveFilename, filesToPack);
+
+		Archive::VolFile archiveFile(archiveFilename);
+
+		SCOPED_TRACE("Empty Volume File");
+		TestEmptyArchive(archiveFile, archiveFilename);
+
+		EXPECT_THROW(archiveFile.GetCompressionCode(0), std::runtime_error);
```
DanRStevens: I don't know if this check is really needed. It seems like a roundabout and less clear way of testing `EXPECT_EQ(0, archiveFile.GetCount());`, which is already checked.

test/Archive/ArchiveFile.test.cpp
```diff
@@ -0,0 +1,69 @@
...
+{
+	const std::string extractDirectory("./TestExtract");
+
+	EXPECT_THROW(archiveFile.GetIndex("TestItem"), std::runtime_error);
+	EXPECT_THROW(archiveFile.GetName(0), std::runtime_error);
+	EXPECT_THROW(archiveFile.GetSize(0), std::runtime_error);
+	EXPECT_FALSE(archiveFile.Contains("TestItem"));
+
+	EXPECT_THROW(archiveFile.OpenStream(0), std::runtime_error);
+	EXPECT_THROW(archiveFile.ExtractFile(0, "Test"), std::runtime_error);
```
DanRStevens: The above checks seem a little bit out of scope. Again, this is largely just checking `EXPECT_EQ(0, archiveFile.GetCount());`, which is already checked elsewhere. Granted, tests like these are important, to ensure there are no bounds overflow in relation to `GetCount`, I'm just not certain this is the best place for it. I sort of feel these should be split off to a separate unit test, which is more targeted towards ensuring access past `GetCount` results in an exception. That may include a case where `GetCount` is 0, or other values.

src/Archive/VolFile.cpp
```diff
@@ -283,6 +283,11 @@ namespace Archive
 		// Calculate the zero padded length of the string table and index table
 		volInfo.paddedStringTableLength = (volInfo.stringTableLength + 7) & ~3;
 		volInfo.paddedIndexTableLength = (volInfo.indexTableLength + 3) & ~3;
+
+		if (volInfo.indexEntries.size() == 0) {
+			return;
+		}
```
DanRStevens: No objections here. This is likely how I would have done it. Efficient, and avoids excessive indentation. Just wanted to comment that I needed to more closely examine the end of this routine to make sure the early exit was appropriate. It also occurred to me that if code is later added to the end of the routine, some special consideration may need to be made for the early exit here.

test/Archive/ArchiveFile.test.cpp
Brett208: The goal was to ensure that when you tried to pull a compression code from an empty volume, the proper error was thrown. If you think it is too redundant I suppose I don't mind if it is stricken.

test/Archive/ArchiveFile.test.cpp
Brett208: Agreed more tests should be made against volume and clm files which may result in pulling some of the tests out of the empty archive test routines. I wasn't planning to write robust tests against everything clm/vol related right now, although it would be a good goal.

Brett208:
Thanks for checking out the code.

-Brett

PR #304, opened 2019-08-16
Zero cost negative signed check

### LINKED ISSUE #275: Ensure zero-cost check conditions in Stream
Currently the Stream code contains checks such as:
```cpp
if (vectorSize < 0) {
  throw std::runtime_error("Vector's size may not be a negative number");
}
```
This is only possible if the data type of `vectorSize` is a signed number. In many cases, users of the library will be specifying an unsigned type. For unsigned types, we should try to ensure zero runtime cost for this check.

One possibility, with C++17, is to add a constexpr-if guard around the check:
```cpp
if constexpr(std::is_signed<SizeType>::value) {
  // ...
}
```

Currently the project is set to use C++14. Though, the use of the Filesystem API in XFile is technically already C++17. If we are to use constexpr-if, we should change the project settings to use C++17 to avoid warnings over use of constexpr-if.

Caveat: It's unknown if adding these constexpr-if guards is needed to ensure zero runtime cost. The compiler may already be smart enough to realize the existing checks are always false for certain data types. This would be compiler specific though, so we may need to check the output from multiple compilers.

----

One way to test this issue, is to check the assembly output for a small test file:
```cpp
#include "src/Stream/Reader.h"
#include <vector>


void f(Stream::Reader& r, std::vector<int>& v) {
  r.Read<unsigned int>(v);
}
```
This file should avoid generating uninteresting code, such as construction of either a Reader or a vector, and instead focus on the call to the Read method, and what that template method expands to. Without optimization though, it still generates quite a bit of code.

Optimized:
```
clang++ Test.cpp -S -O2
```
Unoptimized:
```
clang++ Test.cpp -S
```

DanRStevens:
This closes #275.

### DIFF
--- a/src/Stream/Reader.h
+++ b/src/Stream/Reader.h
@@ -2,6 +2,7 @@
 
 #include <cstddef>
 #include <type_traits>
+#include <limits>
 #include <vector>
 #include <string>
 #include <stdexcept>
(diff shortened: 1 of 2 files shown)


DanRStevens:
Very interesting effects on the debugger.

PR #305, opened 2019-08-17
Update ReadMe.md to add feature details

DanRStevens:
Also updated supported compiler list.

Feel free to adjust. I just thought I'd get a start on it.

### DIFF
--- a/ReadMe.md
+++ b/ReadMe.md
@@ -1,10 +1,10 @@
 # OP2Utility
 
-A C++17 utility library for working with Outpost 2 game resources (maps, graphics, sounds). This is a static library to be used by other code projects.
+A cross platform C++17 utility library that simplifies Outpost 2 game resource manipulation, including archives (vol and clm), maps, graphics, sounds, and music. This is a static library designed for use by other code projects.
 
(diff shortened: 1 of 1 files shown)


DanRStevens:
Hmm, thinking about those references to other libraries, you're right that we could probably come up with something better. I think the first paragraph should focus on what the library is, rather than what it is not. I seem to have set a negative description with the two references I added.

Perhaps in the following paragraphs we could add the details about related repositories, or about other repositories the user may have been looking for.

This also raises the question, how clear is the repository name? Could we do better? I can't think of anything better. Though if we're feeling the need to explain what something is and what it is not, perhaps the repository name is the real issue.

DanRStevens:
I wonder if "OP2Resources" might be a better name for this repository. That might also help set the scope of the library.

----

I'm fine with removing the change log, if you want to do that. We really haven't been tracking changes well for this project, so the only real way to get changes is through the project commit history.

Brett208:
I tried to condense some of your thoughts in a final edit. Hopefully it reads better? I think we should probably merge this soon instead of continuing to edit.

I would be okay switching the name to OP2Resources. It is a bit more specific than OP2Utility. However, that would be a lot of work renaming files, updating project files, and repo name and we would lose the momentum in the forums of the old name. I'm not looking to take that on right now myself (lazy I suppose).

DanRStevens:
Definitely reads better now.

The repository name can be addressed another time.

PR #306, opened 2019-08-18
Remove extra use of public keyword from map.h

Brett208:


Brett208:
I would support better organization of the public members.

PR #310, opened 2019-10-27
Update to latest stable version of Google Mock

Brett208:


test/OP2UtilityTest.vcxproj.filters
```diff
@@ -69,6 +69,8 @@
     <ClCompile Include="Archive\ArchiveFile.test.cpp">
       <Filter>Archive</Filter>
     </ClCompile>
+    <ClCompile Include="$(MSBuildThisFileDirectory)..\..\lib\native\src\gtest\gtest-all.cc" />
+    <ClCompile Include="$(MSBuildThisFileDirectory)..\..\lib\native\src\gmock\gmock-all.cc" />
```
DanRStevens: Uhh, what is this for? Is this really needed? Usually that's handled by one of the static libraries that is normally packaged with Google Test / Google Mock.

Brett208: I used Visual Studio 2019 when updating the nuget package. Not sure why these lines were added this time and not in the past.

DanRStevens: I guess if that's how it was packaged, then go for it. I suppose maybe that's so the library is compiled with the same settings as the project.

I suspect that package setup might add a small bit of time to a fresh compile. Not so bad for a development machine, though it would affect CI builds more as those are always fresh builds. Hard to tell with statistical noise in build times, but AppVeyor builds seem to be taking longer than I remember by about 15 seconds per configuration. Not sure how much comes from this change versus accumulated changes over the last couple of months. Anyway, not a problem. Just thought you might be interested in checking locally.

Brett208: I did notice the lines were added, but sense the IDE did it on its own accord, I thought it would be best to leave it in as standard unless there was a reason to remove/change it. 

I did a quick Google search for why nuget would add a ClCompile line to the filters project, but nothing came up. The filters file is actually optional unless you want filters within your solution navigation view in VS, so I suspect these lines are not necessary. I wouldn't mind the automated build time being quicker.

DanRStevens: Ahh, I hadn't noticed it was the filters file. I thought it was the project file. This likely has no impact on compile time then.

Brett208: I guess what I meant was I use the filters file to organize folders inside the IDE. However, I have no idea what the ClCompile command inside the filters file is for as it doesn't seem to make sense from a filter perspective. Although, sense the IDE put it there, I'd leave it until we had a reason to pull it I suppose.

DanRStevens: You're right, it is a bit odd. I think that's why I was confused and thought it was the project file. I thought it was related to actually compiling something.

Looking at the filters file, it uses `ClCompile` for .cpp files and `ClInclude` for .h files. It's the same in both the project file and the filters file.

PR #313, opened 2019-11-05
Add Read Null Terminated String

### LINKED ISSUE #311: Allow reading a null terminated string from stream when length is unknown
Sometimes strings are stored without providing the length. Sometimes stored strings may either be null terminated or just use the maximum allowed indices without null terminated. 

An example would be the SectionTable::Name field from the PE format.

> An 8-byte, null-padded UTF-8 encoded string. If the string is exactly 8 characters long, there is no terminating null. 

I propose adding something similar to below code to stream::reader:

```C++
// Reads a string until a null terminator is encountered or maxCount is reached
std::string ReadNullTerminatedString(std::size_t maxCount = SIZE_MAX)
{
	std::string str;

	char c;
	for (std::size_t i = 0; i < maxCount; ++i)
	{ 
		stream->Read(c);
		if (c == '\0') {
			break;
		}

		str.push_back(c);
	}

	return str;
}
```

Brett208:
Closes #311 

I think tests are fairly robust, although they are not easy to read

### DIFF
--- a/src/Stream/Reader.h
+++ b/src/Stream/Reader.h
@@ -85,5 +85,26 @@ namespace Stream
 			container.resize(containerSize);
 			Read(container);
 		}
+
+		// Read characters into a string until a null terminator is encountered
+		// Does not include the null terminator in the returned string
+		// Use maxCount to restrict the size of the returned string
+		std::string ReadNullTerminatedString(std::size_t maxCount = SIZE_MAX)
+		{
+			std::string str;
+
+			char c;
+			for (std::size_t i = 0; i < maxCount; ++i)
+			{
+				Read(c);
+				if (c == '\0') {
+					break;
+				}
+
+				str.push_back(c);
+			}
+
+			return str;
+		}
 	};
 }
--- a/test/Stream/MemoryStreamReader.test.cpp
+++ b/test/Stream/MemoryStreamReader.test.cpp
@@ -98,7 +98,7 @@ TEST_F(SimpleMemoryReader, SeekRelativeOutOfBoundsEndPreservesPosition) {
 	EXPECT_EQ(position, stream.Position());
 }
 
-TEST(MemoryReader, ReadNullTerminatedString)
+TEST(MemoryReader, ReadNullTerminatedStringUnbounded)
 {
 	constexpr std::array<char, 5> terminatedBuffer{ 'n', 'u', 'l', 'l', '\0' };
 	Stream::MemoryReader reader(&terminatedBuffer[0], terminatedBuffer.size());
@@ -107,9 +107,13 @@ TEST(MemoryReader, ReadNullTerminatedString)
 	EXPECT_EQ("null", reader.ReadNullTerminatedString());
 	// Stream is adanced by string length + null terminator
 	EXPECT_EQ(5u, reader.Position());
+}
 
-	// Check maxCount stops reading string at proper location
-	reader.Seek(0);
+TEST(MemoryReader, ReadNullTerminatedStringBounded)
+{
+	constexpr std::array<char, 5> terminatedBuffer{ 'n', 'u', 'l', 'l', '\0' };
+	Stream::MemoryReader reader(&terminatedBuffer[0], terminatedBuffer.size());
+
 	// Test bounded read with maxCount characters
 	EXPECT_EQ("nu", reader.ReadNullTerminatedString(2));
 	// Stream is adanced by read length (no null terminator was seen)
--- a/test/Stream/MemoryStreamReader.test.cpp
+++ b/test/Stream/MemoryStreamReader.test.cpp
@@ -2,6 +2,8 @@
 #include "BidirectionalReader.test.h"
 #include "Stream/MemoryReader.h"
 #include <array>
+#include <string>
+#include <stdexcept>
 
 
 template <>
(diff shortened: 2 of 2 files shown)


src/Stream/Reader.h
```diff
@@ -85,5 +85,26 @@ namespace Stream
 			container.resize(containerSize);
 			Read(container);
 		}
+
+		// Read characters into a string until a null terminator is encountered
+		// Does not include the null terminator in the returned string
+		// Use maxCount to restrict the size of the returned string
+		std::string ReadNullTerminatedString(std::size_t maxCount = SIZE_MAX)
```
Brett208: I think maxCount is a good idea here, but could discuss removing as well.

test/Stream/MemoryStreamReader.test.cpp
```diff
@@ -95,3 +97,36 @@ TEST_F(SimpleMemoryReader, SeekRelativeOutOfBoundsEndPreservesPosition) {
 	EXPECT_THROW(stream.SeekForward(6), std::runtime_error);
 	EXPECT_EQ(position, stream.Position());
 }
+
+TEST(MemoryReader, ReadNullTerminatedString)
+{
+	std::array<char, 5> buffer{ 'n', 'u', 'l', 'l', '\0' };
```
DanRStevens: This could probably be declared as `const`.

test/Stream/MemoryStreamReader.test.cpp
```diff
@@ -95,3 +97,36 @@ TEST_F(SimpleMemoryReader, SeekRelativeOutOfBoundsEndPreservesPosition) {
...
+	std::array<char, 5> result;
+
+	// Check null-terminated
+	Stream::MemoryReader reader(&buffer[0], 5);
```
DanRStevens: Using `buffer.size()` might be more self documenting that an integer literal.

test/Stream/MemoryStreamReader.test.cpp
```diff
@@ -95,3 +97,36 @@ TEST_F(SimpleMemoryReader, SeekRelativeOutOfBoundsEndPreservesPosition) {
...
+	std::string str;
+	
+	EXPECT_NO_THROW(str = reader.ReadNullTerminatedString());
+
+	// Copy to array
+	std::copy(str.begin(), str.end(), result.data());
+	
+	// Ensure null-terminator was not copied
+	EXPECT_NE(buffer, result);
```
DanRStevens: An `std::string` can contain embedded nulls. As such, you can actually just use a string compare to ensure it stopped before reading the null.

```cpp
assert(std::string("A", 1) != std::string("A\0", 2));
```

I believe the tests can be written without needing to use `result`.

test/Stream/MemoryStreamReader.test.cpp
```diff
@@ -95,3 +97,36 @@ TEST_F(SimpleMemoryReader, SeekRelativeOutOfBoundsEndPreservesPosition) {
...
+
+	// Add null-terminator to check for equality
+	result[4] = '\0';
+	EXPECT_EQ(buffer, result);
+
+	// Check string without null-termination eventually throws
+	buffer[4] = '0';
+	reader.Seek(0);
+	EXPECT_THROW(str = reader.ReadNullTerminatedString(), std::runtime_error);
+	EXPECT_NE(buffer, result);
```
DanRStevens: Maybe a comment about where the exception comes from. It wasn't obvious to me at first that this relied on the stream to throw an exception, rather than just reading off into random unbounded memory locations until something (maybe) went wrong.

test/Stream/MemoryStreamReader.test.cpp
```diff
@@ -95,3 +97,36 @@ TEST_F(SimpleMemoryReader, SeekRelativeOutOfBoundsEndPreservesPosition) {
...
+
+	// Check maxCount stops reading string
+	reader.Seek(0);
+	EXPECT_NO_THROW(str = reader.ReadNullTerminatedString(4));
+	EXPECT_EQ(4, str.size());
```
DanRStevens: If the string value is tested directly, that also indirectly tests the string length. The test might be a bit more clear about what is happening by testing the string value.

src/Stream/Reader.h
DanRStevens: Heh, I did actually stop to think about that for a bit. It makes it seem almost too much like the fixed length field reading code. Though it's not. The stream position here is only advanced up to and consuming the null byte, not up to the maximum length of the read.

I think the parameter is fine, particularly given the default selected.

test/Stream/MemoryStreamReader.test.cpp
DanRStevens: This check makes an assumption about the last character of the `result` buffer being non-zero, however, the last character is actually left uninitialized, and so could be zero. That's likely the cause of the test failures.

Testing using the `std::string` result directly, rather than copying to a `std::array` may sidestep this problem.

src/Stream/Reader.h
Brett208: Thanks for looking it over.

Brett208:
@DanRStevens I think this addresses the points you brought up??? It is definitely better written now.

Looks like I accidentally committed readme changes into this branch...

DanRStevens:
I went and split out the ReadMe.md updates into PR #316.

I did a bit of work on the unit tests to hopefully make them a little easier to read, and to test a bit more in terms of stream position expectations.

----

After looking at the unit tests more closely, I wonder if we might change the specification for when the end of the stream has been reached. Maybe we should return the string read so far, rather than throwing an exception? This is of course a bit of an arbitrary choice. One interesting case is it might make sense to throw an exception if you were already at the end of the stream, rather than silently return a string of length 0. Or maybe a string of length 0 is entirely desirable.

If we make a change to not throw, but rather return the string read so far, we could consider using the non-throwing `ReadPartial` method internally.

PR #318, opened 2019-11-09
Improve CLM file creation error messages

Brett208:
thrown errors inside ReadAllWaveHeaders did not show the filename

Noticed during testing of OP2Archive.


Brett208:
Well, it is also partially my fault I think.

-Brett

PR #322, opened 2019-12-14
Use Ranged Based Logic in GetFilenamesFromDirectory

### LINKED ISSUE #320: Redundant check in inner loop
Within `XFile::GetFilenamesFromDirectory` there is a check against `filenames.size() == 0` which appears to be redundant. The loop conditions already skip the loop when `size() == 0` (initially), so this check within the loop does nothing. It may have made sense if the size was updated to be 0, however, the size can never be updated to 0 until the loop is finished anyway. Finally, other similar methods (regex filter) don't have a corresponding check.

The method in question is:
```cpp
std::vector<std::string> XFile::GetFilenamesFromDirectory(const std::string& directory, const std::string& extension)
{
	std::vector<std::string> filenames = GetFilenamesFromDirectory(directory);

	// Loop starts at index size - 1 and ends after index 0 executes
	for (std::size_t i = filenames.size(); i-- > 0; )
	{
		if (filenames.size() == 0) {
			return filenames;
		}

		if (fs::path(filenames[i]).extension().string() != extension) {
			filenames.erase(filenames.begin() + i);
		}
	}

	return filenames;
}
```

Brett208:
Closes #320 

I think this is easier to understand and reason with than using a decrementing loop.

### DIFF
--- a/src/XFile.cpp
+++ b/src/XFile.cpp
@@ -107,34 +107,24 @@ std::vector<std::string> XFile::GetFilenamesFromDirectory(const std::string& dir
 
 std::vector<std::string> XFile::GetFilenamesFromDirectory(const std::string& directory, const std::regex& filenameRegex)
 {
-	std::vector<std::string> filenames = GetFilenamesFromDirectory(directory);
+	auto filenames = GetFilenamesFromDirectory(directory);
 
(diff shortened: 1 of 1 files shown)


Brett208:
I don't mind if you want to create an internal template function here

PR #325, opened 2019-12-15
Fix `ResourceManager` path bug

### LINKED ISSUE #321: ResourceManager only loads archives in current working directory
The following test demonstrates the problem:
```cpp
#include "../src/Archive/VolFile.h"

TEST(ResourceManager, ArchiveIsLoaded)
{
	Archive::VolFile::CreateArchive("./data/Test.vol", {});

	ResourceManager resourceManager("./data");
	EXPECT_EQ(std::vector<std::string>{"Test.vol"}, resourceManager.GetArchiveFilenames());
}
```

----

In the `ResourceManager` constructor, the .vol files are found with:
```cpp
const auto volFilenames = GetFilesFromDirectory(".vol");
```

The `GetFilesFromDirectory` method is implemented as:
```cpp
std::vector<std::string> ResourceManager::GetFilesFromDirectory(const std::string& fileExtension)
{
	auto directoryContents = XFile::GetFilenamesFromDirectory(resourceRootDir, fileExtension);
	XFile::EraseNonFilenames(directoryContents);

	return directoryContents;
}
```

The `XFile::GetFilenamesFromDirectory` method returns only the filename, with no directory component. The `XFile::EraseNonFilenames(directoryContents);` call implicitly uses the current working directory to test the files using `XFile::IsFile(path)`. Presumably this is to remove directories, however, due to the difference in the path portion, it also filters out regular files which fail to exist in the current working directory.

The end result is, archives files will not be seen and loaded when the `ResourceManager` directory is not the current working directory (unless the current working directory contains an identically named file).

### LINKED ISSUE #324: Rename `GetFilenamesFromDirectory`
I was thinking, `GetFilenamesFromDirectory` is kind of long, and not very information dense. What if we just renamed it to something like `Dir`. Much like the DOS command (and a common alias on Linux), `Dir` simply returns the contents of a directory. This would also match up with other languages, such as Ruby, where you can get directory listings with syntax such as `Dir['*.vol']`.

----

If we wish to restrict results to files (or rather not-directories), we might also introduce a `DirFiles`.

Alternatively, we could expose the template method to allow for arbitrary filtering of results. Of course, it may be nice to provide a few convenience helper wrappers so end users don't have to use lambdas if they wish to avoid that. There's also the concern of namespace pollution, in that if the template is moved to the header file, the header would also need to import the `filesystem` namespace (or rather the `fs` alias).

DanRStevens:
This is based on PR #323. ~~Review may be easier if we first merge the parent branch, and then rebase this branch on master.~~ Edit: Rebase complete.

This fixes #321. Provided wrapper methods may not exactly match those suggested in the issue. We may want to decide if more should be added, or if some should be removed.

This fixes #324. As I hadn't really made a definite decision on the final name, I went with the original I had suggested. I'd still be open to using an alternate name. Mostly I did this rename because it flowed naturally with the updates to the other issue above.

### DIFF
--- a/src/ResourceManager.cpp
+++ b/src/ResourceManager.cpp
@@ -160,16 +160,10 @@ std::vector<std::string> ResourceManager::GetArchiveFilenames()
 
 std::vector<std::string> ResourceManager::GetFilesFromDirectory(const std::string& fileExtension)
 {
-	auto directoryContents = XFile::GetFilenamesFromDirectory(resourceRootDir, fileExtension);
-	XFile::EraseNonFilenames(directoryContents);
-
-	return directoryContents;
+	return XFile::DirFilesWithExtension(resourceRootDir, fileExtension);
 }
 
 std::vector<std::string> ResourceManager::GetFilesFromDirectory(const std::regex& filenameRegex)
 {
-	auto directoryContents = XFile::GetFilenamesFromDirectory(resourceRootDir, filenameRegex);
-	XFile::EraseNonFilenames(directoryContents);
-
-	return directoryContents;
+	return XFile::DirFiles(resourceRootDir, filenameRegex);
 }
--- a/src/XFile.cpp
+++ b/src/XFile.cpp
@@ -92,7 +92,7 @@ std::string XFile::AppendToFilename(const std::string& filename, const std::stri
 	return newPath.string();
 }
 
-std::vector<std::string> XFile::GetFilenamesFromDirectory(const std::string& directory)
+std::vector<std::string> XFile::Dir(const std::string& directory)
 {
 	// Creating a path with an empty string will prevent the directory_iterator from finding files in the current relative path.
 	auto pathStr = directory.length() > 0 ? directory : "./";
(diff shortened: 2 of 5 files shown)


Brett208:
I'm trying to test this against OP2Archive.

```C++
auto filenames = XFile::GetFilenamesFromDirectory(tempDirectory);
```

I'm trying to replace the function above, but there doesn't seem to be a replacement function that does not require a regex string? Is this on purpose?


```C++
auto filenames = XFile::DirFiles(tempDirectory); // This doesn't work
```

-Brett

DanRStevens:
You're right, that is a rather useful overload that is absent.

I suppose technically the old code would be equivalent to `Dir(tempDirectory)`. However, I believe you want to specifically look for files only, not directories. Hence it would be more correct to use an alternate replacement method. This very much gets at the heart of the original debate about how we should return and filter results.

I'll add an overload to handle returning all files (but not directories) from a folder, without additional filtering.

Brett208:
Thanks. I'll keep testing once the new overload is pushed.

DanRStevens:
Well that was quick :wink:

PR #326, opened 2019-12-17
When attempting to create a file with the same name as a directory, F…

Brett208:
…ileWriter can present a more specific error.

This one does not have to be merged if it is controversial, but it seemed easier to just demonstrate in code than write an issue.

I was confused for a while because earlier I had created a test.vol directory for testing purposes. The code kept telling me the file already existed even though I was passing the overwrite flag. Took me a while to diagnose.

### DIFF
--- a/src/Stream/FileWriter.cpp
+++ b/src/Stream/FileWriter.cpp
@@ -38,6 +38,10 @@ namespace Stream
 	FileWriter::FileWriter(const std::string& filename, OpenMode openMode) :
 		filename(filename)
 	{
+		if (XFile::IsDirectory(filename) && XFile::PathExists(filename)) {
+			throw std::runtime_error("Requested filename already exists as a directory.");
+		}
+
 		// Create directory if it does not exist. ofstream will fail if directory does not exist.
 		auto directory = XFile::GetDirectory(filename);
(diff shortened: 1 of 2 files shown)


src/Stream/FileWriter.cpp
```diff
@@ -38,6 +38,10 @@ namespace Stream
 	FileWriter::FileWriter(const std::string& filename, OpenMode openMode) :
 		filename(filename)
 	{
+		if (XFile::IsDirectory(filename) && XFile::PathExists(filename)) {
```
DanRStevens: My gut here is that `IsDirectory` should be sufficient. To get `true` from `IsDirectory`, I assume the path either needs to exist, or it was the empty string `""`.

We might want to throw with a custom error message for an empty `filename`. I'm not sure what error message that would give currently.

Brett208: Custom error message for filename is now incorporated. I still need to test IsDirectory as mentioned to see actual behaviour. 

FileWriter testing would probably benefit from using GoogleMock.

Brett208:
Ok, ready to merge now. In the future, I'd like to improve the unit tests so we check the correct throw statement is hit.

PR #330, opened 2019-12-27
Add a GetLavaPossible helper function to Map

Brett208:
I was going to add support for checking if valid x and y coordinates are given, but that creates a pretty significant chain reaction of code updates and maybe should be in a different branch. 

Also, I think we need to document the function Map::GetTileIndex better (for my own sake and anyone else who isn't great at bit twiddling).


Brett208:
Actually, adding bounds checks to GetTileIndex is exactly what I attempted to do. Maybe I should add an issue for this to track.

PR #335, opened 2020-02-15
Use vs2019 toolset

Brett208:
I think this is all that is needed to switch to defaulting to VS2019's associated toolset. Any other settings people are looking to update when using VS2019?


Brett208:
I may try switching warning level to 4. It looks like VS2019 has added some additional warnings on its own though that should be addressed. I fixed one, but there are some more.

PR #337, opened 2020-02-16
Fix move warning within FileWriter

Brett208:
I think the warning is telling me that the move isn't happening anyways because the source is labeled const? Not so sure about this one.

Warning	C26478	Don't use std::move on constant variables. (es.56).	OP2Utility	\OP2UTILITY\SRC\STREAM\FILEWRITER.CPP	67

https://docs.microsoft.com/en-us/visualstudio/code-quality/c26478


Brett208:
I think marking class member variables as const is pretty clear cut in C# which probably influences me.

PR #347, opened 2020-11-22
Update gtest Nuget Packet to 1.8.1.3

Brett208:


DanRStevens (approved):
Huh, I forgot we were even still using NuGet on this project.

PR #349, opened 2020-12-20
Improve source code comments in bitmap header

Brett208:


src/Bitmap/BitmapFile.h
```diff
@@ -13,7 +13,7 @@ namespace Stream {
 	class BidirectionalReader;
 }
 
-// BMP Writer only supporting Indexed Color palettes (1, 2, and 8 bit BMPs). 
+// Bitmap File (BMP) only supports indexed color palettes (1, 2, and 8 bit BMPs).
```
DanRStevens, line 16: Technically the bitmap format supports bit depths of:
1, 2, 4, 8, 16, 24 and 32

Source: https://en.wikipedia.org/wiki/BMP_file_format

The original wording may be a bit more accurate in that the writer code in this file only supports a restricted set of bit depths.

DanRStevens (approved):

PR #350, opened 2020-12-20
Specify Image Height when Writing Pixels

Brett208:


DanRStevens (approved):
I like how the change makes the API seem a bit more regular, in that both `width` and `height` are specified together.

Long term, we might want to have a preferred internal format for normal versus upside-down bitmaps, with conversion done during loading or construction of the class. The writing code can then make an assumption about the format being written.

PR #351, opened 2020-12-20
Improve bmp loader palette logic

Brett208:


src/Sprite/ImageMeta.h
```diff
@@ -29,6 +29,11 @@ struct ImageMeta {
 	uint32_t width; // Width of image in pixels
 	ImageType type;
 	uint16_t paletteIndex;
+
+	bool IsShadow() const
+	{
+		return type.bShadow != 0;
+	}
```
DanRStevens, line 36: Considering all fields in the `struct` are public, maybe we should name the field `isShadow` and access it directly.
```cpp
if (imageMeta.type.isShadow) { ... }
```

src/Sprite/OP2BmpLoader.cpp
```diff
@@ -32,6 +31,14 @@ void OP2BmpLoader::ExtractImage(std::size_t index, const std::string& filenameOu
 	BitmapFile::WriteIndexed(filenameOut, imageMeta.GetBitCount(), imageMeta.width, -static_cast<int32_t>(imageMeta.height), palette, pixelContainer);
 }
 
+std::vector<Color> OP2BmpLoader::GetPalette(const ImageMeta& imageMeta)
+{
+	std::vector<Color> palette(artFile.palettes[imageMeta.paletteIndex].size());
+	std::copy(artFile.palettes[imageMeta.paletteIndex].begin(), artFile.palettes[imageMeta.paletteIndex].end(), palette.begin());
+
+	return palette;
+}
```
DanRStevens: Could this be simplified by using a by-value copy of `std::vector`? Do we want to make copies, or should we use references?

src/Sprite/OP2BmpLoader.cpp
```diff
@@ -33,7 +35,18 @@ void OP2BmpLoader::ExtractImage(std::size_t index, const std::string& filenameOu
...
+
+	if (imageMeta.IsShadow())
+	{
+		// Shadow graphic uses a 2 color palette
+		palette.resize(2);
+	}
+	else
+	{
+		palette.resize(artFile.palettes[imageMeta.paletteIndex].size());
+	}
```
DanRStevens, line 48: Err, this looks like it's resizing the original palette in a "Get" method. This is rather unexpected.

DanRStevens (commented):
The palette resizing in a get method looked a bit surprising. The rest seems fine to me.

src/Sprite/OP2BmpLoader.cpp
Brett208: palette is defined in this function before being passed out. I don't think it is resizing the original palette from artFile, just setting the proper size for the new image's palette unless I missed something.

Maybe the function name could be changed to copyImagePaletteFromArtFile.

src/Sprite/ImageMeta.h
Brett208: I was also using a bool type for the helper function so it was easier to interpret pass / fail. The rename to isShadow sounds like a good idea either way though.

src/Sprite/OP2BmpLoader.cpp
Brett208: I don't have too much of an opinion this time besides keeping it simple. We could assume that ArtFile should not be changing content in this situation and pass a reference if desired.

src/Sprite/ImageMeta.h
DanRStevens: Hmm, I hadn't considered the type. I suppose it's not possible to adjust the type without interfering with the bit field.

src/Sprite/OP2BmpLoader.cpp
DanRStevens: Ahh, right, I read that wrong. Was probably too tired when I reviewed it.

This has got me curious though. Does the game really use just the first 2 colors from a full sized palette? Is this this an implementation detail for this particular project.

Brett208: The change is based on @TechCor8's port to C# and reimplementing those changes on OP2Helper. I haven't personally tested yet.

DanRStevens (approved):

PR #352, opened 2020-12-24
Allow initializing OP2BmpLoader using a stream

Brett208:
This allows loading ArtFile without first extracting it from the vol file.

I struggled to right a meaningful unit test if we don't want to keep a copy of an actual ArtFile in memory or within the repository.

### DIFF
--- a/src/Sprite/OP2BmpLoader.cpp
+++ b/src/Sprite/OP2BmpLoader.cpp
@@ -8,6 +8,9 @@
 OP2BmpLoader::OP2BmpLoader(std::string bmpFilename, std::string artFilename) :
 	bmpReader(bmpFilename), artFile(ArtFile::Read(artFilename)) { }
 
+OP2BmpLoader::OP2BmpLoader(std::string bmpFilename, Stream::BidirectionalReader& artFileStream) :
+	bmpReader(bmpFilename), artFile(ArtFile::Read(artFileStream)) { }
+
 void OP2BmpLoader::ExtractImage(std::size_t index, const std::string& filenameOut) 
 {
 	artFile.VerifyImageIndexInBounds(index);
--- a/src/Sprite/OP2BmpLoader.cpp
+++ b/src/Sprite/OP2BmpLoader.cpp
@@ -58,9 +58,9 @@ std::unique_ptr<Stream::FileSliceReader> OP2BmpLoader::GetPixels(std::size_t sta
 }
 
 std::size_t OP2BmpLoader::FrameCount(std::size_t animationIndex) const {
-	return artFile.animations[animationIndex].frames.size();
+	return artFile->animations[animationIndex].frames.size();
 }
 
 std::size_t OP2BmpLoader::LayerCount(std::size_t animationIndex, std::size_t frameIndex) const {
-	return artFile.animations[animationIndex].frames[frameIndex].layers.size();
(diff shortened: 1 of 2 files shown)


src/Sprite/OP2BmpLoader.cpp
```diff
@@ -8,6 +8,9 @@
 OP2BmpLoader::OP2BmpLoader(std::string bmpFilename, std::string artFilename) :
 	bmpReader(bmpFilename), artFile(ArtFile::Read(artFilename)) { }
 
+OP2BmpLoader::OP2BmpLoader(std::string bmpFilename, Stream::BidirectionalReader& artFileStream) :
```
Brett208, line 11: Wasn't sure about adding a comment to the constructor indicating it was useful to allow initializing OP2BmpLoader without first extracting ArtFile from a volume.

src/Sprite/OP2BmpLoader.cpp
```diff
@@ -8,6 +8,9 @@
...
+	bmpReader(bmpFilename), artFile(ArtFile::Read(artFileStream)) { }
```
DanRStevens, line 12: This signature could be relaxed to take a `Stream::Reader`, of which `Stream::BidirectionalReader` is a subclass of. There appear to be no uses of the subclass behaviour. The parameter is simply passed to `ArtFile::Read`, which takes the base class type.

----

Is there a reason to take an stream for the art file, rather than an `ArtFile` instance itself? Might result in less repeated parsing if the same `ArtFile` is used multiple times. That may be relevant here as the art file is a packed format containing many independent bitmaps.

DanRStevens (approved):
The concept here looks good.

Noted one small implementation improvement, and one design decision that may be more important.

src/Sprite/OP2BmpLoader.cpp
Brett208: I hadn't thought about passing in a reference to ArtFile. I was designing so a consumer could extract/load bitmaps without worrying about ArtFile.

At a minimum lets pass Stream:Reader. Good catch there.

src/Sprite/OP2BmpLoader.cpp
Brett208: overcome by other changes

Brett208:
Looks like the branch vs pr build on appveyor caught an issue that would have become a problem on merge with master. Fixed in new commit.

src/Sprite/OP2BmpLoader.cpp
```diff
@@ -5,17 +5,14 @@
...
+OP2BmpLoader::OP2BmpLoader(std::string bmpFilename, std::shared_ptr<ArtFile> artFile) :
+	bmpReader(bmpFilename), artFile(artFile) { }
 
 void OP2BmpLoader::ExtractImage(std::size_t index, const std::string& filenameOut) 
 {
-	artFile.VerifyImageIndexInBounds(index);
+	artFile->VerifyImageIndexInBounds(index);
 
-	ImageMeta& imageMeta = artFile.imageMetas[index];
+	ImageMeta& imageMeta = artFile->imageMetas[index];
```
DanRStevens, line 15: Ahh, I guess storing references to existing data is why you used a `std::shared_ptr`.

DanRStevens (approved):

PR #354, opened 2020-12-26
Palette type update

Brett208:
I haven't tested this code yet.

### DIFF
--- a/src/Sprite/ArtFile.h
+++ b/src/Sprite/ArtFile.h
@@ -14,9 +14,12 @@ namespace Stream {
 	class Writer;
 }
 
+// ArtFile stores sprites in 1 and 8 bit indexed bitmaps. 2 bit bitmaps represent shadows.
+// ArtFile may be packaged within a vol file and still register with Outpost 2.
 struct ArtFile
 {
 public:
+	// Palettes within ArtFile are stored in 8 bits of memory regardless of actual palette size (1 or 8 bit)
 	std::vector<Palette8Bit> palettes;
 	std::vector<ImageMeta> imageMetas;
 	std::vector<Animation> animations;
--- a/src/Sprite/ArtFile.h
+++ b/src/Sprite/ArtFile.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "PaletteHeader.h"
 #include "ImageMeta.h"
 #include "Animation.h"
 #include "../Bitmap/Color.h"
--- a/src/Sprite/ArtFile.h
+++ b/src/Sprite/ArtFile.h
@@ -15,8 +15,12 @@ namespace Stream {
 	class Writer;
 }
 
-// ArtFile stores sprites in 1 and 8 bit indexed bitmaps. 2 bit bitmaps represent shadows.
-// ArtFile may be packaged within a vol file and still register with Outpost 2.
+// ArtFile stores sprite metadata including color palettes. 
+// Composing an in game sprite requires the metadata from ArtFile and the pixel values from OP2_ART.bmp.
+// OP2_ART.bmp does not conform to standard bitmap file formatting.
+// Artfile uses two color palettes for image storage, 1 and 8 bit indexed bitmaps
+//  - All shadows sprites are repesented by 1 bit bitmaps
+//  - All other sprites are represented by 8 bit bitmaps
 struct ArtFile
 {
 public:
--- a/src/Sprite/ArtFile.h
+++ b/src/Sprite/ArtFile.h
@@ -17,7 +17,7 @@ namespace Stream {
 struct ArtFile
 {
 public:
-	std::vector<Palette> palettes;
+	std::vector<Palette8Bit> palettes;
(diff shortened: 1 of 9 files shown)


DanRStevens (approved):
Looks ok.

I maybe have a slight concern that run-time sized palettes might be more workable than compile-time sized palettes that are baked into the type system. Having the size baked into the type system means you can't adapt at runtime to images of different bit depth. Such decisions must be made up front, and if assumptions didn't match up with actual data, the only real course of action would be to throw a runtime exception. More specifically, which palette type would an indexed image reference? The image class might end up needing to either cast palette pointers, or be made a template and create a whole family of classes. Anyway, not too sure how the idea will work out right now. Might be worth exploring though. Should be educational at the very least.

Brett208:
The compile time decision is coming from the fact that we know ArtFile can only contain 1 or 8 bit color palettes (always stored in 8 bits of space). 

Maybe it would be better to move the types into Outpost 2 specific sprite sections instead of in the generic Color header. 

I think what I'm really trying to do is specify the length of a palette when you say the terms 1 bit, 8 bit, etc. Maybe better is we can delete the arrays and use the following:

```C++
// Separate from Color, maybe a public member of BitmapFile?
namespace PaletteLength {
  constexpr std::size_t OneBit = 2;
  constexpr std::size_t TwoBit = 4;
  constexpr std::size_t FourBit = 16;
  constexpr std::size_t EightBit = 256;
}
```

DanRStevens:
The palette length can be calculated mathematically:
```cpp
const auto paletteLength = 1 << bitCount;
```

src/Sprite/OP2BmpLoader.cpp
```diff
@@ -35,19 +36,17 @@ void OP2BmpLoader::ExtractImage(std::size_t index, const std::string& filenameOu
...
+	if (imageMeta.type.isShadow) {
+		bitCount = 1;
 	}
-	else
-	{
-		// All other sprites use a 256 color (8 bit) palette
-		palette.resize(sizeof(Palette8Bit));
+	else {
+		bitCount = 8;
 	}
```
DanRStevens: You can potentially collapse this down to a single line by using the ternary operator. That would also allow the variable to be marked with `const`:
```cpp
const uint16_t bitCount = imageMeta.type.isShadow ? 1 : 8;
```

DanRStevens (approved):
Changes look good.

src/Sprite/ArtFile.h
```diff
@@ -14,10 +15,13 @@ namespace Stream {
 	class Writer;
 }
 
+// ArtFile stores sprites in 1 and 8 bit indexed bitmaps. 2 bit bitmaps represent shadows.
+// ArtFile may be packaged within a vol file and still register with Outpost 2.
```
DanRStevens, line 19: I think "1 bit bitmaps" (2 color) was intended here.

----

The phrasing "still register" isn't super clear.

src/Sprite/ArtFile.h
```diff
@@ -14,10 +15,13 @@ namespace Stream {
...
 struct ArtFile
 {
 public:
-	std::vector<Palette> palettes;
+	// Palettes within ArtFile are stored in 8 bits of memory regardless of actual palette size (1 or 8 bit)
```
DanRStevens, line 23: Is this comment still accurate? Or are palettes now sized according to use? Or are some palettes used by both 1 bpp and 8 bpp bitmaps?

src/Sprite/PaletteHeader.cpp
```diff
@@ -14,7 +14,7 @@ PaletteHeader PaletteHeader::CreatePaletteHeader()
 	PaletteHeader paletteHeader;
 	paletteHeader.remainingTagCount = 1;
 
-	const uint64_t dataSize = sizeof(Palette);
+	const uint64_t dataSize = sizeof(Palette8Bit);
```
DanRStevens, line 17: I wonder if this value should be passed in as a parameter. Could make it an optional parameter, with the current value as default.

Could also potentially set default field values rather than use a factory method to construct the instances. Though that idea doesn't play well with making the above value a parameter.

Is this really a `struct`? It seems like we're trying to maintain some sort of invariant here.

Probably a discussion for the future. Maybe better not to worry about it in this PR.

DanRStevens (approved):
Code looks good. Maybe one nitpick about one of the comment lines from earlier that I just noticed.

Added some thoughts about future design, though probably better not to address here.

src/Sprite/ArtFile.h
Brett208: Yes, currently the ArtFile class represents the file memory footprint and allocates space for 8bpp palettes for all sprites. I'm not advocating for this model over others besides it seemed easiest to implement at the time.

src/Sprite/ArtFile.h
Brett208: I'll rewrite the high level documentation and then merge.

PR #357, opened 2020-12-27
Rename OP2Bmploader

Brett208:
Prep work for loading tilesets (wells).

I wanted to distinguish that this class doesn't load all custom bitmaps, just the non tileset sprites.


Brett208:
I wanted to remove the OP2 prefix from the class name. Eventually if we add a global namespace called OP2Utility, it would start becoming pretty redundant.

DanRStevens (approved):
Looks good. I think the new name adds clarity.

PR #358, opened 2020-12-27
Add function to invert red and blue values of a color

Brett208:
Prep work for eventually loading tile sets


DanRStevens (approved):
Looks good.

I don't know if `Invert` is the best name for this operation. There is such as thing as "color inversion", which I'm pretty sure would be different from swapping color channels. Can always address later though, when we have a better idea what to call the operation.

Brett208:
I'm not particularly tied to the function name of Invert either.

PR #361, opened 2020-12-30
Provide function to validate a tileset

Brett208:
Function will be applied after a tileset is loaded into memory and if in custom format preconditioned into a standard bitmap

### DIFF
--- a/src/Sprite/TilesetLoader.cpp
+++ b/src/Sprite/TilesetLoader.cpp
@@ -1,6 +1,7 @@
 #include "TilesetLoader.h"
+#include "../Bitmap/BitmapFile.h"
 #include "../Stream/BidirectionalReader.h"
-
+#include <stdexcept>
 
 namespace TilesetLoader
 {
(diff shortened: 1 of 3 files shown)


src/Sprite/TilesetLoader.cpp
```diff
@@ -12,4 +13,22 @@ namespace TilesetLoader
...
+		{
+			throw std::runtime_error("Tileset palette must be an 8 bit indexed bitmap");
+		}
+
+		if (tileset.imageHeader.width != 32)
+		{
+			throw std::runtime_error("Tileset width must be exactly 32 pixels");
+		}
+
+		if (std::abs(tileset.imageHeader.height) % 32)
```
DanRStevens, line 29: Technically the `std::abs` wouldn't do anything here. Equivalence classes work in both positive and negative directions. Either way, what we want is congruence to 0.

Not that it matters, though as a stylistic choice, should we compare `== 0` here? The `if` will convert the expression to boolean, that it does look a bit strange to use a non-boolean expression with `if`. Granted it is pretty standard to leave off comparisons with `nullptr` for pointer variables.

DanRStevens (approved):
Looks good.

src/Sprite/TilesetLoader.cpp
Brett208: For some reason I was assuming that the value needed to be positive before applying the modulus operator. Good point that it isn't needed. 

I meant to use != 0 here and accidently left it off. I guess I got lucky and the behaviour was still correct accidently. 

I  will plan to push a change tomorrow removing the std::abs call and including != 0 unless you want to push today.

Also, I forgot to put the math include for abs, but something else in the standard library must have brought it into scope...

PR #362, opened 2020-12-30
Color swap improvements

Brett208:
Discussed in previous PR that Color:Invert may be a confusing term to use due to color inversion existing.

### DIFF
--- a/src/Bitmap/BitmapFile.h
+++ b/src/Bitmap/BitmapFile.h
@@ -47,6 +47,8 @@ public:
 
 	void Validate() const;
 
+	void SwapRedAndBlue();
+
 private:
 	static void VerifyIndexedImageForSerialization(uint16_t bitCount);
 
--- a/src/Bitmap/BitmapFile.h
+++ b/src/Bitmap/BitmapFile.h
@@ -23,7 +23,8 @@ public:
 	std::vector<Color> palette;
 	std::vector<uint8_t> pixels;
 
-	static BitmapFile CreateDefaultIndexed(uint16_t bitCount, uint32_t width, uint32_t height);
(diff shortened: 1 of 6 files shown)


src/Bitmap/BitmapFile.h
```diff
@@ -23,7 +23,8 @@ class BitmapFile
 	std::vector<Color> palette;
 	std::vector<uint8_t> pixels;
 
-	static BitmapFile CreateDefaultIndexed(uint16_t bitCount, uint32_t width, uint32_t height);
+	// @param initialColor will set the first palette entry to the provided color
+	static BitmapFile CreateDefaultIndexed(uint16_t bitCount, uint32_t width, uint32_t height, Color initialColor = DiscreteColor::Black);
```
DanRStevens, line 27: This seems very specific to helping write the test code. That makes it of somewhat questionable design. In particular, it's very curious that a factory method allows constructing an index bitmap while specifying exactly one of the palette colors. More natural would be if you could specify an entire palette.

It should be fairly easy pass an entire palette, allowing an arbitrary number of colors to be set, and allowing any unset colors to default to black.

Of course that would impact error checking, where the bit depth would need to correspond with the palette size. It may also necessitate copying of the palette, which may be inefficient. Perhaps an overloaded method would make more sense than a default parameter for something like that.

DanRStevens (approved):
Added a note about a potential future design update.

src/Bitmap/BitmapFile.h
Brett208: I'm going to merge this so the BitmapFile::SwapRedAndBlue is available in the code base. Would like to circle back around and give passing the whole palette a crack.

PR #363, opened 2020-12-31
Pass palette on bitmap creation

Brett208:
From recommendation in PR#362.

While looking at the naming scheme, I decided having the word Default in `BitmapFile::CreateDefaultIndexed` didn't add any value, so removed it.

### DIFF
--- a/src/Bitmap/BitmapFile.cpp
+++ b/src/Bitmap/BitmapFile.cpp
@@ -21,14 +21,14 @@ BitmapFile BitmapFile::CreateIndexed(uint16_t bitCount, uint32_t width, uint32_t
 	return bitmapFile;
 }
 
-BitmapFile BitmapFile::CreateIndexed(uint16_t bitCount, uint32_t width, uint32_t height, const std::vector<Color>& palette)
+BitmapFile BitmapFile::CreateIndexed(uint16_t bitCount, uint32_t width, uint32_t height, std::vector<Color> palette)
 {
 	if (palette.size() > std::size_t(1) << bitCount) {
 		throw std::runtime_error("Unable to create bitmap. Provided palette length is greater than provided bit count.");
 	}
 
 	auto bitmapFile = BitmapFile::CreateIndexed(bitCount, width, height);
-	std::copy(palette.begin(), palette.end(), bitmapFile.palette.begin());
+	std::move(palette.begin(), palette.end(), bitmapFile.palette.begin());
 
 	return bitmapFile;
 }
--- a/src/Bitmap/BitmapFile.cpp
+++ b/src/Bitmap/BitmapFile.cpp
@@ -2,7 +2,7 @@
 #include <stdexcept>
 #include <cmath>
 
-BitmapFile BitmapFile::CreateDefaultIndexed(uint16_t bitCount, uint32_t width, uint32_t height)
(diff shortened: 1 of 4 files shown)


src/Bitmap/BitmapFile.cpp
```diff
@@ -22,6 +21,18 @@ BitmapFile BitmapFile::CreateDefaultIndexed(uint16_t bitCount, uint32_t width, u
 	return bitmapFile;
 }
 
+BitmapFile BitmapFile::CreateDefaultIndexed(uint16_t bitCount, uint32_t width, uint32_t height, const std::vector<Color>& palette)
```
DanRStevens: Since you're copying the palette anyway, you could potentially pass by-value, and then `std::move` from the parameter to the field. That would effectively move the copy up-front to before the call for lvalue semantics, and pass in a temporary directly for rvalue semantics. That means you can avoid the copy altogether for rvalue semantics, and have only a single constructed `std::vector`.

The example uses in the unit test code would make use of the rvalue semantics.

DanRStevens (approved):
Looks great.

Noticed one potential performance improvement for rvalue semantics.

src/Bitmap/BitmapFile.cpp
```diff
@@ -21,14 +21,14 @@ BitmapFile BitmapFile::CreateIndexed(uint16_t bitCount, uint32_t width, uint32_t
...
-BitmapFile BitmapFile::CreateIndexed(uint16_t bitCount, uint32_t width, uint32_t height, const std::vector<Color>& palette)
+BitmapFile BitmapFile::CreateIndexed(uint16_t bitCount, uint32_t width, uint32_t height, std::vector<Color> palette)
 {
 	if (palette.size() > std::size_t(1) << bitCount) {
 		throw std::runtime_error("Unable to create bitmap. Provided palette length is greater than provided bit count.");
 	}
 
 	auto bitmapFile = BitmapFile::CreateIndexed(bitCount, width, height);
-	std::copy(palette.begin(), palette.end(), bitmapFile.palette.begin());
+	std::move(palette.begin(), palette.end(), bitmapFile.palette.begin());
```
DanRStevens: Ahh, I was thinking:
```cpp
bitmapFile.palette = std::move(palette);
```

I didn't realize there were two versions of `std::move`:
- [`std::move` from `<utility>`](https://en.cppreference.com/w/cpp/utility/move)
- [`std::move` from `<algorithm>`](https://en.cppreference.com/w/cpp/algorithm/move)

Brett208: Yeah, I just read up on the difference a little bit. 

I didn't want to use `bitmapFile.palette = std::move(palette);` as recommended because it opens a discussion about if you can initialize the bitmap with less than a full palette. I think that may be fine, but I wasn't ready to work through some of the implications where this solution will keep the palette at full length of the bit field even if a smaller palette is passed in.

I think I will merge as is. @DanRStevens, no problem if you want to work an improved implementation or writeup an improvement in an issue, but I don't want to stray too far from the tileset conversion work right now.

DanRStevens: That's fine. We may want to revisit later though. Especially in regards to potentially uninitialized palette entries in case a smaller palette is passed in. Or possibly referencing non-existent palette entries if a shorter palette is stored. We may want to refer back to the bitmap file format to see what is supported there.

Issue #348, opened 2020-11-26
Add support for custom bitmap format


DanRStevens:
For some reason I thought this project already had support for the custom bitmap format used for tilesets, particularly since the [OP2MapImager](https://github.com/OutpostUniverse/OP2MapImager) project can image map files. Though after checking both projects, it seems we don't have such code.

I suppose we can go back to the [OP2Editor](https://github.com/OutpostUniverse/OP2Editor) project and base something off of the code there. The relevant method is `CTileSet::LoadTileSet`:
https://github.com/OutpostUniverse/OP2Editor/blob/e2d7b1a4cc75c88e7ec8b155784af2f517c6e4e4/CTileSet.cpp#L349

I figure the basic loading code should go into OP2Utility. The same code could be re-used by both the [OP2MapImager](https://github.com/OutpostUniverse/OP2MapImager) and [OP2BitmapConverter](https://github.com/OutpostUniverse/OP2BitmapConverter) projects.

Brett208:
In particular from `CTileSet::LoadTileSet`
```C++
// Adjust palette color component positions (RGB <-> BGR)
for (i = 0; i < bmInfo->bmiHeader.biClrUsed; i++)
{
	// Swap the Red and Blue components
	temp = bmInfo->bmiColors[i].rgbRed;
	bmInfo->bmiColors[i].rgbRed = bmInfo->bmiColors[i].rgbBlue;
	bmInfo->bmiColors[i].rgbBlue = temp;
}
```

DanRStevens:
Ahh yes, that's an important point.

Could probably update that code to use [`std::swap`](https://en.cppreference.com/w/cpp/algorithm/swap).

Unless perhaps we have two different structs for RGB and BGR style color structs, and have a conversion method that copies like components in a way that naturally reverses the order. Though I may be overthinking things.

Brett208:
Good idea to use std::swap. I was thinking about adding a function to the Color struct
```C++
// Invert color values. RGB <-> BGR
void Color::Invert()
{
    // use std::swap
}
```

---

So a standard Bitmap uses a file signature of BM. The Outpost 2 file format uses a file signature of PB. f

@DanRStevens, do you know if the PB file format is a fully custom format written to support Outpost 2? A very brief google search seemed to indicate it wasn't recognized by others. I was reviewing the notes we have on the PB structure and it appears it contains a lot of the building blocks of a bitmap file format but leaves out assumed keys and values.

[...]

PR #364, opened 2021-01-02
Read custom tilesets

Brett208:
Would like to add some more unit tests at some point.

Related to Issue #348.

### DIFF
--- a/src/Sprite/TilesetCommon.h
+++ b/src/Sprite/TilesetCommon.h
@@ -1,0 +1,15 @@
+// Includes functions and constexpr variables that are useful in multiple tileset
+// source code files to prevent circular include references
+
+#pragma once
+
+#include <string>
+#include <cstdint>
+
+namespace Tileset
+{
+	void throwReadError(std::string propertyName, std::string value, std::string expectedValue);
+	void throwReadError(std::string propertyName, uint32_t value, uint32_t expectedValue);
+
+	std::string formatReadErrorMessage(std::string propertyName, std::string value, std::string expectedValue);
+}
--- a/src/Sprite/TilesetCommon.h
+++ b/src/Sprite/TilesetCommon.h
@@ -3,11 +3,15 @@
 
 #pragma once
 
+#include "../Tag.h"
 #include <string>
 #include <cstdint>
 
 namespace Tileset
 {
+	constexpr auto DefaultTagData = MakeTag("data");
+	constexpr uint32_t DefaultPaletteHeaderSize = 1024;
+
 	void throwReadError(std::string propertyName, std::string value, std::string expectedValue);
 	void throwReadError(std::string propertyName, uint32_t value, uint32_t expectedValue);
 
--- a/OP2Utility.vcxproj
+++ b/OP2Utility.vcxproj
@@ -118,6 +118,7 @@
     </Link>
   </ItemDefinitionGroup>
   <ItemGroup>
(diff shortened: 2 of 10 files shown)


src/Sprite/TilesetCommon.h
```diff
@@ -0,0 +1,19 @@
...
+#include <string>
+#include <cstdint>
+
+namespace Tileset
+{
+	constexpr auto DefaultTagData = MakeTag("data");
+	constexpr uint32_t DefaultPaletteHeaderSize = 1024;
+
+	void throwReadError(std::string propertyName, std::string value, std::string expectedValue);
+	void throwReadError(std::string propertyName, uint32_t value, uint32_t expectedValue);
```
Brett208, line 16: I wanted this function to be a template that allowed passing in any basic type that std::to_string could be applied to. I couldn't figure out how to write the template function though. Then I realized almost all cases in the current code base use uint32_t so this worked ok for now.

```C++
template <typename T>
std::string formatErrorMessage(std::string propertyName, T value, T expectedValue) {
	formatErrorMessage(propertyName, std::to_string(value), std::to_string(expectedValue);
}
```

DanRStevens (approved):
Looks good

src/Sprite/TilesetCommon.h
Brett208: Looks like std::to_string is not being as robust as I would have hoped here. After implementing the suggestion above, I am getting the following errors. I think the tileset error handling will need a bit of polishing as there is some casting going on, some of which is going from signed to unsigned and may produce some unexpected values in the error message. 

If templates don't work, maybe a signed and an unsigned function overload. I guess those overloads could be 64 bit integers or even doubles so when casting it doesn't overfill the integer size.

I'm leaning towards merging the codebase in as it stands and addressing in separate issues since this changeset is so large. 

-----

Error (active)	E0304	no instance of overloaded function "Tileset::throwReadError" matches the argument list	OP2Utility	\OP2Utility\src\Sprite\TilesetLoader.cpp	78	
    argument types are: (const char [10], const uint16_t, const uint32_t)

Error (active)	E0304	no instance of overloaded function "Tileset::throwReadError" matches the argument list	OP2Utility	\OP2Utility\src\Sprite\TilesetLoader.cpp	82	
    argument types are: (const char [12], const int32_t, const uint32_t)

Error	C2665	'std::to_string': none of the 9 overloads could convert all the argument types	OP2Utility	\OP2Utility\src\Sprite\TilesetCommon.h	22	

Error	C2660	'Tileset::throwReadError': function does not take 1 arguments	OP2Utility	\OP2Utility\src\Sprite\TilesetCommon.h	22	

Error	C2664	'void Tileset::throwReadError(std::string,std::string,std::string)': cannot convert argument 2 from 'const uint16_t' to 'std::string'	OP2Utility	\OP2Utility\src\Sprite\TilesetLoader.cpp	66	

Error	C2664	'void Tileset::throwReadError(std::string,std::string,std::string)': cannot convert argument 2 from 'const int32_t' to 'std::string'	OP2Utility	\OP2Utility\src\Sprite\TilesetLoader.cpp	70

DanRStevens: Yeah, just merge and worry about it later.

This sort of problem was address in the NAS2D library, so you might find some inspiration from there. Take a look at the `StringUtils.h` file.

PR #366, opened 2021-01-08
Read tileset

Brett208:


Brett208:
@DanRStevens,

Currently, BItmapFile uses static functions to Write bitmaps. I'm trying to decide why I structured it that way. What do you think about removing the static functions with member functions? Maybe just:

``` C++
BitmapFile::WriteIndexed(Stream::BidirectionalWriter& writer);
BitmapFile::WriteIndexed(std::string& filename); // Helper to automatically create a FileWriter stream
```

Instead of:

``` C++
// Current Function Stack
static void WriteIndexed(std::string filename, uint16_t bitCount, int32_t width, int32_t height, std::vector<Color> palette, const std::vector<uint8_t>& indexedPixels);

// I think the use case below is that you could collect all the bitmap information from different sources 
// instead of first forming a BitmapFile structure in memory. We could leave this one in the overloads, 
// but I'd probably just delete it for our use cases.
static void WriteIndexed(Stream::BidirectionalWriter& seekableWriter, uint16_t bitCount, int32_t width, int32_t height, std::vector<Color> palette, const std::vector<uint8_t>& indexedPixels);

static void WriteIndexed(std::string filename, const BitmapFile& bitmapFile);
void WriteIndexed(Stream::BidirectionalWriter& writer) const; 
```

Probably actually update code base in a different branch if this looks good.

DanRStevens:
I think I agree with using member functions for write. Having a static factory method for read makes sense, though I would generally use a member function for write.

src/Sprite/TilesetLoader.cpp
```diff
@@ -21,6 +21,27 @@ namespace Tileset
...
+	{
+		if (PeekIsCustomTileset(reader)) {
+			return ReadCustomTileset(reader);
+		}
+
+		BitmapFile tileset;
+		
+		try {
+			tileset = BitmapFile::ReadIndexed(reader);
+		}
```
DanRStevens, line 34: This would technically default initialize `tileset`, and then copy assign `tileset`. Moving the declaration into the `try` block would allow for single stage initialization. It would also mean `BitmapFile` wouldn't be required to have a default constructor. That may be a worthy consideration, since it's not immediately clear what kind of usable `BitmapFile` could be generated without any arguments.

I do understand the structure of the remaining code would be affected by such a change, due to scope changes. Though I suppose the rest of the function could be moved into the `try` block. That would affect exception generation from `ValidateTileset`, which would be caught and translated by the `catch` block. Though the `throw` does include the original exception message in the new exception message, so that may be fine.

test/Sprite/TilesetLoader.test.cpp
```diff
@@ -27,14 +27,16 @@ TEST(TilesetLoader, ReadTileset)
 	Stream::DynamicMemoryWriter writer;
 	tileset.WriteIndexed(writer);
 	
-	auto newTileset = Tileset::ReadTileset(writer.GetReader());
+	auto reader = writer.GetReader();
+	auto newTileset = Tileset::ReadTileset(reader);
```
DanRStevens: This is somewhat surprising. I feel like the original syntax should be made possible. Maybe an rvalue overload that delegates to an lvalue overload?

A stream is necessarily modifiable, as the stream position is mutable state, so a `const` reference is not possible. But only a const lvalue reference can bind to an rvalue. To accept a temporary, we need a method that accepts an rvalue. If we had only an rvalue overload, we'd need to cast or `std::move` a named variable to pass an lvalue. Seems like we would need both overloads here to support all typical calling cases.

src/Bitmap/BitmapFile.h
```diff
@@ -35,6 +35,7 @@ class BitmapFile
 	static void WriteIndexed(std::string filename, uint16_t bitCount, int32_t width, int32_t height, std::vector<Color> palette, const std::vector<uint8_t>& indexedPixels);
 	static void WriteIndexed(Stream::BidirectionalWriter& seekableWriter, uint16_t bitCount, int32_t width, int32_t height, std::vector<Color> palette, const std::vector<uint8_t>& indexedPixels);
 	static void WriteIndexed(std::string filename, const BitmapFile& bitmapFile);
+	void WriteIndexed(Stream::BidirectionalWriter& writer) const;
```
DanRStevens, line 38: Is bidirectional really needed here? For reading, the bidirectional helps with peeking to determine the bitmap type. I don't think it would be required for writing though. Should be able to write out the whole bitmap in one go.

DanRStevens (approved):

test/Sprite/TilesetLoader.test.cpp
Brett208: I will have to research more on lvalule/rvalue to implement your suggestions as my understanding is pretty weak of the concept. 

I agree with the end state of being able to make the call based on my original syntax. MSVC allowed for the previous behaviour although I have seen this sort of call fail on Linux based compilers in the past. I think MSVC is being a bit more permissive than specified by the C++ standard by allowing it to compile in the first place.

src/Bitmap/BitmapFile.h
Brett208: The underlying static function behind this call takes a bidirectional writer. Although you are right it doesn't need to seek. Stemming from sloppy work by me in the past.

Sounds like we are on the same page to move from static to member function calls for WriteIndexed. Will plan to switch to WriteIndexed taking a simple Writer interface at that point as well.

src/Sprite/TilesetLoader.cpp
Brett208: I thought about using a unique_ptr to represent `tileset` so it could be initialized within the try block without a default initialization. I didn't want to deal with a pointer though if I could help it although maybe it isn't that big of a deal.

I also considered what you brought up about just allowing the error message to encapsulate the outer throw. The message will appear a bit duplicative to the end reader, but still clear enough. Maybe this is the easier solution and worth avoiding a pointer?

test/Bitmap/IndexedBmpWriter.test.cpp
```diff
@@ -12,8 +12,8 @@ TEST(BitmapFile, InvalidBitCountThrows)
 	const std::vector<uint8_t> pixels(4, 0);
 	const std::string filename = "Sprite/data/MonochromeTest.bmp";
 
-	EXPECT_THROW(BitmapFile::WriteIndexed(filename, 3, 1, 1, palette, pixels), std::runtime_error);
-	EXPECT_THROW(BitmapFile::WriteIndexed(filename, 32, 1, 1, palette, pixels), std::runtime_error);
+	EXPECT_THROW(BitmapFile::CreateIndexed(3, 1, 1, palette, pixels).WriteIndexed(filename), std::runtime_error);
```
Brett208, line 15: Most of these tests are writing to disc during their tests. I suspect I wrote these tests before the `Stream::DynamicMemoryReader` class existed. I'd like to edit the majority of these tests to write to memory dynamically in a future branch. Maybe leave one to test file writing. The test code should become cleaner and maybe run faster.

test/Sprite/TilesetLoader.test.cpp
Brett208: Oh, right. I remember this now. C++ is maddening trying to remember all the nuances of the language.

We implemented this in the Archive code for writing volumes / clm files for the same reason.

src/Sprite/TilesetLoader.cpp
DanRStevens: Yeah, I like to avoid pointers when possible. Heap allocations and not nearly as cheap as local stack allocations.

C++ is usually pretty good at providing alternatives to avoid pointers unless they are actually strictly required from a theoretical standpoint. It doesn't normally make sense to use pointers for easier syntax.

test/Bitmap/IndexedBmpWriter.test.cpp
DanRStevens: I approve of in memory tests. Faster, and less likely to leave crud around on disk, especially if the unit tests fail in the middle of a test. That has the potential to impact future test runs, in case they are sensitive to files already existing, or having the wrong write permissions or user ownership.

Might make sense to assume stream writing to disk is tested by the stream unit tests, and just assume it is good here. No need to test stream code in the bitmap tests.

src/Bitmap/BitmapFile.cpp
```diff
@@ -33,6 +33,14 @@ BitmapFile BitmapFile::CreateIndexed(uint16_t bitCount, uint32_t width, uint32_t
 	return bitmapFile;
 }
 
+BitmapFile BitmapFile::CreateIndexed(uint16_t bitCount, uint32_t width, uint32_t height, std::vector<Color> palette, std::vector<uint8_t> pixels)
+{
+	auto bitmap = CreateIndexed(bitCount, width, height, palette);
+	bitmap.pixels = pixels;
```
DanRStevens, line 39: There's a potential for a `std::vector` size mismatch here. Not sure how sensitive the underlying code is to that.

DanRStevens (approved):
I'm loving these updates.

src/Bitmap/BitmapFile.cpp
Brett208: Good point.

src/Sprite/TilesetLoader.cpp
```diff
@@ -31,20 +31,16 @@ namespace Tileset
...
 		
 		try {
-			tileset = BitmapFile::ReadIndexed(reader);
+			auto tileset = BitmapFile::ReadIndexed(reader);
+			ValidateTileset(tileset);
+			return tileset;
 		}
 		catch (std::exception& e) {
-			// Repackage exceptions from bitmap reading to include source is tileset
+			// Repackage exceptions from bitmap reading to include bitmap exception is related to tileset loading
```
DanRStevens, line 41: This sounds a bit off grammatically.

src/Bitmap/BitmapFile.cpp
```diff
@@ -61,7 +62,7 @@ void BitmapFile::VerifyPixelSizeMatchesImageDimensionsWithPitch() const
 void BitmapFile::VerifyPixelSizeMatchesImageDimensionsWithPitch(uint16_t bitCount, int32_t width, int32_t height, std::size_t pixelsWithPitchSize)
 {
 	if (pixelsWithPitchSize != ImageHeader::CalculatePitch(bitCount, width) * std::abs(height)) {
-		throw std::runtime_error("The size of pixels does not match the image's height time pitch");
+		throw std::runtime_error("The size of pixels does not match the image's height times pitch");
```
DanRStevens, line 65: Very technically accurate. If an end user sees this, or many developers, they may have a weak understanding of what `pitch` is. When I first read this, I mis-read pitch to mean bit depth. I thought it was a small mistake where width had been forgotten. I'm not sure I have any suggestions for how to improve, just wanted to mention this might be easily misunderstood.

PR #371, opened 2021-01-14
Fix syntax error in TilesetLoader.cpp

Brett208:
Caught by gcc and clang


DanRStevens (approved):
Hmm, maybe we should enable warnings as errors for the Linux builds so things like that are easier to notice.

Brett208:
I'm happy with the current setup for automated builds.

PR #372, opened 2021-01-16
Tileset validation type inconsistencies

Brett208:
This fixes several issues with error messaging for tilesets mentioned in issue #348. 

Pending discussion in #348, can actually implement a templated solution for tileset read error messaging.

Part of the problems were because I was straight up not reporting the value parameter, which I should have caught earlier.

### DIFF
--- a/src/Sprite/TilesetLoader.cpp
+++ b/src/Sprite/TilesetLoader.cpp
@@ -119,10 +119,10 @@ namespace Tileset
 	{
 		constexpr int32_t DefaultPixelWidth = 32;
 		constexpr uint32_t DefaultPixelHeightMultiple = DefaultPixelWidth;
-		constexpr uint32_t DefaultBitDepth = 8;
+		constexpr uint16_t DefaultBitCount = 8;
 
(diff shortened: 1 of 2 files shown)


src/Sprite/TilesetLoader.cpp
```diff
@@ -117,7 +117,7 @@ namespace Tileset
 
 	void ValidateTileset(const BitmapFile& tileset)
 	{
-		constexpr uint32_t DefaultPixelWidth = 32;
+		constexpr int32_t DefaultPixelWidth = 32;
 		constexpr uint32_t DefaultPixelHeightMultiple = DefaultPixelWidth;
```
DanRStevens, line 121: Curiously the width and height values are now of different type. Not sure that will lead to any problems, just that it's unexpected.

DanRStevens (approved):
Looks good.

I'm not particularly attached to either BitDepth or BitCount. I'm more used to BitDepth, but if the official documentation says BitCount, lets go with what the documentation says.

PR #374, opened 2021-01-18
String from

Brett208:
Related to #348 

We may want to port the update to StringFrom back into NAS2D if this looks acceptable


Brett208:
I snuck in some changes to what was StringHelper (now file StringUtility). If any of it seems controversial, we can back those changes out before merging.

src/StringUtility.cpp
```diff
@@ -2,96 +2,99 @@
...
+			c = toupper(c);
 		}
 	}
 
-	return stringsToSearch;
-}
+	std::string ConvertToUpper(const std::string& str)
+	{
+		std::string newString = str;
+		ConvertToUpper(newString);
```
DanRStevens, line 17: A bit unrelated to the current change though this method could take the parameter by-value.

The `ConvertToUpper` function is making a copy of the `str` parameter, so might as well use by-value semantics which will push that copy to the calling code, and enable efficient move behaviour for rvalues.

src/StringUtility.h
```diff
@@ -19,6 +19,9 @@ namespace StringUtility
 		if constexpr (std::is_same_v<T, std::string> || std::is_same_v<T, const char*>) {
 			return value;
 		}
+		else if constexpr (std::is_convertible<T, std::string>::value) {
+			return value;
+		}
```
DanRStevens, line 24: I just realized we should probably check how this behaves when the conversion function is marked with `explicit`. Do we need an explicit cast here to handle that case? If we do, it should perhaps be:
```cpp
return static_cast<std::string>(value);
```

----

Not sure if [`std::is_convertible`](https://en.cppreference.com/w/cpp/types/is_convertible) detects when `explicit` is used or not, or if there is a different function that checks differently.

test/StringUtility.test.cpp
```diff
@@ -4,6 +4,15 @@
 
 using namespace StringUtility;
 
+class ImplicitStringConversionTestFixture {
+public:
+	std::string data;
+
+	operator std::string() const {
+		return data;
+	}
```
DanRStevens, line 13: We can try adding `explicit` to this method to check the previous suggestion.

```cpp
explicit operator std::string() const
```

----

https://en.cppreference.com/w/cpp/language/explicit

test/StringUtility.test.cpp
```diff
@@ -39,4 +48,8 @@ TEST(String, stringFrom) {
 	EXPECT_THAT(StringFrom(0.0f), testing::StartsWith("0.0"));
 	EXPECT_THAT(StringFrom(0.0), testing::StartsWith("0.0"));
 	EXPECT_THAT(StringFrom(0.0l), testing::StartsWith("0.0"));
+
+	// Implicit string conversion
+	ImplicitStringConversionTestFixture implicitStringConversionTestFixture{ "testString" };
+	EXPECT_EQ("testString", StringFrom(implicitStringConversionTestFixture));
```
DanRStevens, line 54: This is testing the test code. Perhaps it should be separated from the `stringFrom` test code.

Or is this really testing the compiler itself? Should we be testing this? Maybe we should assume the compiler does the correct thing and call the `std::string` conversion operator.

DanRStevens (approved):
Good idea to port the update back to NAS2D.

Changes look good. I like the improvements.

Had a few additional thoughts for future consideration.

----

So now that you have this working, should you consider a template method that does the comparison between two values, and if they are different throws an exception with a nicely formatted message showing both values? Might be a sensible verify style method. I mean, the comparison between two values of arbitrary type seems to be duplicated in a few places. :wink:

src/StringUtility.cpp
Brett208: I like this idea, but got the following error message when trying to implement:

```C++
void ConvertToUpper(std::string& str);
std::string ConvertToUpper(std::string str);
```

Error (active)	E0308	more than one instance of overloaded function "StringUtility::ConvertToUpper" matches the argument list:	OP2Utility \OP2Utility\src\StringUtility.cpp 16 
* function "StringUtility::ConvertToUpper(std::string &str)" (declared at line 7)
* function "StringUtility::ConvertToUpper(std::string str)" (declared at line 14)
            argument types are: (std::string)

Not sure if there is a way around this unless we change one of the function names?

src/StringUtility.h
Brett208:
```C++
class ImplicitStringConversionTestFixture {
public:
	std::string data;

	explicit operator std::string() const {
		return data;
	}
};
```

Causes

Error	C2665	'std::to_string': none of the 9 overloads could convert all the argument types	OP2UtilityTest	\OP2Utility\src\StringUtility.h	29	

Based on the compiler error and some quick reading on the internet, it looks like std::is_convertible returns false if explicit conversion is required.

There are some references to a boost is_explicitly_convertible and a stack overflow article mentioning a solution I couldn't fully follow https://stackoverflow.com/questions/16893992/check-if-type-can-be-explicitly-converted.

Trying to decide if supporting explicit conversion is worth pursuing or not. If worth pursuing, maybe we should move this to an issue for future work unless there is a quick solution so it isn't lost in this closed PR?

test/StringUtility.test.cpp
Brett208: I would defer to your thoughts if it should be moved or altogether removed. I was trying to test that implicit string conversion worked without throwing an error, etc.

src/StringUtility.cpp
DanRStevens: Ahh, right, those are two different functions. They probably should be named differently, rather than overloaded. One modifies in-place, while the other returns a modified copy. You might want to be able to return a modified copy, rather than modify in-place, even if a variable is not marked `const`. This was probably a bad place to use an overload. If you had a non-const variable that you wanted preserved, you'd have to cast it to `const` during the call if you wanted to select the overload that doesn't modify it. That's not super convenient.

src/StringUtility.h
DanRStevens: Ahh, ok. Let's not worry about it then. I think I was mostly concerned that the constexpr-if could select that case when the code for it didn't always apply, and lead to unexpected and hard to diagnose errors.

It seems the current implementation matches up perfectly with how the `std::is_convertible` check works.

I suppose we could create an issue for it if we wanted to revisit, though it's probably pretty low value, so we might not want to spend the time on it. It may require more effort than is worthwhile to support that case.

test/StringUtility.test.cpp
DanRStevens: Wait a minute, after re-reading this with fresh eyes it's making more sense. Yes, we do want to test that.

I think my tired eyes had deleted the `StringFrom` call. Without it, it would just be testing that the test object can be automatically converted to a string in an equality comparison. With that, it does actually test the `StringFrom` project code, and makes sure it can handle types with auto conversion operators.

We should keep it.

PR #377, opened 2021-01-20
ConvertToUpper improvements

Brett208:


test/StringUtility.test.cpp
```diff
@@ -4,6 +4,24 @@
...
+}
+
+TEST(StringUtility, ConvertToUpper)
+{
+	std::string str("abc");
+	std::string newStr = ConvertToUpper("abc");
+
+	EXPECT_EQ("ABC", newStr);
+	EXPECT_NE(str, newStr);
+}
```
DanRStevens, line 22: Was curious if you'd meant to write `ConvertToUpper(str)`. That would make the final expectation make a bit more sense.

I really like the structure of the new unit tests.

DanRStevens (approved):

test/StringUtility.test.cpp
Brett208: Oh, yeah good catch. I'll fix shortly.

DanRStevens (approved):

PR #378, opened 2021-01-20
Bitmap file tests using dynamic memory writer

### LINKED ISSUE #367: Use DynamicMemoryReader for Bitmap testing
Much of the bitmap file code currently tests by writing and then reading from disk. The DynamicMemoryReader can be used to remove the need to write to disk in the testing. This will reduce the complexity of the test code since there will be no need to cleanup the filesystem and may improve test run time.

Identified in PR #366 as a future enhancement.

Brett208:
Closes #367


DanRStevens (approved):
I really like how this simplifies the unit test code.

PR #380, opened 2021-01-23
BitmapFile scan line orientation

Brett208:
Most of the intermediate commits in this changeset break the tileset unit tests due to the original tests not taking into account proper Windows bitmap file scan line orientation.

### DIFF
--- a/src/Bitmap/BitmapFile.cpp
+++ b/src/Bitmap/BitmapFile.cpp
@@ -82,6 +82,11 @@ void BitmapFile::Validate() const
 	VerifyPixelSizeMatchesImageDimensionsWithPitch();
 }
 
+ScanLineOrientation BitmapFile::ScanLineOrientation() const
+{
+	return imageHeader.height < 0 ? ScanLineOrientation::TopDown : ScanLineOrientation::BottomUp;
+}
+
 void BitmapFile::SwapRedAndBlue()
 {
 	for (auto& color : palette) {
--- a/src/Bitmap/BitmapFile.cpp
+++ b/src/Bitmap/BitmapFile.cpp
@@ -82,7 +82,7 @@ void BitmapFile::Validate() const
 	VerifyPixelSizeMatchesImageDimensionsWithPitch();
 }
 
-ScanLineOrientation BitmapFile::ScanLineOrientation() const
+ScanLineOrientation BitmapFile::GetScanLineOrientation() const
 {
 	return imageHeader.height < 0 ? ScanLineOrientation::TopDown : ScanLineOrientation::BottomUp;
 }
--- a/src/Bitmap/BitmapFile.cpp
+++ b/src/Bitmap/BitmapFile.cpp
@@ -99,6 +99,22 @@ void BitmapFile::SwapRedAndBlue()
 	}
 }
 
+void BitmapFile::InvertScanLines()
+{
+	imageHeader.height *= -1;
+
+	std::vector<uint8_t> invertedPixels;
+
+	const auto pitch = imageHeader.CalculatePitch();
+	for (auto row = AbsoluteHeight(); row; --row)
+	{
+		auto iterator = invertedPixels.end();
+		invertedPixels.insert(iterator, pixels.begin() + (row - 1) * pitch, pixels.begin() + row * pitch);
+	}
+
+	pixels = std::move(invertedPixels);
+}
+
 bool operator==(const BitmapFile& lhs, const BitmapFile& rhs) {
 	return lhs.bmpHeader == rhs.bmpHeader && 
 		lhs.imageHeader == rhs.imageHeader && 
--- a/src/Bitmap/BitmapFile.cpp
+++ b/src/Bitmap/BitmapFile.cpp
@@ -2,12 +2,12 @@
 #include <stdexcept>
 #include <cmath>
 
-BitmapFile BitmapFile::CreateIndexed(uint16_t bitCount, uint32_t width, uint32_t height)
+BitmapFile BitmapFile::CreateIndexed(uint16_t bitCount, uint32_t width, int32_t height)
 {
 	BitmapFile bitmapFile;
 	bitmapFile.imageHeader = ImageHeader::Create(width, height, bitCount);
 	bitmapFile.palette.resize(bitmapFile.imageHeader.CalcMaxIndexedPaletteSize());
-	bitmapFile.pixels.resize(bitmapFile.imageHeader.CalculatePitch() * height);
+	bitmapFile.pixels.resize(bitmapFile.imageHeader.CalculatePitch() * std::abs(height));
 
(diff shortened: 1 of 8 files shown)


DanRStevens:
I'll probably review in the morning, though I suspect this change may require a bit of documentation. In particular, what is the natural orientation, where are negative heights allowed, and what corrections are performed.

I looked briefly and noticed you seem to be using negative height parameters to function calls, so it seems negative heights aren't restricted reading bitmaps from disk.

Brett208:
Windows BitmapFile specification is TopDown if the height is set to negative in the Image Header. Since I've been working closely with the documentation I forgot that is unintuitive and you are right should be documented.

This branch is about fixing the fact that the source code was ignorant of the concept of scan line orientation. 

I would also be ok changing the overloads of BitmapFile::CreateIndexed to include the new enum ScanLineOrientation as an input argument and switching back to an unsigned value for height on creation. . This would allow documentation on function call instead of code comments. CreateBitmap already has quite a few input parameters when adding a palette and pixels on creation, but creating a bitmap file is fairly complex, so maybe it is ok to add another parameter. If adding a ScanLineOrientation parameter to CreateIndexed sounds better to you than documenting, I'll make the change. Maybe there are other solutions too though that I haven't thought of.

> Microsoft bitmaps are usually stored from bottom to top. When the height in the bitmap header is negative then it means that the image is stored from top to bottom, and its real height is abs(height specified in the header).

DanRStevens:
Well it's more that the internal format you use need not necessarily match the Windows bitmap file format specification exactly. It would be perfectly reasonable for your internal format to always use positive `height` values, and to always store data in memory as top-down.

The custom OP2 bitmap format I think is stored with positive `height` values, and as top-down. Most image formats are probably the same. Many image editors with Windows bitmap file support may be able to load bottom-up images, but will only ever save a top-down image. Quite likely they may use a positive height value for top-down images, and adjust pixel data during load for bottom-up bitmaps.

It's because of the possible discrepancy between the Windows bitmap file format and whatever in-memory format is chosen that I ask. I would recommend designing an API that is intuitive to use, more so than support writing bitmap files in every conceivable format that could be supported. I would be perfectly fine if we had no support for written bottom-up images (though it's probably a good idea to at least support reading them).

src/Bitmap/BitmapFile.cpp
```diff
@@ -99,6 +99,22 @@ void BitmapFile::SwapRedAndBlue()
...
+	imageHeader.height *= -1;
+
+	std::vector<uint8_t> invertedPixels;
+
+	const auto pitch = imageHeader.CalculatePitch();
+	for (auto row = AbsoluteHeight(); row; --row)
+	{
+		auto iterator = invertedPixels.end();
+		invertedPixels.insert(iterator, pixels.begin() + (row - 1) * pitch, pixels.begin() + row * pitch);
+	}
```
DanRStevens, line 113: It may make sense to pre-allocate space for the inverted pixels with [`std::vector::reserve`](https://en.cppreference.com/w/cpp/container/vector/reserve). This should be easily calculated as matching `pixels.size()`.

Of course you'd need to update the use of `invertedPixels.end()` within the loop.

----

For a count to zero loop, you can use the pattern:
```cpp
for (auto row = AbsoluteHeight(); row-- > 0; )
```

This pattern works even for unsigned values. It also means you don't need to adjust the loop counter within the loop, so no `- 1` adjustment.

Note the [post-decrement](https://en.cppreference.com/w/cpp/language/operator_incdec) operator. It returns the value before the decrement operation is performed, which is then used in the comparison.
- `height` -> `height - 1`: Checks `height > 0`, loop variable is `height - 1` (first iteration)
- `1` -> `0`: Checks `1 > 0`, loop variable is `0` (last iteration)
- `0` -> wrap around to max: Checks `0 > 0` (`false`), loop variable is not used (stop iteration)

DanRStevens (approved):
Ahh, I see. Looks like you're trying to preserve an in-memory format that exactly matches the Windows bitmap file format. Seems like you've chosen to adjust formats only during write of the custom format. I'm guessing you're going to maximum flexibility in terms of what Windows bitmap formats you can manipulate natively.

src/Bitmap/BitmapFile.cpp
Brett208: I wasn't happy with the loop design. I like the idea of resizing the vector up front. I was struggling to find a vector function to copy elements over top of the reserved space instead of inserting additional elements. I wonder if an external to vector std::move or std::copy could be performed instead of using a vector member function. I may take a quick look at this before merging.

You are right that I was trying to represent the BitmapFile closely in the project instead of just writing to it. It seemed to make sense to me to stay close to the file format since OP2Utility is designed to support manipulating game assets and files and not really to actually present or play a game on screen if that makes sense. Although I can see some value in using our own format that is a smaller subset of the Windows Bitmap format and just writing / reading to the Windows bitmap format. I sort of liked the idea of the BitmapFile code being useful outside of Outpost 2 applications similar to the Stream library. Not that anyone will likely ever find it buried in OP2Utility and rip it out for use else ware... 

In the past I had used BitmapFile without making it aware that it could be presented a file with different scan line orientations or to even make sure height was set to negative as needed. I hope we now have some basic facilities to recognize the scan line orientation and flip it as needed. I think your points are valid that we don't need to really support saving to disk in multiple scan line formats even though we probably have the ability to do so now. We should probably encourage setting the height negative and using the top down scan line for all save operations.

DanRStevens: Reserved space doesn't count towards the `size`, so you'd still need to call `insert`. The only thing `reserve` does is prevent lots of needless allocation as you insert repeatedly into the `std::vector`.

----

I'm not sure if I've expressed my idea entirely clearly.

I would like to emphasize there are two separate decisions being made:
- Whether to store pixel data as top-down or bottom-up (either on disk, or in memory)
- Whether to represent top-down bitmaps internally with a negative height value (outside of the loading function)

We could choose be different on either or both of those decisions. These two settings are only strictly tied together when parsing Windows bitmap files from disk. There's no reason we need to maintain the same guarantee once data is in memory. It would be reasonable for API calls to always use positive height values. You could either translate values internally, or perhaps use an enum field to represent the data orientation.

Or you can effectively erase the difference. Always translate data to top-down when loading, provide a default `save` method for top-down format, and if people care, provide a separate `saveBottomUp` write method, and get rid of the enum. If they really want files written in the alternate old style bottom-up format, they can explicitly ask for it, otherwise default to always using top-down format. This of course forgets which format the data was originally in, though it's not clear that anyone would really care. I suspect this would provide the simplest cleanest API, while still being able to support both orientations for both load and save.

```cpp
if (height < 0) {
  height = -height;
} else {
  invertPixels();
}

// Data is now always top-down, and height is always positive
```

----

I think you should probably just merge and worry about adjustments in a followup PR. Though up to you if you want to adjust the loop design first.

PR #383, opened 2021-01-25
Add global namespace OP2Utility

### LINKED ISSUE #237: Wrap all of OP2Utility in a global namespace
As OP2Utility matures, it would be nice to wrap the entire library in a global namespace to help resolve eventual name collisions. 

Some discussion as to options and syntax were discussed in issue #139.

I think my preference would be OP2Utility. So it would look like:

```C++
OP2Utility::Map
OP2Utility::XFile::GetDirectory
OP2Utility::ArtFile
OP2Utility::BitmapFile
```

Some best practices are only capitalize all letters of an acronym if it is 2 letters or less. We are breaking this, but I think Op2Utility doesn't really look right. I think the community has adopted OP2 as an acronym strongly enough that we wouldn't want to change to something like O2Utility.

The C++ standard library does not use UpperCamelCase and uses underscores between words as well. Since we are using UpperCamelCase for functions and no underscores, I think formatting the namespace name as OP2Utility makes more sense then op2utility. This would make it look like op2utility::MapHeader or op2Utility::ResourceManager. (However all lowercase doesn't really look bad either, if @DanRStevens has a strong preference for op2utility, then I can just live with it). I would like to avoid any underscores though.

In practice, I would probably just declare using OP2Utility at the top of .cpp files of applications using OP2Utility. If a name collision occurred, I would then fall back on full specification.

As a parallel effort, I would be willing to remove the namespace Archives from the code base. Once we wrap the whole library in OP2Utility, then perhaps there isn't enough Archive specific code to justify a sub-namespace.

For now, I'm happy leaving the Stream namespace. So you would have OP2Utility::Stream::Reader. Again, I would probably call using on OP2Utility, so in practice you would see Stream::Reader.

Thanks,
Brett

Brett208:
Closes #237

### DIFF
--- a/src/XFile.cpp
+++ b/src/XFile.cpp
@@ -11,256 +11,258 @@ namespace fs = std::filesystem;
 namespace fs = std::experimental::filesystem;
 #endif
 
-
-std::string XFile::GetFileExtension(const std::string& pathStr)
+namespace XFile
 {
-	return fs::path(pathStr).extension().string();
-}
-
-bool XFile::IsDirectory(const std::string& pathStr)
-{
-	if (pathStr.length() == 0) {
-		return true;
+	std::string GetFileExtension(const std::string& pathStr)
+	{
+		return fs::path(pathStr).extension().string();
 	}
 
-	return fs::is_directory(pathStr);
-}
+	bool IsDirectory(const std::string& pathStr)
+	{
+		if (pathStr.length() == 0) {
+			return true;
+		}
 
-bool XFile::IsFile(const std::string& path)
-{
-	return fs::is_regular_file(path);
-}
-
-bool XFile::ExtensionMatches(const std::string& pathStr, const std::string& extension)
-{
-	std::string pathExtension = GetFileExtension(pathStr);
-	StringUtility::ConvertToUpperInPlace(pathExtension);
-
-	std::string extensionUpper = StringUtility::ConvertToUpper(extension);
-
-	if (extensionUpper.length() > 0 && extensionUpper[0] != '.') {
-		extensionUpper.insert(0, ".");
+		return fs::is_directory(pathStr);
 	}
 
-	return pathExtension == extensionUpper;
-}
-
-std::string XFile::ChangeFileExtension(const std::string& filename, const std::string& newExtension)
-{
-	return fs::path(filename).replace_extension(newExtension).string();
-}
-
-void XFile::NewDirectory(const std::string& newPath)
-{
-	fs::create_directory(fs::path(newPath));
-}
-
-bool XFile::PathExists(const std::string& pathStr)
-{
-	return fs::exists(fs::path(pathStr));
-}
-
-bool XFile::HasRootComponent(const std::string& pathStr)
-{
-	fs::path path(pathStr);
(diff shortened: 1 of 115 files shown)


src/XFile.cpp
```diff
@@ -11,256 +11,258 @@ namespace fs = std::filesystem;
...
-	}
 
-	return filenames;
-}
+		return filenames;
+	}
 
-std::vector<std::string> XFile::Dir(const std::string& directory, const std::regex& filenameRegex)
-{
-	return ::Dir(
```
Brett208, line 127: @DanRStevens I had to remove the global namespace specifier from the Dir function calls in XFile when I moved the namespace syntax from displaying on each function to a block style namespace call.

I wasn't sure while the global namespace specifier was there in the first place. Wanted to call this out considering how much noise there is in the commits.

Brett208:
I didn't think about all the extra tabs causing so much noise in the PR...

DanRStevens:
Ahh, as soon as I read the title, I realized this was going to have merge conflicts with the other two branches that were recently merged.

src/XFile.cpp
DanRStevens: It's there because there are multiple methods with the name `Dir`, some in the `XFile` namespace, and one in the global namespace (defined privately in XFile.cpp). Since the calling code is within a method in the `XFile` namespace, it seems to be resolving the name to the overloads defined in the `XFile` namespace, when the global template method was intended.

DanRStevens:
If you change nothing but the tabs, the diff for that commit is usually pretty easy to read. Though if you change so much as one other character, the diff tends to be extremely messy and hard to read.

In the interest of keeping diffs easy to read and review, sometimes I'll make an intermediate commit with a code change, but no corresponding indentation change. For instance, adding a namespace wrapper block, but not adding corresponding indentation. That means there will be a commit with bad indentation, though I think this is acceptable, as it's temporary in one commit, generally fixed in the following commit, and can make review much easier due to the simpler diffs.

Brett208:
@DanRStevens, If you would prefer, I am ok redoing this branch trying to separate the indentation change from the actual insertion of code change. If you do not prefer the rework, I will just fix the merge conflict. It is probably about 30 minutes for me to redo in a new branch which isn't a big deal assuming I can get Visual Studio's auto complete functions to assist me instead of hindering me...

Regarding the comment about not mixing template and non-template functions, I had not heard of this guideline but I am admittedly weak with templates. What are your thoughts about changing the name of the template function? I'm not experienced or picky about it either way. Could just name it DirInternal or DirTemplate.

I wanted to complete the namespace addition and a bit of work with the organization of the source code files before switching towards work in OP2BitmapConverter.

DanRStevens:
I was kind of thinking renaming the template function might be a good idea.

Not really a strong preference for rework versus patch up. Maybe a slight preference towards rework if other changes are going to be made. I suppose it would be easier to verify nothing goes badly trying to fix merge conflicts if there was some rework instead.

----

Downstream projects will need to be updated after this change is merged. Actually, as it stands, downstream projects might already need to be updated due to the StringHelper rename.

PR #384, opened 2021-01-28
Global namespace try2

### LINKED ISSUE #237: Wrap all of OP2Utility in a global namespace

Brett208:
Closes #237

Try this one instead...

### DIFF
--- a/src/Archive/AdaptiveHuffmanTree.cpp
+++ b/src/Archive/AdaptiveHuffmanTree.cpp
@@ -3,7 +3,7 @@
 #include <string>
 #include <stdexcept>
 
-namespace Archive
+namespace OP2Utility::Archive
 {
 	// Creates an (adaptive) Huffman tree with terminalNodeCount at the bottom.
 	AdaptiveHuffmanTree::AdaptiveHuffmanTree(NodeType terminalNodeCount) :
--- a/src/Archive/AdaptiveHuffmanTree.h
+++ b/src/Archive/AdaptiveHuffmanTree.h
@@ -1,6 +1,6 @@
 #include <vector>
 
-namespace Archive
+namespace OP2Utility::Archive
 {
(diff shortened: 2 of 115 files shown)


DanRStevens (approved):
Nicely structured changes. The diffs ended up being significantly easier to read this time around. Looks like git still managed to get a little bit confused with the diffs, though not by nearly as much. It was just a few small sections here and there that needed closer attention.

PR #385, opened 2021-01-31
Update master include

Brett208:
Probably should review readme.md in the near future as well.

Caught the name change during OP2Archive integration of the OP2Utility namespace changes.


Brett208:
I'm going to commit this without review to unblock an OP2Archive pull request

DanRStevens (commented):
Ahh yes, I like the idea for the test. I guess OP2Utility.h isn't used by the project itself.

PR #386, opened 2021-02-01
Remove reference to Visual Studio 2017

Brett208:
I think the goal is to remain relevant on a couple of rolling versions of Visual Studio that support C++17 feature sets instead of calling out only one version. Although I am currently testing against VS2019.


DanRStevens (approved):
Yeah, those details tend to stagnate. If someone really wants to know specific version information, they can check the CI config.

PR #388, opened 2021-02-04
Slice seek backward bug

Brett208:
Guess if I was doing this correctly according to test driven development, I should have written the unit test first to verify it would fail and then write the corrective test...

### DIFF
--- a/src/Stream/SliceReader.h
+++ b/src/Stream/SliceReader.h
@@ -75,7 +75,7 @@ namespace OP2Utility::Stream
 		}
 
 		void SeekBackward(uint64_t offset) override {
-			if (Position() - offset < startingOffset) {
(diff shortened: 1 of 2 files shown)


DanRStevens (approved):
Ahh yes, that is an easy one to get confused by.

We probably need to check `SeekForward`. It looks like it has the same sort of error. We'll probably need to add tests where `SeekForward` is used near the end of a sliced stream.

DanRStevens:
As for test driven development, yes, generally your write the tests first and verify the code fails, then fix the code. It's a way of verifying the test code actually works and tests what you think it's testing.

With that said, I usually commit the fix first, and then commit the new tests second. That way I avoid committing broken code where unit tests don't pass.

PR #389, opened 2021-02-04
Fix SeekForward bug

DanRStevens:
Related: #388

I updated the unit tests so they would actually catch the bug that was fixed in #388.

After pondering how to write a test for the current bug, I was not able to think of a way that could be easily inserted into existing unit test code. The comparison that leads to an exception basically goes the wrong way, where the old buggy code would not throw an exception when it should have. The only way to test the current code would be to check if it throws an exception when it should, however, not all stream classes that derive from `BidirectionalReader` will `throw` if you seek past the end. In particular, `FileReader` will not throw, so adding a `throw` check to the test will fail for that class.

I find the current stream tests somewhat hard to navigate and reason about. Maybe we should revisit the organization of that code at some point. I remember certain Google Test concepts were pretty new at that point, and we were still experimenting. Perhaps it would be clearer if we revisited the design now.

### DIFF
--- a/src/Stream/SliceReader.h
+++ b/src/Stream/SliceReader.h
@@ -63,7 +63,7 @@ namespace OP2Utility::Stream
 
 
 		void SeekForward(uint64_t offset) override {
-			if (Position() + offset > startingOffset + sliceLength)
+			if (Position() + offset > sliceLength)
 			{
 				throw std::runtime_error(
 					"Seek forward by offset of " + std::to_string(offset) + " is beyond the bounds of the stream slice."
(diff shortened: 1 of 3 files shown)


Brett208:
Oh, yeah good point about the slice starting place mattering... Funny that I mentioned test driven development but didn't use it and then wrote a test that didn't actually properly test the problem.

I would agree the test code could use some improvement. I suppose if the behaviour is purposefully different for different implementations of the abstract class, it should be tested individually outside of the core test cases... I guess since the slice reader could take a slice of something in memory or from file, it may have different boundary conditions to test against?

Brett208 (approved):

PR #390, opened 2021-02-17
Reader peek

Brett208:
I was on the fence about adding a lot of the helper template function versions of Read to Peek. Maybe in a future branch?

I think it was observed in another branch, but the structure of the Stream test code is poor. I implemented the test against the MemoryReader because of this.

### DIFF
--- a/src/Stream/BidirectionalReader.h
+++ b/src/Stream/BidirectionalReader.h
@@ -7,7 +7,7 @@ namespace OP2Utility::Stream
 {
 	class BidirectionalReader : public ForwardReader {
 	public:
-		inline void Peek(void* buffer, std::size_t size) {
+		void Peek(void* buffer, std::size_t size) {
 			ReadImplementation(buffer, size);
 			SeekBackward(size);
 		}
--- a/src/Stream/BidirectionalReader.h
+++ b/src/Stream/BidirectionalReader.h
@@ -6,10 +6,26 @@
 namespace OP2Utility::Stream
 {
 	class BidirectionalReader : public ForwardReader {
-	public:
-		void Peek(void* buffer, std::size_t size) {
+	protected:
+		// Generic peek method, which raises an exception if the full data cannot be read
+		// Note: This is named separately from Peek to prevent name hiding in derived classes
+		// All Peek methods are syntax sugar which delegate to this method
+		void PeekImplementation(void* buffer, std::size_t size)
+		{
 			ReadImplementation(buffer, size);
 			SeekBackward(size);
+		}
+
+	public:
+		void Peek(void* buffer, std::size_t size) {
+			PeekImplementation(buffer, size);
+		}
+
+		// Trivially copyable data types
+		template<typename T>
+		std::enable_if_t<std::is_trivially_copyable<T>::value>
+			Peek(T& object) {
+			PeekImplementation(&object, sizeof(object));
 		}
 
 		// Seek backward by a relative amount, given as offset from current position
--- a/src/Stream/BidirectionalReader.h
+++ b/src/Stream/BidirectionalReader.h
@@ -7,6 +7,11 @@ namespace OP2Utility::Stream
 {
 	class BidirectionalReader : public ForwardReader {
 	public:
+		inline void Peek(void* buffer, std::size_t size) {
+			ReadImplementation(buffer, size);
+			SeekBackward(size);
+		}
+
 		// Seek backward by a relative amount, given as offset from current position
 		virtual void SeekBackward(uint64_t offset) = 0;
 
--- a/src/Stream/BidirectionalReader.h
+++ b/src/Stream/BidirectionalReader.h
@@ -6,26 +6,17 @@
 namespace OP2Utility::Stream
 {
 	class BidirectionalReader : public ForwardReader {
-	protected:
-		// Generic peek method, which raises an exception if the full data cannot be read
-		// Note: This is named separately from Peek to prevent name hiding in derived classes
-		// All Peek methods are syntax sugar which delegate to this method
-		void PeekImplementation(void* buffer, std::size_t size)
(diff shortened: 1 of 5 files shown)


src/Stream/BidirectionalReader.h
```diff
@@ -7,6 +7,11 @@ namespace OP2Utility::Stream
 {
 	class BidirectionalReader : public ForwardReader {
 	public:
+		inline void Peek(void* buffer, std::size_t size) {
+			ReadImplementation(buffer, size);
+			SeekBackward(size);
+		}
```
DanRStevens, line 13: Is there a reason this was explicitly declared [`inline`](https://en.cppreference.com/w/cpp/language/inline)? I believe that's already implied in this context.
> A function defined entirely inside a class/struct/union definition, whether it's a member function or a non-member friend function, is implicitly an inline function.

----

This method is rather low level, in that it takes an explicit buffer size, much like the internal read method. This makes it potentially error prone to use. It is unlike the externally used read template methods, which take a `T&` and use `sizeof` internally to determine the buffer size.

I'm not sure we really need a low level version of this method.

Actually the corresponding low level read method is `protected`, rather than `public`, so we probably want to avoid introducing a `public` low level peek method.

DanRStevens (commented):
We'll probably have to create an issue to revisit the design of the unit test code. I expect there are a lot of subtle nuanced decisions to be made about the test code which we maybe weren't completely aware of before and possibly didn't get quite right. There are a lot of orthogonal concerns from the different facets of how streams are used, which are perhaps being mixed together poorly. For instance, concerns that apply to streams of size zero have implications for various methods, which may or may not exist in any given stream interface, such as the various seek methods.

src/Stream/BidirectionalReader.h
Brett208: Err, I'm pretty sure the corresponding function is public. See line 29 of Reader.h unless I'm reading it wrong.

```C++
inline void Read(void* buffer, std::size_t size) {
	ReadImplementation(buffer, size);
}
```

Do you want to make the original function protected?

I repeated the style when using the inline keyword from Reader.h. Guess I don't really have a preference. Since you are bringing up I'll assume you would prefer we not use the inline keyword in this sort of context.

DanRStevens: Ahh, right. I was looking at line 17.

I remember now, we had an internal function that was named differently to avoid name hiding in derived classes when just that one implementation method was overridden.

I suppose it does make sense to offer the public low level `Peek` function then.

I'm also seeing what you mean about multiple template methods. For some reason I was thinking it was just one template method. I forgot there were various restrictions and optimizations, which we handled with multiple overloads.

I wonder if there is a way to reduce overloads by using `constexpr if`. Or if there is a way to factor out those templates so they could be applied to both `Read` and `Peek`. That's starting to sound like future work though.

----

Hmm, that's true, we do use `inline` in a few places, and use it inconsistently. I probably didn't know what I was doing back then. We should probably look at removing those.

DanRStevens (approved):

src/Stream/BidirectionalReader.h
Brett208: Now that I think about it, I would liberally use inline in the past and probably started the precedence. I'll add an issue to remove inline keyword use. I think leeor had commented a while back that it was not needed for me to use inline so much as well...

Just added a trivial version of Peek. 

I think for now I'm happy with have 2 overloads for Peek. I was trying to think of Peek use cases. Typically, I am using it to check for something like the file type or maybe an integer representing the size of a data header tag. I don't know that it is super useful to Peek at full containers or the other more complex Read overloads. You would probably just be reading data in those situations.

src/Stream/BidirectionalReader.h
```diff
@@ -6,12 +6,28 @@
...
-		inline void Peek(void* buffer, std::size_t size) {
+	protected:
+		// Generic peek method, which raises an exception if the full data cannot be read
+		// Note: This is named separately from Peek to prevent name hiding in derived classes
+		// All Peek methods are syntax sugar which delegate to this method
+		void PeekImplementation(void* buffer, std::size_t size)
+		{
 			ReadImplementation(buffer, size);
 			SeekBackward(size);
 		}
```
DanRStevens: Err, I don't think the name hiding concern is relevant to `Peek`, as it delegates to `ReadImplementation`, which does the class specific portion. There should be no need for derived classes to override `Peek` (or `PeekImplementation`). (It's not declared `virtual`).

Sorry, I probably wasn't being very clear on that point. It's probably fine to have a `public` low level `Peek` implementation, with the actual full implementation, and there should be no need to declare it `virtual`.

DanRStevens (commented):
Thank you for making the `inline` change.

src/Stream/BidirectionalReader.h
DanRStevens: That's a good point about `Peek`. Usually it's to check for fixed sized sentinel values.

src/Stream/BidirectionalReader.h
```diff
@@ -7,6 +7,18 @@ namespace OP2Utility::Stream
...
+			ReadImplementation(buffer, size);
+			SeekBackward(size);
+		}
+
+		// Trivially copyable data types
+		template<typename T>
+		std::enable_if_t<std::is_trivially_copyable<T>::value>
+			Peek(T& object) {
+			Peek(&object, sizeof(object));
+		}
```
DanRStevens, line 20: Something I just noticed post-merge. The indentation for the `Peek` method is a bit odd. The function signature is indented the same amount as the function implementation. I think when I initially read this, I interpreted it as a 2 line function body.

Brett208: Yeah good catch. I'll fix it this weekend.

PR #393, opened 2021-02-19
Remove inline keyword from class function definitions

### LINKED ISSUE #392: Remove inline calls for functions that are already inline in class and struct header files
See PR #390 for discussion on topic. 

Standardize to not using inline unless there is a specific reason to use the keyword.

Brett208:
Closes #392

### DIFF
--- a/src/Tag.h
+++ b/src/Tag.h
@@ -67,7 +67,7 @@ namespace OP2Utility
 
 
 	// Factory method to construct a Tag from a string literal
-	inline constexpr Tag MakeTag(const char(&tagText)[5]) {
+	constexpr Tag MakeTag(const char(&tagText)[5]) {
 		return Tag(tagText);
 	}
 }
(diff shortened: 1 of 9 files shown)


src/Tag.h
```diff
@@ -67,7 +67,7 @@ namespace OP2Utility
 
 
 	// Factory method to construct a Tag from a string literal
-	inline constexpr Tag MakeTag(const char(&tagText)[5]) {
+	constexpr Tag MakeTag(const char(&tagText)[5]) {
 		return Tag(tagText);
 	}
```
DanRStevens, line 72: Note for posterity: This use of `inline` is outside of a class, though because of the `constexpr` it is still implicitly `inline`.

https://en.cppreference.com/w/cpp/language/inline
> A function declared constexpr is implicitly an inline function.

----

All the other methods with `inline` being removed are defined within a class definition:
> A function defined entirely inside a class/struct/union definition, whether it's a member function or a non-member friend function, is implicitly an inline function.

DanRStevens (approved):

src/Tag.h
Brett208: Thank you for reviewing.

PR #394, opened 2021-02-19
Add ability to peek at stream to determine if it is a bitmap file or not

Brett208:


src/Bitmap/BitmapFile.cpp
```diff
@@ -1,9 +1,25 @@
...
+		std::array<char, 2> fileSignature;
+		reader.Peek(fileSignature);
+
+		return fileSignature == BmpHeader::FileSignature;
+	}
+
+	bool BitmapFile::PeekIsBitmap(Stream::BidirectionalReader&& reader)
+	{
+		return PeekIsBitmap(reader); // Delegate to lvalue overload
+	}
```
DanRStevens, line 21: It would be somewhat peculiar to use a temporary rvalue in this manner. The programmer would need to create a stream, pass it to this method to check if it was a bitmap, and upon completion the stream would be discarded.

That may still have some valid uses, just seems odd to discard the stream before potentially using it after classifying it as bitmap data.

I think I might want to accept this though just due to consistency in the API.

DanRStevens (approved):

PR #397, opened 2021-02-27
Allow detecting & reporting when v4 or v5 bitmap image headers are encountered

Brett208:
I was trying to import a bitmap the other day, and OP2Utility reported the header size was incorrect. This was confusing because the bitmap was well formed. Turned out I was trying to load a bitmap saved using the v5 header format. This should provide a better error experience, especially for someone who does not understand how to inspect the size property of a bitmap using a hex editor...

### DIFF
--- a/src/Bitmap/ImageHeader.cpp
+++ b/src/Bitmap/ImageHeader.cpp
@@ -1,4 +1,5 @@
 #include "ImageHeader.h"
+#include "../StringUtility.h"
 #include <string>
 #include <stdexcept>
 #include <algorithm>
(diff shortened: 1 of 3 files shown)


DanRStevens (approved):
Huh, interesting failure mode. I wonder how the alternate formats were generated.

I wonder if it would make sense to support these bitmaps by basically skipping over the extra header fields. Both newer formats contain the original fields as the initial part of the header, so it might be feasible to check the header size is a certain minimum, and to skip any fields beyond the core set.

